PR #444, opened 2022-05-17
Move viewer theme state management to context

msnyder-msft:
### Summary of changes 🔍 
Pulling out the state management for the viewer theme into a context with a reducer so that it is more portable.

No visual changes.
image

### Testing 🧪
- [x] Play with the theme picker and make sure it does what you expect

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Models/Context/SceneThemeContext/SceneThemeContext.tsx
+++ b/src/Models/Context/SceneThemeContext/SceneThemeContext.tsx
@@ -28,7 +28,7 @@ const logDebugConsole = getDebugLogger('SceneThemeContext', debugLogging);
 const SceneThemeContextInstance = React.createContext<ISceneThemeContext>(null);
 export const useSceneThemeContext = () => useContext(SceneThemeContextInstance);
 
-let shouldPersistTheme = true;
+let isThemePersistenceEnabled = true;
 const DEFAULT_OBJECT_COLOR_INDEX = 2;
 
 export const SceneThemeContextReducer: (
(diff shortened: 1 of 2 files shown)


src/Components/ModelViewerModePicker/ModelViewerModePicker.tsx
```diff
@@ -134,105 +89,41 @@ const ModelViewerModePicker: React.FC<ModelViewerModePickerProps> = ({
...
-            background: backgroundColors[0].color,
-            style: ViewerModeStyles.Default
-        });
-
-        viewerModeUpdated({
-            objectColor: null,
-            background: backgroundColors[0].color,
-            style: ViewerModeStyles.Default
-        });
+        // // eslint-disable-next-line react-hooks/exhaustive-deps
```
ccrowley96, line 92: nit double comment here -- may also be able to remove this line entirely

src/Models/Context/SceneThemeContext/SceneThemeContext.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    // skip wrapping if the context already exists
+    const existingContext = useSceneThemeContext();
+    if (existingContext) {
+        return <>{children}</>;
+    }
+
+    const { initialState, usePersistedTheme = true } = props;
+    const { t } = useTranslation();
+    shouldPersistTheme = usePersistedTheme; // store the selection globally
```
ccrowley96, line 109: nit: starting a variable name with `use` is unofficially reserved for react hooks.  Maybe rename to `isThemePersisted`?

ccrowley96 (approved):
Great refactor.  Code looks good.  Just left a couple nits.

src/Models/Context/SceneThemeContext/SceneThemeContext.tsx
msnyder-msft: Fair. Updated.

samjosephgates (approved):
Looks good. The only difference in behavior that I can see is as follows:

Current implementaton: 
Select transparent style (it defaults to blue)
Change object color to gold
Change back to default style
Change back to transparent
It is still gold

New implementation
Select transparent style (it defaults to blue)
Change object color to gold
Change back to default style
Change back to transparent
It defaults back to blue

Not sure which is right, they are just different.

PR #458, opened 2022-05-19
Edit Dialog Refactor

abalogun-dev:
### Summary of changes 🔍 
Refactored to enable the component to use a reducer for its more complex state management

### Testing 🧪
 Just check code

### Checklist ✔️
- [ ] Added relevant labels


src/Components/SceneList/Internal/SceneDialog.state.ts
```diff
@@ -0,0 +1,74 @@
...
+    isSelectedFileExistInBlob: false,
+    isOverwriteFile: false,
+    blobsInContainer: [],
+    selectedFile: null,
+    selected3DFilePivotItem: SelectionModeOf3DFile.FromContainer
+};
+
+export const SceneDialogReducer: (
+    draft: SceneDialogState,
+    action: IAction
```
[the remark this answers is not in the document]
msnyder-msft: this one

src/Components/SceneList/Internal/SceneDialog.state.ts
```diff
@@ -0,0 +1,85 @@
...
+            draft.blobsInContainer = payload;
+            break;
+        case SET_SELECTED_FILE:
+            draft.selectedFile = payload;
+            break;
+        case SET_SELECTED_3D_FILE_PIVOT_ITEM:
+            draft.selected3DFilePivotItem = payload;
+            break;
+        case RESET_FILE:
+            draft.isSelectedFileExistInBlob = payload.isSelectedFileExistInBlob;
```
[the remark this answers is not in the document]
msnyder-msft: agreed. let's not take payloads for those actions.

src/Components/SceneList/Internal/SceneDialog.tsx
```diff
@@ -195,35 +203,49 @@ const SceneDialog: React.FC<ISceneDialogProps> = ({
...
         }
-    }, [selected3DFilePivotItem]);
+    }, [state.selected3DFilePivotItem]);
 
     const handleNameChange = useCallback(
-        (e) => {
+        (e, newValue?: string) => {
             if (sceneToEdit) {
                 const selectedSceneCopy: IScene = deepCopy(sceneRef.current);
                 selectedSceneCopy.displayName = e.currentTarget.value;
```
msnyder-msft, line 226: let's not use any `e.currentTarget` in this file. You can get the values off the second prop.

src/Components/SceneList/Internal/SceneDialog.tsx
```diff
@@ -195,35 +203,49 @@ const SceneDialog: React.FC<ISceneDialogProps> = ({
...
         },
         [sceneToEdit, scene]
     );
 
     const handleSceneDescriptionChange = useCallback(
-        (e) => {
+        (e, newValue?: string) => {
             if (sceneToEdit) {
                 const selectedSceneCopy: IScene = deepCopy(sceneRef.current);
                 selectedSceneCopy.description = e.currentTarget.value;
```
msnyder-msft, line 242: same here. let's clean up the use of target

src/Components/SceneList/Internal/SceneDialog.tsx
```diff
@@ -103,24 +123,12 @@ const SceneDialog: React.FC<ISceneDialogProps> = ({
     onEditScene,
     renderBlobDropdown
 }) => {
-    const [newSceneName, setNewSceneName] = useState('');
-    const [newSceneDescription, setNewSceneDescription] = useState('');
-    const [newLatitudeValue, setNewLatitudeValue] = useState(undefined);
-    const [newLongtitudeValue, setNewLongtitudeValue] = useState(undefined);
-    const [newSceneBlobUrl, setNewSceneBlobUrl] = useState('');
+    const [state, dispatch] = useReducer(
+        SceneDialogReducer,
```
msnyder-msft, line 127: shouldn't we use the sceneToEdit to populate the reducer initial state? We shouldn't need to fork all the update functions based on edit/create. They should all be operating on the same state object, that's the point of hte reducer.

src/Components/SceneList/Internal/SceneDialog.tsx
```diff
@@ -238,7 +260,10 @@ const SceneDialog: React.FC<ISceneDialogProps> = ({
                 selectedSceneCopy.latitude = lat;
```
msnyder-msft, line 260: we should not need the forked logic for `sceneToEdit` everything should go through the reducer.

src/Components/SceneList/Internal/SceneDialog.tsx
```diff
@@ -264,23 +292,29 @@ const SceneDialog: React.FC<ISceneDialogProps> = ({
...
     const handleFileOverwriteChange = useCallback((_e, checked: boolean) => {
-        setIsOverwriteFile(checked);
+        dispatch({
+            type: SET_IS_OVER_WRITE_FILE,
+            payload: checked
+        });
     }, []);
 
     const getNewScene = (assetUrl: string) => {
         return {
```
msnyder-msft, line 312: teh state should already have a constructed scene object, this should not be required. Please have the reducer return the built out state object directly.

src/Components/SceneList/Internal/SceneDialog.tsx
```diff
@@ -321,72 +356,83 @@ const SceneDialog: React.FC<ISceneDialogProps> = ({
...
-        setNewSceneBlobUrl('');
-        setIsSelectedFileExistInBlob(false);
-        setIsOverwriteFile(false);
-        setBlobsInContainer([]);
-        setSelectedFile(null);
-        setSelected3DFilePivotItem(SelectionModeOf3DFile.FromContainer);
-    }, [isShowOnGlobeEnabled]);
+    }, []);
 
     const isSubmitButtonDisabled = useMemo(() => {
```
msnyder-msft, line 405: can we have hte reducer compute the "isValid" state? That can we updated anytime a property is updated and stored on the state there. The reducer alreayd knows all the values so it would make sense to have this computed there.

msnyder-msft (changes requested):
I think there are still outstanding comments. Primarily the type for the action payloads, and resetting back to default state rather than taking it from the payload.

src/Components/SceneList/Internal/SceneDialog.state.ts
```diff
@@ -0,0 +1,85 @@
...
+        case RESET_SCENE:
+            draft = defaultSceneDialogState;
+            break;
+        case RESET_OVERWRITE_FILE_AND_EXIST_IN_BLOB:
+            draft.isSelectedFileExistInBlob = payload.isSelectedFileExistInBlob;
+            draft.isOverwriteFile = payload.isOverwriteFile;
+            break;
+        default:
+            break;
+    }
```
darsney, line 84: @msnyder-msft, do you mean doing something after this switch statement where we calculate isValid?  not sure what that looks like, but seems like it makes sense like...

`draft.isValid = calculateIsSceneSubmitPossible(draft)`

not sure how else to do that

src/Components/SceneList/Internal/SceneDialog.tsx
darsney: im not clear on how the reducer would continuously calculate the isValid state, check my comment above on the naive way i thought about it

src/Components/SceneList/Internal/SceneDialog.state.ts
msnyder-msft: if you wanna check validity on every action, sure. But I think we only care about a few fields so I would only update it when those particular actions are called. Your sample there is how I would do it.

src/Components/SceneList/Internal/SceneDialog.tsx
msnyder-msft: it will only need to calculate when particular actions are triggered to update those specific fields it cares about.

msnyder-msft:
Can you merge latest into here? At first glance it looked like the snapshot updates it's tring to do are before some bug fixes went in.

src/Components/SceneList/Internal/SceneDialog.state.ts
```diff
@@ -0,0 +1,104 @@
...
+        id: '',
+        displayName: '',
+        description: '',
+        latitude: undefined,
+        longitude: undefined,
+        elements: [],
+        behaviorIDs: [],
+        assets: []
+    },
+    sceneBlobUrl: scene ? scene?.assets?.[0]?.url || '' : '',
```
msnyder-msft, line 23: this doesn't seem quite right. Can we get it to be the nicely formatted value? main on the left, your branch on the right.
image

msnyder-msft (approved):
Aside from the uploaded file name looking nicer in Main, I think this is ready to go. Let's see if we can figure out how to make that look pretty and let's send it.

PR #461, opened 2022-05-19
Zarmada/working/ADTS-145

miltonzarmada:
####   ADTS-145 Ability to save and manage multiple projects

###% Testing 🧪

#### Checklist ✔️
- [ ] Added relevant labels


src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -126,6 +128,18 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
...
+        // Reload elements on project change
+        if (reloadProject) {
+            setElements(getStoredElements());
+
+            dispatch({
+                type: SET_OAT_RELOAD_PROJECT,
+                payload: false
+            });
+        }
+    }, [reloadProject]);
```
jamescarpinter, line 144: I wonder if setting a flag is the right approach here. Perhaps we should be pull the elements state up into the editor page? So we can set it from there on project change?

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -1,19 +1,43 @@
...
+    elements,
+    onImportClick,
+    dispatch,
+    state
+}: OATHeaderProps) => {
     const { t } = useTranslation();
     const headerStyles = getHeaderStyles();
+    const [subMenuActive, setSubMenuActive] = useState(false);
+    const [modalOpen, setModalOpen] = useState(false);
+    const [modalBody, setModalBody] = useState('save');
```
jamescarpinter, line 40: What does this default state value mean?

Should this be an enum?

src/Components/OATHeader/internal/FileSubMenu.tsx
```diff
@@ -0,0 +1,178 @@
...
+    const handleSave = () => {
+        setSubMenuActive(false);
+
+        if (isFileStored) {
+            // Update file
+            const editorData = JSON.parse(
+                localStorage.getItem(OATDataStorageKey)
+            );
+            files[fileIndex].data = editorData;
+            localStorage.setItem(OATFilesStorageKey, JSON.stringify(files));
```
jamescarpinter, line 53: I think it would be useful to encapsulate the localStorage code into some kind of service class.

I.e. so we can expose more meaningful operations to save/load etc.

src/Components/OATHeader/internal/FileSubMenu.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+    const handleSave = () => {
+        setSubMenuActive(false);
+
+        if (isFileStored) {
+            // Update file
+            const editorData = JSON.parse(
+                localStorage.getItem(OATDataStorageKey)
+            );
+            files[fileIndex].data = editorData;
```
jamescarpinter, line 52: If we're changing state, we probably should be calling setState?

src/Components/OATHeader/internal/FormOpen.tsx
```diff
@@ -0,0 +1,113 @@
...
+                    text: file.name
+                };
+            });
+
+            setFiles(formattedFiles);
+        }
+    };
+
+    useEffect(() => {
+        formatFilesToDropDownOptions();
```
jamescarpinter, line 77: Can this value change? Should we have a dependency here?

Or could we just call formatFiles whenever we need it?

src/Components/OATHeader/internal/FormOpen.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+        dispatch({
+            type: SET_OAT_PROJECT_NAME,
+            payload: selectedFile.text
+        });
+
+        dispatch({
+            type: SET_OAT_RELOAD_PROJECT,
+            payload: true
+        });
```
jamescarpinter, line 49: I wonder if we want to create a new operation to dispatch that sets all three of these?

src/Components/OATHeader/internal/FormSaveAs.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+            <div className={headerStyles.modalRow}>
+                <Text>{t('OATHeader.saveAs')}</Text>
+                <TextField
+                    placeholder={t('OATHeader.enterAName')}
+                    value={projectName}
+                    onChange={(e, v) => handleProjectNameChange(v)}
+                    errorMessage={
+                        error
+                            ? 'Project with same name already exists, would you like to override'
```
jamescarpinter, line 114: Should this be localized?

src/Components/OATHeader/internal/FormSaveAs.tsx
```diff
@@ -0,0 +1,136 @@
...
+                            : null
+                    }
+                />
+            </div>
+
+            <div className={headerStyles.modalRowFlexEnd}>
+                <PrimaryButton
+                    text={error ? 'Overwrite' : t('OATHeader.save')}
```
jamescarpinter, line 122: Localize?

src/Components/OATHeader/internal/ModalSaveCurrentProjectAndClear.tsx
```diff
@@ -0,0 +1,95 @@
...
+    setModalBody,
+    state
+}: IModal) => {
+    const { t } = useTranslation();
+    const headerStyles = getHeaderStyles();
+    const promptTextStyles = getPromptTextStyles();
+    const { projectName } = state;
+
+    const handleOnSave = () => {
+        const files = JSON.parse(localStorage.getItem(OATFilesStorageKey));
```
jamescarpinter, line 36: We parse from storage *a lot*. Can't we keep this in state? And manage it there?

jamescarpinter (changes requested):
A few random comments to start this review. Happy to discuss.

src/Components/OATHeader/internal/FileSubMenu.tsx
miltonzarmada: I think in this case we do want to use localStorage and we're overwriting an existing file with new progress

jamescarpinter (approved):
Approving so we can get a build out and, as discussed, we'll sort out the storage piece separately.

PR #483, opened 2022-05-23
Added default theme environment reflections

coryboyle:
| Before Reflections  | After Reflections Added |
| ------------- | ------------- |
| MicrosoftTeams-image |  MicrosoftTeams-image (1) |
| MicrosoftTeams-image (2) | MicrosoftTeams-image (3)  |

### Testing 🧪
 > Please test the default in all environment scenes. A reflection image is there for the material of the mesh to reflect, if it looks "too shiny", then the model itself needs to be made less reflective. We can create a custom reflection map to effect color at some point in the future.

### Checklist ✔️
- [ ] Added relevant labels


darsney (approved):
LGTM

PR #486, opened 2022-05-23
Zarmada/working/ADTS-133

miltonzarmada:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Pages/OATEditorPage/Internal/OATErrorPage.tsx
+++ b/src/Pages/OATEditorPage/Internal/OATErrorPage.tsx
@@ -1,0 +1,32 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { getEditorPageStyles } from '../OATEditorPage.Styles';
+
+const OATErrorPage = ({ error }) => {
+    const { t } = useTranslation();
+    const EditorPageStyles = getEditorPageStyles();
+
+    return (
+        <div className={EditorPageStyles.errorContainer}>
+            <h2 className={EditorPageStyles.errorPageHeader}>
+                {t('OATErrorBoundary.modalHeader')}
+            </h2>
+            <h3 className={EditorPageStyles.errorPageMessageHeader}>
+                {t('OATErrorBoundary.messageHeader')}
+            </h3>
+            <div>
+                <pre>
+                    {error.name} | {error.message}
+                </pre>
+            </div>
+            <h3 className={EditorPageStyles.errorPageStackHeader}>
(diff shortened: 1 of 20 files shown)


src/Pages/OATEditorPage/Internal/OATErrorPage.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { getEditorPageStyles } from '../OATEditorPage.Styles';
+
+const OATErrorPage = ({ error }) => {
+    const { t } = useTranslation();
+    const EditorPageStyles = getEditorPageStyles();
```
jamescarpinter, line 7: Should this be `editorPageStyles`? I.e. lower case `E`. Depends on the pattern that's used elsewhere.

src/Pages/OATEditorPage/OATEditorPage.tsx
```diff
@@ -5,12 +5,14 @@ import OATGraphViewer from '../../Components/OATGraphViewer/OATGraphViewer';
 import OATPropertyEditor from '../../Components/OATPropertyEditor/OATPropertyEditor';
 import OATImport from './Internal/OATImport';
 import { getEditorPageStyles } from './OATEditorPage.Styles';
+import { ErrorBoundary } from 'react-error-boundary';
```
jamescarpinter, line 8: Is this the package used by the `iot-cardboard-js` team in their work?

jamescarpinter (changes requested):
Minor comments only. Thanks!

src/Pages/OATEditorPage/OATEditorPage.tsx
miltonzarmada: No, on other `pages` the use of `{ ErrorBoundary } from 'react-error-boundary'` was not found.
Most similar approach found is 
`{ withErrorBoundary } from '.../Models/Context/ErrorBoundary'` where function wraps component  withErrorBoundary(Component) and replaces it with a gray error box whenever an error occurs

jamescarpinter: Is this a new package then? If so, are we missing a package.json change?

miltonzarmada: It just worked on import, I did not have to install it as a new package.
So, I think it is already pre enabled

jamescarpinter (approved):

PR #490, opened 2022-05-24
ADTS-113

JoseZarmada:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Components/OATHeader/OATHeader.tsx
```diff
@@ -77,6 +91,42 @@ const OATHeader = ({ elements, onImportClick }: OATHeaderProps) => {
...
+                try {
+                    const content = await current.text();
+                    newItem.content = JSON.parse(content);
+                } catch (error) {
+                    console.log(Error(error));
+                    alert(Error(error));
+                }
+                items.push(newItem.content);
+            }
+            dispatch;
```
jamescarpinter, line 122: What is this line?

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -77,6 +91,42 @@ const OATHeader = ({ elements, onImportClick }: OATHeaderProps) => {
...
+                const newItem = {
+                    name: current.name,
+                    size: prettyBytes(current.size),
+                    status: FileUploadStatus.Uploading
+                } as IFileItem;
+                try {
+                    const content = await current.text();
+                    newItem.content = JSON.parse(content);
+                } catch (error) {
+                    console.log(Error(error));
```
jamescarpinter, line 117: Why do we need to wrap this is an error?

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -77,6 +91,42 @@ const OATHeader = ({ elements, onImportClick }: OATHeaderProps) => {
         }
     ];
 
+    useEffect(() => {
+        const newFiles = [];
+        acceptedFiles.forEach((sF) => {
+            if (sF.type === 'application/json') {
+                newFiles.push(sF);
+            }
```
jamescarpinter, line 99: How do we inform the user about files that we don't accept?

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -131,85 +131,100 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
...
                     (element) => element.id === input['@id']
                 );
-                if (!node) {
-                    let relationships = [];
-                    let contents = [];
-                    input['contents'].forEach((content) => {
-                        if (content['@type'] === OATComponentHandleName) {
-                            const componentRelationship = new ElementEdge(
+                if (node) {
+                    importModelsList = importModelsList.reduce(
```
jamescarpinter, line 141: Can you help me understand what's happening in this reduce function? If we're filtering, then `filter` might be a better option?

jamescarpinter (changes requested):
Initial comment inline. Thanks.

src/Components/OATHeader/OATHeader.tsx
JoseZarmada: Typo Just Fixed

src/Components/OATHeader/OATHeader.tsx
JoseZarmada: Took that from the example class, and Milton is actually going to be working In the Error Handling Component, so we should be using that instead eventually

src/Components/OATHeader/OATHeader.tsx
JoseZarmada: Then Again using The Error Handling Component should be a good Idea for this Also

jamescarpinter: Let's use an alert until then.

src/Components/OATHeader/OATHeader.tsx
jamescarpinter: Can you please provide a link to the example class?

I would've expected a read/parse error to result in `error` already being of type `Error`. I.e. can't we just `console.log(error);`?

jamescarpinter (approved):

PR #493, opened 2022-05-24
Zarmada/working/adts 128

JoseZarmada:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -49,35 +49,127 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
     } = useContext(ElementsContext);
     const graphViewerStyles = getGraphViewerStyles();
     const theme = useTheme();
+    sourceY = sourceY - 6;
```
jamescarpinter, line 52: Why `6`? Let's define it as a well known constant.

Also, I'd prefer not to edit props. We should create a new variable.

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -49,35 +49,127 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
...
 
     const element = elements.find((x) => x.id === id);
+    let inheritancePolygon = '';
```
jamescarpinter, line 55: This is a significant amount of logic to have in the main render flow. Can we please break it up into well named methods? We should also consider using `useMemo` so we can ensure it doesn't run unless something actually changes. Also, a few comments might be good to explain the underlying algorithm that has been implemented.

src/Components/OATGraphViewer/OATGraphViewer.styles.tsx
```diff
@@ -49,29 +54,76 @@ export const getGraphViewerStyles = () => {
...
                 background: theme.semanticColors.variantBorder
             } as IStyle
         ],
+        untargetRelationshipHandleHidden: [
+            classNames.untargetRelationshipHandleHidden,
+            {
+                left: '60% !Important',
+                top: '50% !Important',
+                background: 'transparent !Important',
+                border: '0px !Important'
```
jamescarpinter, line 124: Why is important used so heavily through here? Can't we just provide more specific CSS selectors to override the inherited styles?

jamescarpinter (changes requested):

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
jamescarpinter: Thanks for turning this into a `useMemo`. Can we please also break it up into well named methods? Rather than one big long one?

JoseZarmada: Created 2 new methods getSourceComponents and getTargetComponents let me know if you have any more questions

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -89,44 +216,19 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
             const sourceBetaAngle = Math.asin(sourceBase / sourceHypotenuse);
             const alphaAngle = Math.asin(triangleHeight / triangleHypotenuse);
             const betaAngle = 1.5708 - alphaAngle;
```
jamescarpinter, line 218: What's `1.5708`? Can we please define it as a well named constant?

JoseZarmada: Its the Radian Value for 90 degree angles. I'll name it as a constant

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -50,58 +51,326 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
...
+                sourceBetaAngle,
+                sourceHeight,
+                alphaAngle,
+                adjustedSourceY,
+                baseVector,
+                heightVector
+            );
+            componentPolygon = sourceComponents.componentPolygon;
+            polygonSourceX = sourceComponents.polygonSourceX;
+            polygonSourceY = sourceComponents.polygonSourceY;
```
jamescarpinter, line 316: We could do something like: `polygons = { polygons, ...sourceComponents };` (i.e. rather than setting everything using `let` at the start, it could just be `let polygons = {};`.

jamescarpinter (approved):

PR #497, opened 2022-05-25
Zarmada/working/adts 143

JoseZarmada:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -68,11 +69,13 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
     const [elements, setElements] = useState(
         !storedElements ? [] : storedElements
     );
-    const defaultPosition = 100;
+    let defaultPositionX = 100;
+    let defaultPositionY = 100;
```
jamescarpinter, line 73: These will get reset to the default values each time the component re-renders.

Maybe we return these from the positionLookUp method instead?

jamescarpinter (changes requested):
Minor comment only. Thanks!

src/Components/OATGraphViewer/OATGraphViewer.tsx
JoseZarmada: I was counting on that to happen, so, the LookUp method will fill the gaps.

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -653,6 +655,44 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
...
+    const positionLookUp = (node = null) => {
+        let defaultPositionX = 100;
+        let defaultPositionY = 100;
+        const areaDistanceX = 250;
+        const areaDistanceY = 80;
+        const maxWidth = 800;
+        const defaultPosition = 100;
+        const minWidth = 300;
+        const minHeight = 100;
+        const lookUpElements = node ? [...elements, ...node] : [...elements];
```
jamescarpinter, line 667: Is `node` the best name for this param? It looks like it's an array of elements?

jamescarpinter (approved):
Please see minor comment re renaming a parameter. If it's fixed then this is good to go.

Issue #496, opened 2022-05-25
Pop-over should be removed when switching scenes in Viewer


pasanchMSFT:
**Describe the bug**
In View mode, when switching scenes with the breadcrumb dropdown the pop-over showing object properties remains in the screen and has null values.

**To Reproduce**
1. Go to any scene on view mode
2. Click on an element or a behavior to show pop-over
3. Change the scene with dropdown on the breadcrumb
4. See pop-over with null values

**Expected behavior**
Pop-over with null values shouldn't be shown on scene change.

**Screenshots**
![image](https://user-images.githubusercontent.com/55852250/170331344-b9adbb64-abea-4ed0-a1ae-60304d1931a4.png)

**Browser (please complete the following information):**
 Edge

**Additional context**
N/A

PR #499, opened 2022-05-26
Close modal on scene change

pasanchMSFT:
### Summary of changes 🔍 
On view mode, when a user changes scene and has a selected element a pop-over persisted. Now the pop-over no longer persists the scene swap.

Resolves: #496 

### Testing 🧪
1. Go to view mode
2. Select an item
3. Change scenes with breadcrumb
4. Pop-over should no longer appear

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/SceneBreadcrumb/Internal/SceneDropdown.tsx
+++ b/src/Components/SceneBreadcrumb/Internal/SceneDropdown.tsx
@@ -7,7 +7,7 @@ import { SceneDropdownProps } from './SceneDropdown.types';
 
 const SceneDropdown: React.FC<SceneDropdownProps> = ({
     sceneId,
-    onCloseBehaviorsModal
+    onSceneChange
 }) => {
     const scenePageContext = useContext(ADT3DScenePageContext);
     if (!scenePageContext) {
@@ -24,8 +24,8 @@ const SceneDropdown: React.FC<SceneDropdownProps> = ({
         : [];
     const onChange = useCallback(
         (_e, option: IDropdownOption) => {
-            if (onCloseBehaviorsModal) {
-                onCloseBehaviorsModal();
+            if (onSceneChange) {
+                onSceneChange();
             }
             scenePageContext.handleOnSceneSwap(String(option.key));
         },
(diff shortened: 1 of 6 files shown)


src/Components/SceneBreadcrumb/Internal/SceneDropdown.tsx
```diff
@@ -5,7 +5,10 @@ import { ADT3DScenePageContext } from '../../../Pages/ADT3DScenePage/ADT3DSceneP
 import { dropdownStyles } from './SceneDropdown.styles';
 import { SceneDropdownProps } from './SceneDropdown.types';
 
-const SceneDropdown: React.FC<SceneDropdownProps> = ({ sceneId }) => {
+const SceneDropdown: React.FC<SceneDropdownProps> = ({
+    sceneId,
+    onCloseBehaviorsModal
```
msnyder-msft, line 10: I like the pattern, but would rather use a name like `onSelectionChange` or `onChange`.

src/Components/SceneBreadcrumb/SceneBreadcrumb.types.ts
```diff
@@ -15,6 +15,7 @@ export interface IBaseBreadcrumbProps {
     classNames: BreadcrumbClassNames;
     onSceneClick?: () => void;
     onCancelForm?: () => void;
+    onCloseBehaviorsModal?: () => void;
```
msnyder-msft, line 18: onSceneChange?

msnyder-msft (approved):
LGTM. I think this looks clean and works great. Would just update the naming to be more generic in all the places.

src/Components/SceneBreadcrumb/SceneBreadcrumb.types.ts
pasanchMSFT: I'll think of another name then, maybe `triggerCloseBehaviorsModal` because `onSceneChange` is too generic and this is specific to the modal in View mode. The modal is not present in Build mode.

pasanchMSFT: I see what you mean, will make the change

PR #502, opened 2022-05-30
Zarmada/working/ADTS-154

miltonzarmada:
### Summary of changes 🔍 
-  State refractor

### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Pages/OATEditorPage/OATEditorPage.tsx
```diff
@@ -47,11 +42,27 @@ const OATEditorPage = ({ theme }) => {
         }
     }, []);
 
+    // Handle models persistence
+    useEffect(() => {
+        // Update oat-data storage
+        const editorData = JSON.parse(localStorage.getItem(OATDataStorageKey));
```
jamescarpinter, line 48: Can we use the Utils methods for this and the set operation too?

src/Components/OATHeader/internal/ModalSaveCurrentProjectAndClear.tsx
```diff
@@ -66,9 +69,9 @@ export const ModalSaveCurrentProjectAndClear = ({
 
             <div className={headerStyles.modalRowCenterItem}>
                 <Text styles={promptTextStyles}>
-                    {`${t('OATHeader.doYouWantToSaveChangesYouMadeTo')} ${
-                        project.projectName
-                    }?`}
+                    {`${t(
+                        'OATHeader.doYouWantToSaveChangesYouMadeTo'
+                    )} ${projectName}?`}
```
jamescarpinter, line 74: We should be using interpolation in scenarios like this one: https://www.i18next.com/translation-function/interpolation

Some languages might have `projectName` in different places within the string so this provides additional flexibility.

jamescarpinter (changes requested):
This is looking really good! Minor comments only.

jamescarpinter:
@JoseZarmada would be great if you could take a look at this as well.

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -108,6 +200,18 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
...
+        const potentialElements = getGraphViewerElementsFromModels(
+            models,
+            modelPositions
+        );
+
+        if (JSON.stringify(potentialElements) !== JSON.stringify(elements)) {
+            setElements(potentialElements);
+        }
+    }, [models]);
+
```
JoseZarmada, line 214: Is there a way models can change outside this component as far as I can see we update Models on the translated output  and then translating them back into the element list right here but the translated output is a memo triggered by the element list

src/Components/OATHeader/internal/FormSaveAs.tsx
```diff
@@ -57,17 +64,28 @@ export const FormSaveAs = ({
...
         files.push({
             name: projectName,
-            data: editorData
+            data: {
+                modelPositions: newProject.modelPositions,
+                models: newProject.models,
+                projectName: newProject.projectName,
+                projectDescription: newProject.projectDescription,
+                templates: newProject.templates
+            }
```
JoseZarmada, line 87: Same here use like in line 67

src/Pages/OATEditorPage/Internal/Utils.ts
```diff
@@ -0,0 +1,24 @@
...
+export const getStoredEditorModelsData = () => {
+    const oatData = getStoredEditorData();
+    return oatData && oatData.models ? oatData.models : [];
+};
+
+// Get stored models' positions OAT-data
+export const getStoredEditorModelPositionsData = () => {
+    const oatData = getStoredEditorData();
+    return oatData && oatData.modelPositions ? oatData.modelPositions : [];
+};
```
JoseZarmada, line 24: This should be in the global Utils if you plan to use any in the components, since components can be load outside of the editor page

src/Pages/OATEditorPage/OATEditorPage.tsx
```diff
@@ -47,11 +43,27 @@ const OATEditorPage = ({ theme }) => {
...
+        const oatEditorData = {
+            ...editorData,
+            models,
+            modelPositions: modelPositions,
+            projectName,
+            projectDescription: '',
+            templates: templates
+        };
+
+        localStorage.setItem(OATDataStorageKey, JSON.stringify(oatEditorData));
```
jamescarpinter, line 59: Can we use the utils class for this set operation as well?

jamescarpinter (approved):

PR #505, opened 2022-05-31
Bug fix: Fix resolution of alias twin ids

msnyder-msft:
#### The problem
There was a bug in our resolution of the aliased twin mappings where we only included the twinId of the last element assigned to an alias. This was due to an issue where we were returning a map of alias name to twin ids, but since multiple elements on a behavior will map to the same alias, that means we only end up returning the last twinId instead of all of them. 
[Bug 14558938](https://msazure.visualstudio.com/One/_workitems/edit/14558938): Twin aliases don't work....

#### The fix
Return an array of the twin ids that aliases resolve to. This is the only info we ever used anyways so it's pretty straightforward.

Before: 
image

After:
image

### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
+++ b/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
@@ -16,7 +16,7 @@ import {
 interface IBuildModelledPropertiesParams {
     adapter: IModelledPropertyBuilderAdapter;
     primaryTwinIds: string[];
-    aliasedTwinMap?: Record<string, string>;
+    aliasedTwinIds?: string[];
     allowedPropertyValueTypes: Array<PropertyValueType>;
 }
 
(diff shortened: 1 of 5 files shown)


src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
```diff
@@ -372,7 +371,7 @@ export const mergeTagsAndMapTwinIdsToModelIds = async (
                 twinModelIdMap.set(twinId, modelId);
             });
 
-        for (const [aliasTag, aliasTwinId] of Object.entries(aliasedTwinMap)) {
+        for (const [aliasTag, aliasTwinId] of Object.entries(aliasedTwinIds)) {
```
ccrowley96, line 374: Don't we use the alias tags here?  If we're changing aliasTagMap to a list of just Ids, It seems like this will break.  I'm pretty sure this method needs to access the actual alias to construct paths like `temperatureTag.temperature` where `temperatureTag` is the alias.

PR #507, opened 2022-06-01
Fix: Aliased Twin properties

ccrowley96:
### Summary of changes 🔍 
- Fixes [ADO bug (Twin Aliases don't work...)](https://msazure.visualstudio.com/One/_workitems/edit/14558938) by updating `getTwinIdsForBehaviorInScene` method to return Ids for all aliased twins.  Refactored modelled property builder to account for this change.  Both **PrimaryTwins** & **AliasedTwins** will now show a merged set of properties derived from the combined models of each referenced twin.  The image below shows the ModelledPropertyBuilder merging properties for `TestAlias` where `A` are the properties from a Tesla model & `B` are the properties from a Pasteurization machine model.

  image

- Fixes [ADO bug (Aliased twins intellisense exposes the PrimaryTwin's properties)](https://msazure.visualstudio.com/One/_workitems/edit/14547801).  This seems to fix this bug as well.  Unless there is a special edge case that is causing intellisense aliases from breaking.
  image

### Testing 🧪
- Create two element in the scene
- Create a behavior and add both elements to the behavior
- Navigate to the Twins tab of the behavior and create a new twin alias with distinct twin Ids for each attached element
  For example: 

  image

- Check the property dropdown (Status tab) & Intellisense control (Alerts tab) and verify that properties from the primary & aliased twins look correct.  Also verify that [aliased twin intellisense is working](https://msazure.visualstudio.com/One/_workitems/edit/14547801).

- Create a value widget using an aliased property.  Verify that in the viewer, elements on the behavior you created show data for the value widget (should only show data if the element's primary twin / alias actually has the property you selected)
- I did an additional sanity check by setting up a string value widget & visualizing the $dtId property for the alias I set up.  This shows that the runtime is fetching data for the different aliased twin Ids and shows them in the popover.
  image
  image
  image

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
+++ b/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
@@ -16,7 +16,7 @@ import {
 interface IBuildModelledPropertiesParams {
     adapter: IModelledPropertyBuilderAdapter;
     primaryTwinIds: string[];
-    aliasedTwinMap?: Record<string, string>;
+    aliasedTwinMap?: Record<string, string[]>;
     allowedPropertyValueTypes: Array<PropertyValueType>;
 }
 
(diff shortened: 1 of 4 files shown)


src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
```diff
@@ -330,53 +324,46 @@ const addEntity = (
 export const mergeTagsAndMapTwinIdsToModelIds = async (
```
msnyder-msft, line 324: can we add some test cases for this file?

msnyder-msft (approved):
This looks good to me! I'd lvoe to see some tests before we complete it if we can.

ccrowley96:
> This looks good to me! I'd lvoe to see some tests before we complete it if we can.

Created an issue for adding tests to this: https://github.com/microsoft/iot-cardboard-js/issues/515.  Will check out in another PR.

PR #510, opened 2022-06-01
"clear status" button

BrennaCourtney:
Addresses [Bug 14324520](https://msazure.visualstudio.com/One/_workitems/edit/14324520): There's no way to remove "status" from a behavior once it's been configured 

Creates a clear button based on [these designs](https://www.figma.com/file/LsmZeVOZJgwZs75HPTQmgU/3D-Scenes-Mock-Screens?node-id=45547%3A225577) to be used in the Status tab.

image

image

### Testing 🧪

ModelledPropertyBuilder -> Toggle Mode With Clear Button -> 

1. Select value from dropdown when Single property is toggled -> click Clear -> should say "None"
2. Toggle Custom (advanced) -> click Clear -> switch to Single property & "None"
3. Toggle Custom (advanced) -> enter value -> click Clear -> switch to Single property & "None"

Repeat with Behavior Form -> Edit Status Tab

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/StatusTab.tsx
+++ b/src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/StatusTab.tsx
@@ -171,6 +171,7 @@ const StatusTab: React.FC<IStatusTabProps> = ({ onValidityChange }) => {
                         selectedElements
                     }}
                     mode={ModelledPropertyBuilderMode.TOGGLE}
+                    isClearEnabled={true}
                     propertyExpression={{
                         expression:
                             getStatusFromBehavior(behaviorToEdit)
--- a/src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/StatusTab.tsx
+++ b/src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/StatusTab.tsx
@@ -87,6 +87,7 @@ const StatusTab: React.FC<IStatusTabProps> = ({ onValidityChange }) => {
                 produce((draft) => {
                     // Assuming only 1 alert visual per behavior
                     let statusVisual = getStatusFromBehavior(draft);
+
                     // Edit flow
                     if (statusVisual) {
                         // selected the none option, clear the data
@@ -95,7 +96,8 @@ const StatusTab: React.FC<IStatusTabProps> = ({ onValidityChange }) => {
                             draft.visuals.splice(index, 1);
                         }
                         statusVisual[propertyName as any] = value as any;
-                    } else {
+                    } else if (value !== '') {
+                        // Create status visual if, not present & value is not 'none' aka ''
                         statusVisual = deepCopy(defaultStatusColorVisual);
                         statusVisual[propertyName as any] = value as any;
                         statusVisual.valueRanges =
--- a/src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/StatusTab.tsx
+++ b/src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/StatusTab.tsx
@@ -96,7 +96,7 @@ const StatusTab: React.FC<IStatusTabProps> = ({ onValidityChange }) => {
                             draft.visuals.splice(index, 1);
                         }
                         statusVisual[propertyName as any] = value as any;
-                    } else if (value !== '') {
+                    } else if (value) {
                         // Create status visual if, not present & value is not 'none' aka ''
                         statusVisual = deepCopy(defaultStatusColorVisual);
                         statusVisual[propertyName as any] = value as any;
--- a/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
+++ b/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
@@ -103,7 +103,7 @@ export const propertyExpressionLabelStyles: Partial<ILabelStyles> = {
     }
 };
 
-export const getActionButtonStyles = memoizeFunction(
+export const getClearButtonStyles = memoizeFunction(
     (theme: Theme): Partial<IButtonStyles> => ({
         root: {
             color: theme.palette.themePrimary,
--- a/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
+++ b/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
@@ -20,6 +20,7 @@ const classNames = {
     dropdownTitleText: `${modelledPropertyBuilderClassPrefix}-dropdown-title-text`,
     loadingSpinnerContainer: `${modelledPropertyBuilderClassPrefix}-loading-spinner-container`,
     labelContainer: `${modelledPropertyBuilderClassPrefix}-label-container`,
+    clearButton: `${modelledPropertyBuilderClassPrefix}-clear-button`,
     root: `${modelledPropertyBuilderClassPrefix}-root`
 };
 
@@ -58,6 +59,7 @@ export const getStyles = memoizeFunction(() => {
                 alignItems: 'center'
             } as IStyle
         ],
+        clearButton: [classNames.clearButton],
         root: [
             classNames.root,
             {
--- a/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
+++ b/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
@@ -110,7 +110,6 @@ export const getClearButtonStyles = memoizeFunction(
         root: {
             color: theme.palette.themePrimary,
             fontWeight: FontWeights.semibold,
-            textDecoration: 'underline',
             fontSize: FontSizes.size14,
             border: 'none',
             height: 'auto'
@@ -121,6 +120,9 @@ export const getClearButtonStyles = memoizeFunction(
             marginLeft: 0,
             marginRight: 0,
             paddingTop: isTooltipVisible ? 0 : 5
+        },
+        labelHovered: {
+            textDecoration: 'underline'
         }
     })
 );
--- a/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
+++ b/src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
@@ -111,9 +111,14 @@ export const getClearButtonStyles = memoizeFunction(
             textDecoration: 'underline',
             fontSize: FontSizes.size14,
             paddingBottom: 0,
+            paddingLeft: 0,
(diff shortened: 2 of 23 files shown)


src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
```diff
@@ -100,3 +102,18 @@ export const propertyExpressionLabelStyles: Partial<ILabelStyles> = {
         paddingTop: 5
     }
 };
+
+export const getActionButtonStyles = memoizeFunction(
```
msnyder-msft, line 106: let's call it `clearButtonStyles` since it's specific to that control.

msnyder-msft, line 106: it seems like it should be right aligned according to the designs. 
image

Actual
image

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
```diff
@@ -100,3 +102,18 @@ export const propertyExpressionLabelStyles: Partial<ILabelStyles> = {
         paddingTop: 5
     }
 };
+
```
msnyder-msft, line 105: the vertical spacing seems off to me. Can we fix the padding to align with the "property expression" label?
image

msnyder-msft (commented):
That's a great start! Just a couple style tweaks.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
msnyder-msft: Design link - https://www.figma.com/file/LsmZeVOZJgwZs75HPTQmgU/3D-Scenes-Mock-Screens?node-id=45549%3A254023

darsney: yeah also def no italics IMO

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.tsx
```diff
@@ -141,6 +145,13 @@ const ModelledPropertyBuilder: React.FC<ModelledPropertyBuilderProps> = ({
         [onChange]
     );
 
+    const onClickClearButton = useCallback(() => {
```
iremgokce, line 148: nit: for on click handler naming convention the style is usually on....Click, or even you can use handleClearButtonClick: https://stackoverflow.com/questions/60048249/what-is-the-right-name-of-event-handler-onclick-or-handleclick
```suggestion
    const onClearButtonClick = useCallback(() => {
```

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.tsx
```diff
@@ -141,6 +145,13 @@ const ModelledPropertyBuilder: React.FC<ModelledPropertyBuilderProps> = ({
...
+        onChange({
+            expression: ''
+        });
+        setInternalMode(ModelledPropertyBuilderMode.PROPERTY_SELECT);
```
iremgokce, line 152: do we really want to set the mode back to property selection? for example if I am in intellisense mode, when I click on clear, as a user I wouldn't want the UI switch back to property selection when I click on clear. Let's confirm the expected behavior with Matt

iremgokce:
great work for a start! I just left couple comments :)

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.tsx
BrennaCourtney: @darsney @azimamansuri what do you guys think?

ccrowley96:
Clicking 'Clear' multiple times seems to toggle the status on / off.  Multiple clicks should persist the Status off state.

https://user-images.githubusercontent.com/18181049/171686194-a28e404e-84a4-4b43-b5cf-a4a4aeb20cc2.mp4

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.tsx
msnyder-msft: +1 for the onClearButtonClick name

ccrowley96 (commented):
Code LGTM - good work : )

Only comment is the video recording I left.  Clicking clear multiple times seems to have some funky behaviors.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.tsx
darsney: ahh i think thats fine actually

BrennaCourtney:
> Clicking 'Clear' multiple times seems to toggle the status on / off. Multiple clicks should persist the Status off state.
> 
>  Pages._.ADT3DScenePage.-.Mock.3D.scene.page.Storybook.-.Google.Chrome.2022-06-02.10-11-33.mp4

Super weird -- seems to happen even when just clicking "None" in the dropdown multiple times, or toggling from Custom to Single when set to "None" -- looking at it now

ccrowley96 (approved):
🚀 🌔

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/StatusTab.tsx
```diff
@@ -95,7 +96,8 @@ const StatusTab: React.FC<IStatusTabProps> = ({ onValidityChange }) => {
                             draft.visuals.splice(index, 1);
                         }
                         statusVisual[propertyName as any] = value as any;
-                    } else {
+                    } else if (value !== '') {
```
msnyder-msft, line 99: do we want to specifically handle this empty string case? or should we handle more braodly to not do this in the cases of undefined/null as well?

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
```diff
@@ -103,17 +103,22 @@ export const propertyExpressionLabelStyles: Partial<ILabelStyles> = {
     }
 };
 
-export const getActionButtonStyles = memoizeFunction(
-    (theme: Theme): Partial<IButtonStyles> => ({
+export const getClearButtonStyles = memoizeFunction(
+    (theme: Theme, isTooltipVisible: boolean): Partial<IButtonStyles> => ({
         root: {
             color: theme.palette.themePrimary,
```
msnyder-msft, line 109: nit: can we sort the properties alphabetically pleazze.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.tsx
```diff
@@ -261,16 +261,18 @@ const ModelledPropertyBuilder: React.FC<ModelledPropertyBuilderProps> = ({
                             }}
                         />
                     )}
+                </Stack>
+                <Stack horizontal>
                     {isClearEnabled && (
                         <ActionButton
```
msnyder-msft, line 267: can we give it a classname as well in case someone wants to move it around specifically? you can do that by simply adding a property in the `getStyles` funciton and having it only include the `[classNamePrefix.whatever]` and just have no styles there.

msnyder-msft: Like 
```ts
clearButton: [classNames.clearButton],
```

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
```diff
@@ -100,3 +102,23 @@ export const propertyExpressionLabelStyles: Partial<ILabelStyles> = {
...
+
+export const getClearButtonStyles = memoizeFunction(
+    (theme: Theme, isTooltipVisible: boolean): Partial<IButtonStyles> => ({
+        root: {
+            color: theme.palette.themePrimary,
+            fontWeight: FontWeights.semibold,
+            textDecoration: 'underline',
+            fontSize: FontSizes.size14,
+            paddingBottom: 0,
+            paddingLeft: 0,
```
msnyder-msft, line 114: can we fix it with margin instead of padding? This causes the focus state (when you tab) to be on top of the text. 
image

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/StatusTab.tsx
ccrowley96: Value is typed as a string here so I think it's fine, but if ya wanna be extra safe, you could use

`else if(value)`

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
```diff
@@ -20,6 +20,7 @@ const classNames = {
     dropdownTitleText: `${modelledPropertyBuilderClassPrefix}-dropdown-title-text`,
     loadingSpinnerContainer: `${modelledPropertyBuilderClassPrefix}-loading-spinner-container`,
     labelContainer: `${modelledPropertyBuilderClassPrefix}-label-container`,
+    clearButton: `${modelledPropertyBuilderClassPrefix}-clear-button`,
```
msnyder-msft, line 23: nit: can we alphabetize the list. 😆 I'm a pain, I know.

BrennaCourtney: haha no prob

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
msnyder-msft: Still doesn't look right on the most recent build. Do you see something else?
image

msnyder-msft: The default padding for the button is what will give us the right focus border. If we can stick with the default it should be good out of the box.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.styles.ts
```diff
@@ -100,3 +104,23 @@ export const propertyExpressionLabelStyles: Partial<ILabelStyles> = {
         paddingTop: 5
     }
 };
+
+export const getClearButtonStyles = memoizeFunction(
+    (isTooltipVisible: boolean, theme: Theme): Partial<IButtonStyles> => ({
+        root: {
+            color: theme.palette.themePrimary,
+            fontWeight: FontWeights.semibold,
+            textDecoration: 'underline',
```
darsney, line 113: This should only underline on hover IMO, so it feels like an anchor tag

PR #520, opened 2022-06-02
coryboyle/handleinstancedmeshes

coryboyle:
### Testing 🧪
 > Testing with [this model](https://3dvdevstorage.blob.core.windows.net/cory/Conveyor%20Motor.glb).

### Checklist ✔️
- [x] Added relevant labels


src/Components/3DV/SceneView.tsx
```diff
@@ -224,8 +225,53 @@ function SceneView(props: ISceneViewProps, ref) {
     debugLog('debug', 'debug', 'SceneView Render');
     const url = modelUrl === 'Globe' ? globeUrl : modelUrl;
 
-    const sortMeshesOnLoad = () => {
+    const preProcessMeshesOnLoad = () => {
```
msnyder-msft, line 228: is it possible to pull this function out to module scope or a utility class and just pass in what you need as props?

src/Components/3DV/SceneView.tsx
```diff
@@ -224,8 +225,53 @@ function SceneView(props: ISceneViewProps, ref) {
...
+        //Initial loop to check for issues and cleanup meshes.
         for (const mesh of sceneRef.current.meshes) {
+            //Let's make sure that all mesh ids are unique
+            const matchingId = sceneRef.current.meshes.filter(
+                (x) => x.id == mesh.id
+            );
+            if (matchingId.length > 1) {
+                //Throw an error here
+                console.warn(
+                    'Loaded model contains objects with matching names. 3D Scenes Studio only supports 3d models with unique object names. Attempting to recover by forcing unique names...'
```
msnyder-msft, line 239: I'd change "matching" to "duplicate". It's clearer IMO.

src/Components/3DV/SceneView.tsx
```diff
@@ -224,8 +225,53 @@ function SceneView(props: ISceneViewProps, ref) {
...
+            if (mesh.isAnInstance) {
+                debugLog('debug', 'Breaking mesh instance: ', mesh.name);
+                const instancedMesh = mesh as BABYLON.InstancedMesh;
+                const deInstancedMesh = instancedMesh.sourceMesh.clone(
+                    instancedMesh.name + uniqueNumber,
+                    instancedMesh.parent
+                );
+                uniqueNumber++;
+                sceneRef.current.addMesh(deInstancedMesh);
+                console.log(deInstancedMesh.id);
```
msnyder-msft, line 257: got a stray console here.

src/Components/3DV/SceneView.tsx
```diff
@@ -224,8 +225,53 @@ function SceneView(props: ISceneViewProps, ref) {
...
+                sceneRef.current.removeMesh(deInstancedMesh);
+                instancedMesh.dispose();
+            }
+        }
+
+        //Loop again with the cleaned up mesh list
```
msnyder-msft, line 263: do we need to do it "again"? or could we just wait to do it till we have done the processing?

src/Components/3DV/SceneView.tsx
```diff
@@ -224,8 +225,53 @@ function SceneView(props: ISceneViewProps, ref) {
...
+        for (const mesh of sceneRef.current.meshes) {
+            //Store the original materials for color highlights
+            if (!originalMaterials.current) {
+                originalMaterials.current = {};
```
msnyder-msft, line 267: why do we loop the meshes inside of looping the meshes? not sure I track why we need the nested loop.

src/Components/3DV/SceneView.tsx
msnyder-msft: Though, I'm not seeing where we already looped

src/Components/3DV/SceneView.tsx
coryboyle: Sounds good to me

src/Components/3DV/SceneView.tsx
coryboyle: I'm definitely open to more discussion on this!  I think it makes sense where it is as the SceneView function configures and returns the babylon scene and this preprocess is modifying the meshes array attached to that scene. We are also setting up the originalMaterials array which is used in the various useEffects within the function. My guess is that overwriting the scene meshes array may create issues within the babylon scene as some things like TransformNode and AbstractMesh have specific functions they want you to call on the scene to add or remove them.

src/Components/3DV/SceneView.tsx
darsney, line 257: prolly remove this console log

darsney (approved):
Gotta play with it a bit to make sure nothing explodes but on the surface LGTM

src/Components/3DV/SceneView.tsx
coryboyle: @msnyder-msft Would you mind taking a look at the latest version of this?

msnyder-msft: sure. ok. if we can't move it fine. Just trying to not make the issue worse with a 2000+ line function. 😆

src/Components/3DV/SceneView.tsx
```diff
@@ -224,8 +225,39 @@ function SceneView(props: ISceneViewProps, ref) {
...
+        let uniqueNumber = 0;
+        //Initial loop to check for issues and cleanup meshes.
         for (const mesh of sceneRef.current.meshes) {
+            //Let's make sure that all mesh ids are unique
+            const matchingId = sceneRef.current.meshes.filter(
+                (x) => x.id == mesh.id
+            );
+            if (matchingId.length > 1) {
+                //Throw an error here
+                console.warn(
```
msnyder-msft, line 238: should we mention the offending name so they can go fix it?

msnyder-msft (approved):

PR #521, opened 2022-06-03
Scene View better handling for not loading 3D file (including CORS)

iremgokce:
### Summary of changes 🔍 
- Changed the request method from **fetch** to **axios** since fetch was returning `TypeError: Failed to fetch` in its catch method without detailed response information. Also since the promise returned from fetch() [won't reject on HTTP error status](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#checking_that_the_fetch_was_successful) even if the response is an HTTP 404 or 500, it would be better to use axios.
- However, even with axios's catch block (that falls out of the range of 2xx and when there is no response), it is hard to differentiate the CORS error and other network errors since there is no response returned. I also tried to use XMLHttpRequest to get the status code but it is always 0 for both CORS and network error like offline mode). Therefore, I just changed the content of the error.

_References:_
- https://stackoverflow.com/questions/19325314/how-to-detect-cross-origin-cors-error-vs-other-types-of-errors-for-xmlhttpreq
- https://stackoverflow.com/questions/872206/what-does-it-mean-when-an-http-request-returns-status-code-0
- https://stackoverflow.com/questions/47412384/how-to-get-the-status-code-of-a-cors-http-failed-preflight-requets
- https://stackoverflow.com/questions/5005960/xmlhttprequest-status-0-responsetext-is-empty
- https://stackoverflow.com/questions/52458341/unable-to-catch-403-in-try-catch-during-fetch
- https://axios-http.com/docs/handling_errors
- https://github.com/axios/axios/issues/838
- https://github.com/axios/axios/issues/383

image

### Testing 🧪
- Go to ADT 3D Scene Page component and update the 3D file asset url of your scene with some file url, the container of which CORS is not enabled for the origin that you run your code. Easy way is entering a file url which is stored in a container which is different than the one you access your 3d config file and make sure CORS is not enabled for it.

### Checklist ✔️
- [x] Added relevant labels


src/Components/3DV/SceneView.tsx
```diff
@@ -543,7 +547,7 @@ function SceneView(props: ISceneViewProps, ref) {
         resetCamera: (meshIds: string[]) => {
             if (meshIds?.length) {
                 createOrZoomCamera(meshIds);
-            } else {
+            } else if (sceneRef.current?.meshes) {
```
darsney, line 550: whats the story with this change?

darsney (approved):
Looks fine, just 1 comment

src/Components/3DV/SceneView.tsx
```diff
@@ -1663,9 +1681,30 @@ function SceneView(props: ISceneViewProps, ref) {
                     barHeight={10}
                 />
             )}
-            {isLoading === undefined && (
-                <div className={customStyles.errorMessage}>
-                    Error loading model. Try Ctrl-F5
+            {loadingError === 'common' && (
+                <div className={customStyles.commonErrorMessage}>
```
darsney, line 1685: Is there a reason we dont want to use an illustration message for this as well?  Probably a different illustartion than the CORS error, but might as well use that component since the styling is nice (vs just literal text)

src/Components/3DV/SceneView.tsx
iremgokce: if we don't have any 3d model loaded, the reset button in the camera control was throwing error as we click on it since the 3d scene is not loaded:
image

src/Components/3DV/SceneView.tsx
iremgokce: yeah, we can :)

src/Components/3DV/SceneView.tsx
```diff
@@ -1663,9 +1681,30 @@ function SceneView(props: ISceneViewProps, ref) {
...
+                    {t(
+                        'scenePageErrorHandling.3dAssetLoadingCommonErrorMessage'
+                    )}
+                </div>
+            )}
+            {loadingError === 'network' && (
+                <div className={customStyles.networkErrorMessage}>
```
msnyder-msft, line 1692: can we pull these errors out to a separate component and pass in teh type? this file is already massive. Then it could have common styles for both messages and would simply swap out the text that gets shown/image.

src/Resources/Locales/en.json
```diff
@@ -938,6 +938,9 @@
         "noContainerUrlErrorMessage": "Your container url is not set. Click below to set it.",
         "noADTInstanceUrlTitle": "ADT instance not set",
         "noADTInstanceUrlMessage": "Your ADT instance url is not set. Click below to set it.",
-        "configureEnvironment": "Configure environment"
+        "configureEnvironment": "Configure environment",
+        "3dAssetLoadingNetworkErrorTitle": "Error loading model",
```
msnyder-msft, line 942: can we make a sceneView section of the file?

src/Resources/Locales/en.json
```diff
@@ -938,6 +938,9 @@
...
+        "3dAssetLoadingNetworkErrorMessage": "This could be a CORS issue, invalid blob url or a network error. Please contact to the owner of the container where this asset file is stored or try again later.",
```
msnyder-msft, line 943: typo here. "Please contact the owner"

src/Resources/Locales/en.json
```diff
@@ -938,6 +938,9 @@
...
+        "3dAssetLoadingCommonErrorMessage": "Error loading model. Try Ctrl-F5."
```
msnyder-msft, line 944: I don't think we should say "Ctrl-F5". Can we just say "Error loading model. Try clearing your cache and try again"

src/Components/3DV/SceneView.tsx
```diff
@@ -842,9 +845,22 @@ function SceneView(props: ISceneViewProps, ref) {
...
+                    if (e.isAxiosError && typeof e.response === 'undefined') {
+                        // Network error, this could be a CORS issue or a dropped internet connection. It is not possible for us to know.
+                        console.error(
+                            'Error loading model. This could be a CORS issue, invalid blob url or network error.',
+                            s,
+                            e
+                        );
+                        success = false;
+                        setIsLoading(false);
+                        setLoadingError('network');
```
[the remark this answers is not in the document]
msnyder-msft: cool. That's great. Can we just mention that here then above the if condition that axios signals network errors that way?

src/Components/3DV/SceneView.tsx
```diff
@@ -842,9 +845,22 @@ function SceneView(props: ISceneViewProps, ref) {
                 engineRef.current,
                 (e: any) => onProgress(e),
                 (s: any, m: any, e: any) => {
-                    console.error('Error loading model. Try Ctrl-F5', s, e);
-                    success = false;
-                    setIsLoading(undefined);
+                    if (e.isAxiosError && typeof e.response === 'undefined') {
```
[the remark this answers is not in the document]
iremgokce: nope it can be any

src/Components/3DV/SceneView.tsx
```diff
@@ -842,9 +845,22 @@ function SceneView(props: ISceneViewProps, ref) {
...
+                        // Network error, this could be a CORS issue or a dropped internet connection. It is not possible for us to know.
+                        console.error(
+                            'Error loading model. This could be a CORS issue, invalid blob url or network error.',
+                            s,
+                            e
```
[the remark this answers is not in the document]
iremgokce: it is coming from the catch block of loadPromise method which is not just a axios request for the 3d model but something in the babylon loader as well. I removed exposing error in the network error case, but kept the other one as users may want to see the error like in this case:
image

src/Components/3DV/SceneView.styles.ts
```diff
@@ -55,6 +56,15 @@ export const getSceneViewStyles = memoizeFunction((theme: Theme) =>
                 transform: 'translate(-50%, 0%)'
             } as IStyle
         ],
+        networkErrorMessage: [
```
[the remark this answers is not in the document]
iremgokce: because it was for a different type of error representation, but now made it common as Matt asked to use illustration for the other generic type of errors as well.

src/Components/3DV/SceneView.tsx
```diff
@@ -81,35 +83,30 @@ function debounce(func: any, timeout = 300) {
...
+            if (
+                dt.toString() === 'Invalid Date' ||
+                dt.toISOString() === '1970-01-01T00:00:00.000Z'
+            ) {
+                return '';
+            }
+            return dt.toISOString();
+        })
+        .catch((error) => {
+            throw error;
```
ptallettms, line 108: So far as I can tell, this looks a risky change as it promotes the fetch for the modified date to the front line, and fails the load if there's a network error fetching the modified date. I have seen MANY instances where this call fails, but the model load succeeds. I originally had something similar, but backed off that as it failed too many loads.

Please ignore this comment if that is not true.

ptallettms: Also this line is technically wrong, but quite nice in practice.

mod = mod ? '?' + mod : '';

If we don't get modified date, we load without a query string which will never cache bust, so if you update the model, we will never load it if the previous version is in cache.

Now actually the getModifiedDate often fails and its a pain to have to keep reloading the model as most of the time, we haven't changed the model, so the behaviour is wrong, but IMHO desirable. If a customer meets this scenario, they can upload the new model with a new name and change their 3DScenesConfig (but better would be to fix their environment so the getModifiedDate works).

If we want to cache bust if the getModifiedDate fails, then this line could be

mod = mod ? '?' + mod : '?' + new Date().toIsoString();

but I don't like it. I've been meaning to discuss this on standup as I heard a rumour that one of our customers was hitting this.

iremgokce: Looks like the line above `engine.disableManifestCheck = true;` is causing the BABYLON.SceneLoader.Load method not surfacing the network error in its onError callback. I asked it to Babylon people and waiting for their reply if they are interconnected. @ptallettms why do we need disabling manifest check, was it for caching related?

iremgokce:
> Okay there has been a fix from Babylon team to expose the network error in the loader even caching is enabled @ptallettms @darsney: [BabylonJS/Babylon.js#12647](https://github.com/BabylonJS/Babylon.js/pull/12647) whenever they merge, I will bump up the Babylon version in our package.json and move the CORS/network checking logic to the loader instead of getModifiedTime.

**UPDATE**: Babylon team is going to publish the fix with Thursday's release: 5.11.0, then I am going to bump Babylon and move the network error handling logic to loader instead.

src/Components/3DV/SceneView.tsx
iremgokce: Hey @ptallettms what is you final input on this mod related line? Is `mod = '?' + mod || new Date().toIsoString();` better option now?

ptallettms: Whilst technically incorrect, I'd leave it as it is as that is what the user probably intended.

msnyder-msft (approved):

src/Components/3DV/SceneView.tsx
iremgokce: thanks for your reply!

PR #523, opened 2022-06-03
ADTS-163 Changes

JoseZarmada:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/OATGraphViewer/OATGraphViewer.tsx
+++ b/src/Components/OATGraphViewer/OATGraphViewer.tsx
@@ -84,16 +84,20 @@ const getGraphViewerElementsFromModels = (models, modelPositions) => {
                     new ElementEdgeData(
                         `${input['@id']}${OATComponentHandleName}${content['schema']}`,
                         content['name'],
-                        content['name'],
+                        '',
                         OATComponentHandleName
                     )
                 );
                 relationships = [...relationships, componentRelationship];
             } else if (content['@type'] === OATRelationshipHandleName) {
+                const relationshipAmount = relationships.filter(
+                    (element) =>
+                        element.source && element.source === input['@id']
+                ).length;
                 const relationship = new ElementEdge(
                     content['@id']
                         ? content['@id']
-                        : `${input['@id']}${OATRelationshipHandleName};${versionClassBase}`,
+                        : `${input['@id']}${OATRelationshipHandleName};${relationshipAmount}`,
                     OATRelationshipHandleName,
                     input['@id'],
                     OATRelationshipHandleName,
@@ -101,7 +105,7 @@ const getGraphViewerElementsFromModels = (models, modelPositions) => {
                     new ElementEdgeData(
                         content['@id']
                             ? content['@id']
-                            : `${input['@id']}${OATComponentHandleName};${versionClassBase}`,
+                            : `${input['@id']}${OATRelationshipHandleName};${relationshipAmount}`,
                         content['name'],
                         content['displayName'],
                         OATRelationshipHandleName
@@ -283,6 +287,25 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
                 node.id = newId;
                 node.data.id = newId;
                 node.data.name = model['displayName'];
+                node.data.content = propertyItems;
+                setElements([...elements]);
+                currentNodeIdRef.current = newId;
+            } else if (node && node.source) {
+                if (node.sourceHandle === OATComponentHandleName) {
+                    elements.find(
+                        (element) => element.id === node.target
+                    ).data.name = model['name'];
+                }
+                const newId = model['@id'];
+                const propertyItems = model.properties.filter(
+                    (property) =>
+                        typeof property['@type'] === 'object' &&
+                        property['@type'][0] === 'property'
+                );
+                node.id = newId;
+                node.data.id = newId;
+                node.data.name = model['name'];
+                node.data.displayName = model['displayName'];
                 node.data.content = propertyItems;
                 setElements([...elements]);
                 currentNodeIdRef.current = newId;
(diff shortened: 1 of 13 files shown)


src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -84,24 +84,28 @@ const getGraphViewerElementsFromModels = (models, modelPositions) => {
...
+                        '',
                         OATComponentHandleName
                     )
                 );
                 relationships = [...relationships, componentRelationship];
             } else if (content['@type'] === OATRelationshipHandleName) {
+                const relationshipAmount = relationships.filter(
+                    (element) =>
+                        element.source && element.source === input['@id']
+                ).length;
```
jamescarpinter, line 96: I thought the value after the `;` was version?

Shouldn't `@id` be unique? Do we need to add an int here?

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -631,9 +654,19 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
         );
         if (target) {
             if (currentHandleIdRef.current !== OATUntargetedRelationshipName) {
+                const relationshipAmount = elements.filter(
+                    (element) =>
+                        element.source &&
+                        element.source === currentNodeIdRef.current
+                ).length;
```
jamescarpinter, line 661: Will this always generate the right "next" value? I.e. if I add relationship_0 and relationship_1, then remove relationship_0, won't the next relationship end up at relationship_1 again?

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -129,16 +130,20 @@ const OATHeader = ({ elements, dispatch, state }: OATHeaderProps) => {
                     const validJson = await parseModel(content);
                     if (!validJson) {
                         items.push(newItem.content);
+                    } else {
+                        allValidFiles = false;
```
jamescarpinter, line 134: We can break out of the loop at this point I guess?

JoseZarmada: Actually I was asked to look for all the errors 
image
If I break the loop only the first one will be shown

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -129,16 +130,20 @@ const OATHeader = ({ elements, dispatch, state }: OATHeaderProps) => {
...
                     }
                 } catch (error) {
                     console.log(error);
                     alert(error);
```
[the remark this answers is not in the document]
JoseZarmada: Done

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -875,6 +930,16 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
...
     };
 
+    const onNodeMouseEnter = (evt, node) => {
+        currentHoveredRef.current = node.id;
+        setElements([...elements]);
+    };
+
+    const onNodeMouseLeave = (evt) => {
+        currentHoveredRef.current = '';
+        setElements([...elements]);
```
jamescarpinter, line 940: Why do the elements need to be reset in this scenario? They haven't changed?

If we need to trigger a re-render then perhaps we need to make the component on the ref too 🤔

src/Components/OATPropertyEditor/PropertiesModelSummary.tsx
```diff
@@ -50,40 +50,72 @@ export const PropertiesModelSummary = ({
...
                     borderless
                     disabled={!model}
-                    value={model ? model['@id'] : ''}
-                    placeholder={t('OATPropertyEditor.assetId')}
+                    value={
+                        model && model.displayName
+                            ? typeof model.displayName === 'string'
+                                ? model.displayName
+                                : Object.values(model.displayName)[0]
+                            : ''
```
jamescarpinter, line 113: I think this code would be an ideal candidate for a new utils.ts function.

src/Components/OATPropertyEditor/PropertiesModelSummary.tsx
```diff
@@ -50,40 +50,72 @@ export const PropertiesModelSummary = ({
...
+                    <TextField
+                        styles={textFieldStyes}
+                        borderless
+                        disabled={!model}
+                        value={
+                            model && model.name
+                                ? typeof model.name === 'string'
+                                    ? model.name
+                                    : Object.values(model.name)[0]
+                                : ''
```
jamescarpinter, line 88: I think there's a utils.ts function for this operation.

src/Components/OATGraphViewer/OATGraphViewer.tsx
JoseZarmada: I need to make the re-render because the component the mouse event is leaving is indeed changing but that change is not visualized until render is done, what would you recomend

jamescarpinter: If the styling - or layout - is conditional on the value of currentHoveredRef, then it should be a state property rather than a ref. Changes to state will trigger the appropriate update to the layout.

src/Models/Services/Utils.ts
```diff
@@ -412,3 +412,8 @@ export const getStoredEditorModelPositionsData = () => {
     const oatData = getStoredEditorData();
     return oatData && oatData.modelPositions ? oatData.modelPositions : [];
 };
+
+// Returns the string value of a key if the object is not a string already
+export const getStringFromObject = (value) => {
+    return typeof value === 'string' ? value : Object.values(value)[0];
+};
```
jamescarpinter, line 419: There is a similar function already in the PropertyEditor utils file. Perhaps we can promote that this utils file? And use that instead?

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -195,7 +195,7 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
     const graphViewerMinimapStyles = getGraphViewerMinimapStyles();
     const currentNodeIdRef = useRef('');
     const currentHandleIdRef = useRef('');
-    const currentHoveredRef = useRef('');
+    const [currentHovered, setCurrentHovered] = useState(null);
```
jamescarpinter, line 198: This is way cleaner 🎉 👍

jamescarpinter (approved):

PR #527, opened 2022-06-03
Bug fix: Fix deeplinks not respecting layers

### LINKED ISSUE #526: Deeplinks do not respect selected layers in most recent update
**Describe the bug**
Starting in the version released the week of May 16 deeplinks stopped respecting the layers in the URL when navigating to the viewer.

**To Reproduce**
1. Go to the viewer
2. Change default layer selections to select less than ALL of the layers
3. Copy the deeplink
4. Open a new tab and paste the deeplink
5. See that ALL layers get selected instead of the ones from the deeplink

**Expected behavior**
Should respect the layers in the deeplink, if included

msnyder-msft:
### Summary of changes 🔍 
Fixes #526 
[Bug 14628325](https://msazure.visualstudio.com/One/_workitems/edit/14628325): Shortlink does not respect layers that have been toggled off
The layer values in the deeplink were getting ignored with the most recent changes to auto select all layers. This fixes that issue by checking if we have values in the deeplink context on component mount and then respecting those values instead of overriding them.

### Testing 🧪
- [x] Have no deeplink, all layers selected
- [x] Switching scenes will select all available layers
- [x] Changing selected layers by clicking in dropdown works as expected
- [x] Copy deeplink and paste in new tab will load the scene with the layers selected

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/ADT3DViewer/ADT3DViewer.tsx
+++ b/src/Components/ADT3DViewer/ADT3DViewer.tsx
@@ -108,12 +108,16 @@ const ADT3DViewerBase: React.FC<IADT3DViewerProps> = ({
     const prevLayerCount = useRef<number>(-1); // track the count of layers so we know to refresh the selections if the count changes
     const hasUserChangedLayers = useRef<boolean>(false); // need to know once a user makes a selection so we stop auto selecting items
     const prevSceneId = useRef<string>(sceneId); // need to know if user swaps scenes (using scene dropdown) since we won't get remounted
+    const initialDeeplinkLayers = useRef<string[]>(
+        deeplinkState.selectedLayerIds
+    ); // grab the initial layers on mount
 
     // reset the refs when the scene changes
     if (prevSceneId.current !== sceneId) {
         hasUserChangedLayers.current = false;
         prevLayerCount.current = -1;
         prevSceneId.current = sceneId;
+        initialDeeplinkLayers.current = deeplinkState.selectedLayerIds;
     }
 
     const behaviorIdsInScene = useMemo(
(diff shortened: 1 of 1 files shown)


src/Components/ADT3DViewer/ADT3DViewer.tsx
```diff
@@ -108,12 +108,16 @@ const ADT3DViewerBase: React.FC<IADT3DViewerProps> = ({
     const prevLayerCount = useRef<number>(-1); // track the count of layers so we know to refresh the selections if the count changes
     const hasUserChangedLayers = useRef<boolean>(false); // need to know once a user makes a selection so we stop auto selecting items
     const prevSceneId = useRef<string>(sceneId); // need to know if user swaps scenes (using scene dropdown) since we won't get remounted
+    const initialDeeplinkLayers = useRef<string[]>(
```
iremgokce, line 111: nit: for refs we used to use Ref at the end of the name of the variable to differentiate from other state variables but it seems all the useRefs are not using that convention, just dropping here for the future.

src/Components/ADT3DViewer/ADT3DViewer.tsx
```diff
@@ -255,24 +259,30 @@ const ADT3DViewerBase: React.FC<IADT3DViewerProps> = ({
         const noUserUpdate = !hasUserChangedLayers.current;
         const noSelectedLayers = !deeplinkState?.selectedLayerIds?.length;
         prevLayerCount.current = layersInScene.length;
-        if (noUserUpdate && (noSelectedLayers || layerCountChanged)) {
+        const layersSetOnDeeplink = initialDeeplinkLayers.current.length;
```
iremgokce, line 262: nit:
```suggestion
        const layersCountSetOnDeeplink = initialDeeplinkLayers.current.length;
```

iremgokce (approved):

PR #528, opened 2022-06-06
fix: allow all layers to be selected with prop

samjosephgates:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/ADT3DViewer/ADT3DViewer.stories.tsx
+++ b/src/Components/ADT3DViewer/ADT3DViewer.stories.tsx
@@ -362,9 +362,7 @@ export const MockWithSelection = (_args, { globals: { theme, locale } }) => {
 
 export const LayerSelect = (_args, { globals: { theme, locale } }) => {
     const scenesConfig = mockVConfig as I3DScenesConfig;
-    const [selectedLayerIds, setSelectedLayerIds] = useState<string[]>(
-        undefined
-    );
+    const [selectedLayerIds, setSelectedLayerIds] = useState<string[]>(null);
 
     return (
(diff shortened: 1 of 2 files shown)


msnyder-msft (approved):
Didn't test it out but seems self contained and the logic matches how we get "all layers" in the normal flow.

ccrowley96 (approved):

msnyder-msft (changes requested):
Something is funky.

msnyder-msft:
There is a scenario that doesn't seem to work but used to work on the main branch.
1. Load up mock data for outdoor tanks
2. Remove the behaviors from all the layers
3. Go to the viewer
4. See that only Default shows up and is dropdown is disabled.
5. Go back to edit a behavior
6. Add one to a layer and save
7. Switch to Viewer, don't see that layer in the dropdown
8. Switch to builder and back to Viewer, now the layer is auto selected

samjosephgates:
> There is a scenario that doesn't seem to work but used to work on the main branch.
> 
> 1. Load up mock data for outdoor tanks
> 2. Remove the behaviors from all the layers
> 3. Go to the viewer
> 4. See that only Default shows up and is dropdown is disabled.
> 5. Go back to edit a behavior
> 6. Add one to a layer and save
> 7. Switch to Viewer, don't see that layer in the dropdown
> 8. Switch to builder and back to Viewer, now the layer is auto selected

Fixed this issue. I now check that the prop is not undefined as we only want to set the layers if the prop has been set by the consumer. If its set as null we set all the layers as selected.

msnyder-msft (approved):

ptallettms (approved):

PR #529, opened 2022-06-06
fix: limits to camera calculated based on model

samjosephgates:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/3DV/SceneView.tsx
+++ b/src/Components/3DV/SceneView.tsx
@@ -376,10 +376,11 @@ function SceneView(props: ISceneViewProps, ref) {
                         cameraRef.current.zoomOn(meshes, true);
                         cameraRef.current.radius = radius;
                         cameraRef.current.lowerRadiusLimit = 0;
-                        cameraRef.current.upperRadiusLimit =
-                            Math.max(width, height, depth) * 3;
+                        // set upperRadiusLimit to be greater than the starting radius to allow the user to zoom out if they wish
+                        cameraRef.current.upperRadiusLimit = radius * 2;
+                        // set the maxZ of the camera to be higher than the upperRadiusLimit to ensure it will not clip when zoomed out
                         cameraRef.current.maxZ =
-                            Math.max(width, height, depth) * 5;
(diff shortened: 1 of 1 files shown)


src/Components/3DV/SceneView.tsx
```diff
@@ -372,10 +372,14 @@ function SceneView(props: ISceneViewProps, ref) {
                         );
 
                         camera.attachControl(canvas, false);
-                        camera.lowerRadiusLimit = 0;
                         cameraRef.current = camera;
                         cameraRef.current.zoomOn(meshes, true);
                         cameraRef.current.radius = radius;
+                        cameraRef.current.lowerRadiusLimit = 0;
```
msnyder-msft, line 378: can we add a comment to the properties that help with the clipping so we remember?

msnyder-msft: also a note on why *3 and *5?

msnyder-msft (approved):

src/Components/3DV/SceneView.tsx
```diff
@@ -372,10 +372,14 @@ function SceneView(props: ISceneViewProps, ref) {
                         );
 
                         camera.attachControl(canvas, false);
-                        camera.lowerRadiusLimit = 0;
                         cameraRef.current = camera;
                         cameraRef.current.zoomOn(meshes, true);
```
coryboyle, line 376: Not sure if it helps, but if you set the zoomOn bool to false the camera will auto adjust the maxZ to include the bounds of the object.

src/Components/3DV/SceneView.tsx
samjosephgates: Added comments and made the values make a little more sense

src/Components/3DV/SceneView.tsx
samjosephgates: You can set the bool on zoomOn, but that then clips when the user zooms out, so we need to set the MaxZ to be higher than the upperRadiusLimit. Updated code to make it clearer what its doing.

darsney (approved):

PR #530, opened 2022-06-06
Zarmada/working/ADTS-161

miltonzarmada:
### Summary of changes 🔍 
- ADTS-161 Update property pop up 

### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/OATPropertyEditor/PropertyList.tsx
+++ b/src/Components/OATPropertyEditor/PropertyList.tsx
@@ -13,7 +13,10 @@ import {
 } from '../../Models/Constants/ActionTypes';
 import { DTDLProperty, IAction } from '../../Models/Constants/Interfaces';
 import { IOATEditorState } from '../../Pages/OATEditorPage/OATEditorPage.types';
-import { getModelPropertyCollectionName } from './Utils';
+import {
+    getModelPropertyCollectionName,
+    shouldClosePropertySelectorOnMouseLeave
+} from './Utils';
 
 type IPropertyList = {
     currentPropertyIndex: number;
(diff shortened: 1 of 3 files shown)


src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -182,49 +181,55 @@ export const PropertyList = ({
...
+        const boundingRect = e.target.getBoundingClientRect();
+        setPropertySelectorPosition({
+            ...propertySelectorPosition,
+            top: boundingRect.top,
+            left: boundingRect.left
+        });
+        setPropertySelectorTriggerElementsBoundingBox(boundingRect);
+    };
+
+    const getIsMouseLeaveNotNorth = (e) => {
```
jamescarpinter, line 194: This can be simplified... i.e. `return e.clientY >= propertySelectorTriggerElementsBoundingBox.bottom ||
            e.clientX < propertySelectorTriggerElementsBoundingBox.left ||
            e.clientX > propertySelectorTriggerElementsBoundingBox.right;`

jamescarpinter: Also... what exactly is this function doing? I'm not sure if the name is helping 🤔

src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -307,41 +320,43 @@ export const PropertyList = ({
...
-                        !propertyOnHover && (
+                {propertyList && propertyList.length > 0 && (
+                    <div
+                        className={
+                            propertyInspectorStyles.addPropertyBarPropertyListWrap
+                        }
+                        onMouseLeave={(e) => {
+                            if (getIsMouseLeaveNotNorth(e)) {
+                                setPropertySelectorVisible(false);
+                            }
```
jamescarpinter, line 336: It might be cleaner to wrap this logic into a separate function... i.e. `handleOnMouseLeave`. We could use it elsewhere too.

src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -307,41 +320,43 @@ export const PropertyList = ({
...
+                        }}
+                    >
+                        {true && (
```
jamescarpinter, line 339: What's `true` for?

src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -307,41 +320,43 @@ export const PropertyList = ({
...
                             <AddPropertyBar
-                                onMouseOver={() => {
+                                onMouseOver={(e) => {
                                     setPropertySelectorVisible(true);
                                     setLastPropertyFocused(null);
+                                    handleSelectorPosition(e);
```
jamescarpinter, line 344: I'd prefer to extract this logic into a separate function rather than doing it inline.

src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -182,49 +181,55 @@ export const PropertyList = ({
...
+        if (
+            e.clientY >= propertySelectorTriggerElementsBoundingBox.bottom ||
+            e.clientX < propertySelectorTriggerElementsBoundingBox.left ||
+            e.clientX > propertySelectorTriggerElementsBoundingBox.right
+        ) {
+            return true;
+        }
+
+        return false;
+    };
```
jamescarpinter, line 204: Given this function is passed as a prop to another function, we should probably wrap it in a `useCallback` so it only changes when the state property changes.

Or, perhaps a cleaner solution would be to pass the state (i.e. the bounding box) as a prop and define this function outside of the component and export it?

jamescarpinter (changes requested):

jamescarpinter (approved):

PR #541, opened 2022-06-08
Left panel padding, borders & overflow consistency

ccrowley96:
## Summary of changes 🔍 
Addresses: Builder left rail should have right padding that prevents it from being slammed to edge ([ADO item](https://msazure.visualstudio.com/One/_workitems/edit/14615576/))

This PR adds full bleed (edge to edge) borders / separators to the left panel.  To do this, each block on the left panel needs to be padded individually.  Additionally blocks which vertically overflow now have internal padding & show their scroll bar snapped to the right border of the builder left panel.

## Testing 🧪
-  Walk through each screen of the builder & viewer and verify that padding, borders, spacing, & overflow perform as expected.

image

### Checklist ✔️
- [x] Added relevant labels


src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.styles.ts
```diff
@@ -26,7 +26,8 @@ export const getStyles = (
             classNames.wrapper,
             {
                 flexGrow: 1,
-                position: 'relative'
+                position: 'relative',
+                padding: '16px 0px'
```
msnyder-msft, line 30: What would you think of putting the spacing in a constant so we can update all the places in one go?

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/AlertsTab.tsx
```diff
@@ -153,91 +153,95 @@ const AlertsTab: React.FC = () => {
...
-                                LOC_KEYS.notificationPlaceholder
-                            )}
-                        />
-                        <Text styles={{ root: { fontSize: FontSizes.small } }}>
-                            {t(LOC_KEYS.notificationLabelDescription)}{' '}
-                            <i>
-                                {t(
-                                    LOC_KEYS.notificationLabelDescriptionExample
+        <div className={commonPanelStyles.paddedLeftPanelBlock}>
+            <Stack tokens={{ childrenGap: 8 }}>
```
msnyder-msft, line 157: nit: can we just put the class on teh stack?

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/StatusTab.tsx
```diff
@@ -146,48 +146,50 @@ const StatusTab: React.FC<IStatusTabProps> = ({ onValidityChange }) => {
...
-                />
-                {showRangeBuilder && <Separator />}
-            </div>
-            {showRangeBuilder && (
-                <ValueRangeBuilder
-                    valueRangeBuilderReducer={valueRangeBuilderReducer}
-                />
-            )}
-        </Stack>
+        <div className={commonPanelStyles.paddedLeftPanelBlock}>
```
msnyder-msft, line 149: can we put the class on the stack?

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetForm/WidgetForm.styles.ts
```diff
@@ -23,6 +24,7 @@ export const getWidgetFormStyles = memoizeFunction((theme: Theme) => {
         ],
         widgetFormContents: [
             classNames.widgetFormContents,
+            leftPanelBuilderBlock,
```
msnyder-msft, line 27: LOVE IT!! Could we make a common style for the stacks taht I commented on and just drop this in there? That might be even better than putting the class name on each stack directly.

src/Components/BaseComponent/BaseComponent.tsx
```diff
@@ -53,7 +54,7 @@ const BaseComponent: React.FC<BaseComponentProps> = ({
 
     const BaseContents = (
         <div
-            className={`cb-base-component ${
+            className={`${disableDefaultStyles ? '' : 'cb-base-component '} ${
```
msnyder-msft, line 57: oooo.

src/Models/Constants/StyleConstants.ts
```diff
@@ -3,7 +3,7 @@ export const behaviorsModalPreviewContainerLeftOffset =
     builderLeftPanelWidth + 36;
 
 export const VIEWER_HEADER_TOP_OFFSET = 10;
-export const BUILDER_HEADER_TOP_OFFSET = 10;
+export const BUILDER_HEADER_TOP_OFFSET = 26;
```
msnyder-msft, line 6: why did the vertical spacing change for this?

src/Pages/ADT3DScenePage/ADT3DScenePage.styles.ts
```diff
@@ -7,7 +7,7 @@ const classNames = {
     header: `${scenePageClassPrefix}-header`
 };
 
-export const getStyles = memoizeFunction(() => {
+export const getStyles = memoizeFunction((isBuildMode: boolean) => {
```
msnyder-msft, line 10: Minor thing, but can we make it an enum for `mode: 'build'|'viewer'`? Just makes it easier to extend if for some reaosn we need more than 2 states.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Behaviors.tsx
```diff
@@ -184,66 +184,79 @@ const SceneBehaviors: React.FC<Props> = ({
     const customStyles = getStyles(theme);
     return (
         <div className="cb-scene-builder-pivot-contents">
-            <SearchHeader
```
msnyder-msft, line 187: when the scrolling shows up, the right padding feels wrong. I now see what you were talking about before where the scrollbar should be indented instead. @darsney 
image

msnyder-msft: not sure how that same logic works here though. Cause this kinda feels off too. I think they would both be better if they had the padding outside the scroll. Then the scroll is under the search box.
image

msnyder-msft: But as a counter, if we followed the same pattern here, we would have no padding right of the text boxes??? hmmm that doesn't seem great.
image

src/Components/ADT3DSceneBuilder/Internal/Shared/PanelFooter.styles.ts
```diff
@@ -19,6 +19,6 @@ export const PanelFooterStyles = mergeStyleSets({
 export const PanelFooterStackStyles: IStackStyles = {
     root: {
         display: 'flex',
-        padding: `12px 4px ${BUILDER_FOOTER_BOTTOM_OFFSET}px 4px` // separator has weird built in padding on top
+        padding: `12px 16px ${BUILDER_FOOTER_BOTTOM_OFFSET}px 16px` // separator has weird built in padding on top,
```
msnyder-msft, line 22: why did this change? They now seem like they don't match
image

msnyder-msft (approved):
LGTM, I would prefer to move a bunch of those extra divs into class names on the root stacks instead. Adding more DOM elements is less ideal if we already have a div that is wrapping the same contents.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Behaviors.tsx
ccrowley96: Yeah IMO, it's better to snap the scroll bar to the far right & pad the left of the scroll bar.  It looks v jank when the content touches the scroll bar.  I think having some space between the list item hover color & the scroll bar is the best option here.

src/Models/Constants/StyleConstants.ts
ccrowley96: Needed 16 more px of padding, without this, the top controls were misaligned

image

src/Pages/ADT3DScenePage/ADT3DScenePage.styles.ts
ccrowley96: This actually needs to be a boolean, because we're doing multiple checks 
```ts
deeplinkState.mode === ADT3DScenePageModes.BuildScene && state.currentStep === ADT3DScenePageSteps.Scene
```
To get the `isBuildMode` param.  If we were just passing the mode through, I think this makes sense, but keeping as is for now.

src/Components/ADT3DSceneBuilder/Internal/Shared/PanelFooter.styles.ts
ccrowley96: This is from `main`

image

This change only updates the left & right padding in the `padding: top right bottom left` syntax

msnyder-msft: ahhh fair enough. probalby the difference in the 12 & 16 there then.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/TwinsTab.tsx
```diff
@@ -232,71 +232,74 @@ const TwinsTab: React.FC<ITwinsTabProps> = ({
...
-                        '3dSceneBuilder.twinAlias.calloutDescription'
-                    )}
-                    listKey={'common-properties-callout-list'}
-                    listItems={primaryTwinProperties}
-                    isListLoading={isCommonPrimaryTwinPropertiesLoading}
-                    onDismiss={toggleIsPrimaryTwinPropertiesCalloutVisible}
-                    filterPlaceholder={t(
-                        '3dSceneBuilder.twinAlias.searchProperties'
+        <div className={commonPanelStyles.paddedLeftPanelBlock}>
+            <Stack tokens={{ childrenGap: 12 }}>
```
msnyder-msft, line 236: can we move the class name here too?

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/WidgetsTab.tsx
```diff
@@ -147,12 +147,13 @@ const WidgetsTab: React.FC = () => {
...
     const actionButtonStyles = getActionButtonStyles(theme);
     return (
-        <>
-            <Stack tokens={{ childrenGap: 8 }}>
-                <Text className={commonPanelStyles.text}>
-                    {t(LOC_KEYS.tabDescription)}
-                </Text>
-                <div className={commonPanelStyles.formTabContents}>
+        <div className={commonPanelStyles.formTabContents}>
+            <div className={commonPanelStyles.paddedLeftPanelBlock}>
```
msnyder-msft, line 151: can we move teh padding to the stack here too? I don't think we really need the padding around the dialog right?

src/Components/ADT3DSceneBuilder/Internal/Elements/Internal/AliasedTwinsTab.tsx
```diff
@@ -65,27 +65,31 @@ const AliasedTwinsTab: React.FC = () => {
...
-            <div>
-                {twinAliasList.length === 0 ? (
-                    <div className={commonPanelStyles.noDataText}>
-                        {t('3dSceneBuilder.elementFormTwinTab.noDataMessage')}
-                    </div>
-                ) : (
-                    <>
-                        <Text className={commonPanelStyles.text}>
+        <div className={commonPanelStyles.paddedLeftPanelBlock}>
+            <Stack tokens={sectionStackTokens}>
```
msnyder-msft, line 69: can we move this class name as well?

src/Models/Constants/StyleConstants.ts
msnyder-msft: but why? what changed vertically that threw this off? They used to be aligned just fine.

msnyder-msft: now this spacing isn't going to match between builder & viewer mode right? so the controls are going to bounce around.

msnyder-msft: bouncing

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.styles.ts
```diff
@@ -39,7 +39,7 @@ export const getStyles = (
                     alignItems: 'center',
                     position: 'absolute',
                     right: 0,
-                    top: BUILDER_HEADER_TOP_OFFSET,
+                    top: BUILDER_CONTROLS_TOP_OFFSET,
```
msnyder-msft, line 42: I still don't see why this is changing just to add borders and horizontal padding on the left panel items. What is giving us the vertical changes?

src/Components/ADT3DSceneBuilder/Internal/Shared/PanelFooter.styles.ts
```diff
@@ -1,5 +1,4 @@
 import { mergeStyleSets, IStyle, IStackStyles } from '@fluentui/react';
-import { BUILDER_FOOTER_BOTTOM_OFFSET } from '../../../../Models/Constants/StyleConstants';
```
msnyder-msft, line 2: probs can delete this constant now?

msnyder-msft (approved):

msnyder-msft:
There is still a 2-4px shift left->right when switching from builder to viewer mode. 
image

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.styles.ts
msnyder-msft: And it needs to be local so the right border goes fully to the top?

msnyder-msft (approved):

src/Components/ADT3DSceneBuilder/Internal/Shared/PanelForms.styles.ts
```diff
@@ -15,7 +16,7 @@ const classNames = {
     pivot: `${classPrefix}-pivot`
 };
 export const getPanelFormStyles = memoizeFunction(
-    (_theme: Theme, formHeaderHeight: number) => {
+    (_theme: Theme, formHeaderHeight: number, padContent = false) => {
```
iremgokce, line 19: nit: changing the name to reflect it is a bool
```suggestion
    (_theme: Theme, formHeaderHeight: number, shouldPadContent = false) => {
```

iremgokce (approved):
left minor comments, looks good to me!

PR #544, opened 2022-06-09
Auto fill new twin alias or behavior name with searched text in callout

iremgokce:
### Summary of changes 🔍 
[PBI Task: ](https://msazure.visualstudio.com/One/_sprints/taskboard/ADT%20ExACT/One/Custom/IoT/2206?workitem=14336940)There was a request when user is editing a behavior and they're trying to add twin aliases, they will try to search for an alias they want. If the alias doesn't exist, then they'll select "create twin alias". Whichever string they typed in the search box should automatically populate as the Alias name.

UPDATE: based on the feedback, the logic is changed in a way that we always prepopulate even there are search results, and apply it to other callouts where there is search and primary action is create.
UPDATE: added the same logic to add behavior callout in element form

image
image
image
image

### Testing 🧪
You can test it in ADT3DSceneBuilder -> select a behavior -> Twins tab -> Add twin -> search text and see if there is _no result_ click on Create and see the BehaviorTwinAliasFor has the search text populated as the name

### Checklist ✔️
- [x] Added relevant labels


src/Components/CardboardListCallout/CardboardListCallout.tsx
```diff
@@ -127,7 +127,16 @@ const CardboardListCallout = <T extends unknown>({
...
                         data-testid={dataButtonTestId}
                         styles={cardboardListCalloutPrimaryButtonStyles}
-                        onClick={primaryActionProps.onPrimaryActionClick}
+                        onClick={(_event) => {
+                            // when where is no search results and if expose search term with primary action is set,
+                            // dispatch the search text within onPrimaryClick action to be consumed for a purpose by the handler
+                            primaryActionProps.isSearchTermExposedWithPrimaryActionClick &&
+                            filteredListItems.length === 0
+                                ? primaryActionProps.onPrimaryActionClick(
+                                      searchText
```
msnyder-msft, line 136: can we not simply always pass the search text? Why only do this when the results are 0? If they are clicking create after searching, wouldn't the same assumptions hold that they did not find what they were looking for?

src/Components/CardboardListCallout/CardboardListCallout.types.ts
```diff
@@ -52,6 +52,7 @@ export interface ICardboardListCalloutProps<T = void> {
 
 type PrimaryActionProps = {
     primaryActionLabel: string;
-    onPrimaryActionClick: () => void;
+    onPrimaryActionClick: (searchTerm?: string) => void;
```
msnyder-msft, line 55: can't we just always send hte value and the consumer can decide whether to use the value? after all, they pass in the handler for this click action.

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.types.ts
```diff
@@ -153,6 +153,7 @@ export type BehaviorTwinAliasFormInfo = null | {
     twinAlias: IBehaviorTwinAliasItem;
     mode: TwinAliasFormMode;
     twinAliasIdx?: number;
```
msnyder-msft, line 155: how can I get the search box to show up? when I edit a behavior I dont' get the option to search.
image

msnyder-msft:
please link the work item in the description.

src/Components/CardboardListCallout/CardboardListCallout.tsx
```diff
@@ -127,7 +127,16 @@ const CardboardListCallout = <T extends unknown>({
                     <PrimaryButton
```
msnyder-msft, line 127: please add jest tests for this component to validate this new functionality. You can look at `ViewerConfigUtility.ts` for examples.

msnyder-msft (commented):
2 things. 
1. Please add unit tests. If we don't start adding them for small changes we'll never end up with any tests. And this is the kind of thing we don't want to have to do manual tests on and is likely to be overlooked if it's broken at some point.
2. Let's think about whether we can simplify the props to always pass the search text in the action and let the caller decide whether to use it. You could include the count of searched results if you want to let them make a more informed decision but that seems more flexible long term and doesn't require the extra boolean prop.

src/Components/CardboardListCallout/CardboardListCallout.types.ts
iremgokce: I think the primary action and search term are not tightly related, that is why I added isSearchTermExposedWithPrimaryActionClick prop to expose the search term in onPrimaryActionClick only when it is set. I don't think we should always send the search term when we click on primary action click (e.g. cases where the primary action is irrelevant like learn more).

msnyder-msft: there won't be a search term in the case that it's Learn More though, cause there won't be a searchbox shown right? Also, the consumer would know they dont' care about the value and can simply ignore the argument. You get the flexibility without any extra complexity of more props and edge cases on when you should expect an argument value or not.

src/Resources/Locales/en.json
```diff
@@ -514,9 +514,8 @@
         "notImplemented": "Not implemented",
         "noPrimaryTwinProperties": "No primary twin properties to display",
         "twinAlias": {
-            "title": "Twin alias",
```
msnyder-msft, line 517: please force a rev of the other languages too.

iremgokce: I will I will :)

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.types.ts
iremgokce: good catch

src/Components/CardboardListCallout/CardboardListCallout.tsx
```diff
@@ -88,18 +88,17 @@ const CardboardListCallout = <T extends unknown>({
                 {description && (
                     <Text className={styles.description}>{description}</Text>
                 )}
-                {listItems.length > 0 && (
```
msnyder-msft, line 91: was there a case where we don't want search to show? 🤔 I was surprised it wasn't an option. But maybe we don't need it? I guess there is not much reason to have one, but in other places in the app we always show it even when the list is empty so this is good.

iremgokce: no, we should have always showed the searchbox

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.types.ts
iremgokce: good catch, fixed

src/Components/CardboardListCallout/CardboardListCallout.tsx
iremgokce: can you point me where exactly in ViewerConfigUtility.ts? or did you mean like jest test as in PropertyInspectorModel.test.ts?

msnyder-msft: yeah. like that. if you look at the `ViewerConfigUtility.test.ts` you'll see the tests I wrote for that file.

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.types.ts
```diff
@@ -261,7 +261,7 @@ export interface IADT3DSceneBuilderElementsProps {
         isSelected: boolean
     ) => void;
     clearSelectedElements?: () => void;
-    onCreateBehaviorClick?: () => void;
```
msnyder-msft, line 260: the arg here would be `initialBehaviorName`, it might get filled with the search term value but as far as this function is concerned, it should be provided to fill in the name.

msnyder-msft (approved):

ccrowley96 (approved):
lgmt!  Great work Irem, experience is 🔥

src/Components/CardboardListCallout/CardboardListCallout.tsx
iremgokce: Will be separate PR, working on it

PR #547, opened 2022-06-09
Bug fix: Fix the materials on default mode

msnyder-msft:
### Summary of changes 🔍 
When the default theme is selected, the model materials were painted white. This restores the materials when on the default style.

After
image

Before
image

### Testing 🧪
Switch the themes all around with different color & style combinations. Switching between different styles and checking the style gets restored on app reload. 

### Checklist ✔️
- [ ] Linked associated issue (if present)
- [ ] Added relevant labels
- [ ] Requested developer reviews
- [ ] Request UI review from design / PM
- [ ] Verify all code checks are passing

PR #548, opened 2022-06-09
Encapsulate theme object colors inside the context instead of forcing consumers to deal with it

msnyder-msft:
### Summary of changes 🔍 
Minor polish changes that weren't worth muddying up the #547 PR
Improving the consumption pattern so that the logic is contained inside the context instead of forcing consumers to do it.

No visual changes, purely moving where the state and logic is managed.
image

image

### Testing 🧪
- [x] Switch themes and colors in different combinations
- [x] Try reloading the page and make sure the theme is persisted
 
### Checklist ✔️
- [x] Added relevant labels


src/Components/3DV/SceneViewWrapper.tsx
```diff
@@ -196,30 +166,30 @@ const SceneViewWrapper: React.FC<ISceneViewWrapperProps> = (props) => {
         wrapperMode === WrapperMode.Builder
             ? 'cb-sceneview-builder-wrapper'
             : 'cb-sceneview-viewer-wrapper';
-
+    console.log('****Wrapper', sceneThemeState);
```
samjosephgates, line 169: Do we need to keep this console.log?

samjosephgates (approved):

src/Components/3DV/SceneViewWrapper.tsx
msnyder-msft: nope. good catch.

src/Models/Context/SceneThemeContext/SceneThemeContext.tsx
```diff
@@ -43,10 +44,9 @@ export const SceneThemeContextReducer: (
         );
         switch (action.type) {
             case SceneThemeContextActionType.SET_OBJECT_COLOR: {
-                draft.objectColor =
-                    action.payload.color ||
-                    draft.objectColorOptions?.[0]?.color ||
-                    '';
+                draft.objectColor = draft.objectColorOptions.find(
+                    (x) => x.color === action.payload.color
```
pasanchMSFT, line 48: should there be a fallback option here in case the color is not found in the payload?

src/Models/Context/SceneThemeContext/SceneThemeContext.tsx
```diff
@@ -78,10 +80,9 @@ export const SceneThemeContextReducer: (
                 break;
             }
             case SceneThemeContextActionType.SET_SCENE_BACKGROUND: {
-                draft.sceneBackground =
-                    action.payload.color ||
-                    draft.sceneBackgroundOptions?.[0]?.color ||
-                    '';
+                draft.sceneBackground = draft.sceneBackgroundOptions.find(
+                    (x) => x.color === action.payload.color
```
pasanchMSFT, line 84: Same for this, does this need a fallback color?

pasanchMSFT (approved):
Left 2 comments about null-checking, otherwise looks good to me.

src/Models/Context/SceneThemeContext/SceneThemeContext.tsx
msnyder-msft: hmm. I mean ideally yeah. That only really comes into play if consumers customize and somehow botch it. I could fallback to the default color object I guess.

msnyder-msft: added

PR #553, opened 2022-06-10
Zarmada/working/adts 189

JoseZarmada:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -57,6 +57,15 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
...
     }, [data]);
 
+    useEffect(() => {
+        if (!state.model) {
+            onNameBlur();
+        }
+        if (state.model && state.model['@id'] !== data.id) {
+            onNameBlur();
+        }
+    }, [state]);
```
jamescarpinter, line 67: Can't we combine these `if` statements with an `||`?

Aslo, can we make this effect dependent on model, rather than the whole state? I.e. extract model `const { model } = state;` and then use that as the dependency.

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -604,31 +613,12 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
...
-                                            color:
-                                                theme.semanticColors.actionLink,
-                                            marginTop: '-35px',
-                                            marginRight: '-10px'
-                                        }
-                                    }}
-                                />
-                            </ActionButton>
-                        )}
-                    </body>
```
jamescarpinter, line 631: Why has this been removed?

jamescarpinter (changes requested):

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
JoseZarmada: The Idea of having the Save button Icon was because when we added the cancel Icon to delete the relationship we lost the ability to use the onBlur of the text field since clicking the X was dispatching the onBlur after the delete and the state was not being properly set, I Fix that and using the Model state as indicator of focus and that way we can save the relationship name without the use of the button

jamescarpinter (approved):

PR #554, opened 2022-06-10
ADTS-183: manually calculate the source and target for edges

jamescarpinter:
### Summary of changes 🔍 
* Workaround bug where edge layout is wrong after edit
* Fix restoring saved positions for untargeted relationships
* Ensure edges connect from center point of node

### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/OATGraphViewer/Internal/OATGraphCustomNode.tsx
+++ b/src/Components/OATGraphViewer/Internal/OATGraphCustomNode.tsx
@@ -1,6 +1,6 @@
-import React, { useState, useContext } from 'react';
+import React, { useState, useContext, useEffect } from 'react';
 import { Icon, ActionButton, TextField, Label } from '@fluentui/react';
-import { Handle, removeElements } from 'react-flow-renderer';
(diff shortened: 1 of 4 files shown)


miltonzarmada (approved):

src/Components/OATGraphViewer/Internal/OATGraphCustomNode.tsx
```diff
@@ -171,20 +171,22 @@ const OATGraphCustomNode: React.FC<IOATGraphCustomNodeProps> = ({
 
     return (
         <>
-            <Handle
-                type="target"
-                position="top"
-                className={graphViewerStyles.handle}
-                isConnectable={isConnectable}
-            />
+            {data.type === OATUntargetedRelationshipName && (
```
JoseZarmada, line 174: Why are we having this handle just for the Untargeted Relationship? I'm aware we stop needing it since you are reading the target position on the Custom edge from the state instead the props but why still need it in this case.

PR #555, opened 2022-06-10
Zarmada/working/ADTS-181

miltonzarmada:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Components/OATPropertyEditor/PropertiesModelSummary.tsx
```diff
@@ -84,21 +127,59 @@ export const PropertiesModelSummary = ({
...
+            } else {
+                setIdValidDTMIError(true);
+            }
+        } else {
+            setIdLengthError(true);
+        }
+    };
+
+    const handleDisplayNameChange = (value) => {
+        //Check length
```
jamescarpinter, line 161: Add a space between `//` and the start of the comment.

src/Components/OATPropertyEditor/PropertiesModelSummary.tsx
```diff
@@ -40,41 +40,84 @@ export const PropertiesModelSummary = ({
...
+    );
+    const [name, setName] = useState(model && model.name ? model.name : '');
+    const [id, setId] = useState(model && model['@id'] ? model['@id'] : '');
 
-            setDisplayNameError(null);
-        } else {
-            setDisplayNameError(true);
+    useEffect(() => {
+        if (model) {
+            setDisplayName(model.displayName);
```
jamescarpinter, line 69: What happens if this display name is localized? Do we need to do something different here or will it figure it out? 🤔

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -736,6 +737,16 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
...
+                let relationshipAmount = 0;
+                // Increment relationship amount for every item with the same type of relationship
+                elements.forEach((element) => {
+                    if (
+                        element.type === ModelTypes.relationship &&
+                        element.sourceHandle === OATRelationshipHandleName
+                    ) {
+                        relationshipAmount++;
+                    }
+                });
```
jamescarpinter, line 749: This would be a good use case for `.reduce` (rather than `.forEach`).

There is similar logic elsewhere to calculate `relationshipAmount`. Can we reuse it?

src/Components/OATGraphViewer/Internal/OATGraphCustomNode.tsx
```diff
@@ -136,7 +159,13 @@ const OATGraphCustomNode: React.FC<IOATGraphCustomNodeProps> = ({
     const onIdBlur = () => {
         setIdEditor(false);
         const prevId = data.id;
-        if (data.id !== idText) {
+        // Only apply change if value is error free
+        if (
+            data.id !== idText &&
```
jamescarpinter, line 164: Do you know if IDs are case sensitive? If not, then we might need to do a case insensitive comparison here.

jamescarpinter (changes requested):

src/Components/OATGraphViewer/Internal/OATGraphCustomNode.tsx
miltonzarmada: @jamescarpinter [DTMI](https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/dtdlv2.md#digital-twin-model-identifier) mentions `Equivalence of DTMIs is case-sensitive.`

jamescarpinter (approved):

PR #564, opened 2022-06-16
ADTS-201 ADTS-202

JoseZarmada:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Components/OATModelList/OATModelList.tsx
```diff
@@ -103,10 +106,13 @@ const OATModelList = ({
...
             });
+
             dispatch({
-                type: SET_OAT_DELETED_MODEL_ID,
-                payload: id
+                type: SET_OAT_CONFIRM_DELETE_TYPE,
+                payload: SET_OAT_DELETED_MODEL_ID
             });
+            dispatch({ type: SET_OAT_CONFIRM_DELETE_PAYLOAD, payload: id });
+            dispatch({ type: SET_OAT_CONFIRM_DELETE_OPEN, payload: true });
```
jamescarpinter, line 115: Rather than having to dispatch three times, can we please dispatch once with an object with all of the required properties?

src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -221,6 +247,100 @@ export const PropertyList = ({
         handleSelectorPosition(e);
     };
 
+    const listItemOnClick = (item) => {
+        setCurrentPropertyIndex(item.index);
+        setModalOpen(true);
+        setModalBody(FormBody.property);
+    };
```
jamescarpinter, line 254: Could we define this funciton inside of `getListItems`?

I.e. so we don't need to pass it as a parameter?

src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -67,8 +81,15 @@ export const PropertyList = ({
...
         setPropertySelectorTriggerElementsBoundingBox
     ] = useState(null);
+    const [propertyListItems, setPropertyListItems] = useState<
+        ICardboardListItem<IOATProperty>[]
+    >([]);
     const { model, templates } = state;
 
+    useEffect(() => {
+        getListItems(listItemOnClick);
+    }, [propertyList]);
```
jamescarpinter, line 91: Could we create propertiesListItems as a `useMemo` instead?

It looks like its value is determined by changes to `propertyList` only?

src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -221,6 +247,100 @@ export const PropertyList = ({
...
+                {
+                    key: 'edit',
+                    text: t('OATPropertyEditor.addToTemplate'),
+                    iconProps: {
+                        iconName: 'Add',
+                        className:
+                            propertyInspectorStyles.propertySubMenuItemIcon
+                    },
+                    onClick: () => handleTemplateAddition(item),
+                    id: `addOverflow`,
```
jamescarpinter, line 296: Why are we using back ticks here? Rather than normal quotes?

src/Pages/OATEditorPage/Internal/OATConfirmDeleteWrapper.tsx
```diff
@@ -0,0 +1,67 @@
...
+                    <PrimaryButton
+                        onClick={() => {
+                            dispatch({
+                                type: confirmDeleteType,
+                                payload: confirmDeletePayload
+                            });
+                            dispatch({
+                                type: SET_OAT_CONFIRM_DELETE_OPEN,
+                                payload: false
+                            });
```
jamescarpinter, line 54: I would suggest invoking a callback here.

I.e. when the caller dispatches to open the confirm dialog, they include a callback function to invoke if the user confirms deletion.

jamescarpinter (changes requested):

src/Pages/OATEditorPage/OATEditorPage.types.ts
```diff
@@ -29,5 +34,5 @@ export interface IOATEditorState {
     edge?: string;
     confirmDeleteType?: string;
     confirmDeletePayload?: any;
```
jamescarpinter, line 36: Do we still need the two above properties?

src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -204,12 +200,16 @@ export const PropertyList = ({
         setLastPropertyFocused(null);
         const newModel = deepCopy(model);
         newModel[propertiesKeyName].splice(index, 1);
+        const dispatchDelete = () => {
+            dispatch({
+                type: SET_OAT_PROPERTY_EDITOR_MODEL,
+                payload: newModel
```
jamescarpinter, line 206: Can you help me understand what's going on here? Why would setting the model delete it?

src/Components/OATModelList/OATModelList.tsx
```diff
@@ -102,17 +102,20 @@ const OATModelList = ({
...
         if (!modified) {
+            const dispatchDelete = () => {
+                dispatch({
+                    type: SET_OAT_PROPERTY_EDITOR_MODEL,
+                    payload: null
+                });
+                dispatch({
+                    type: SET_OAT_DELETED_MODEL_ID,
+                    payload: id
+                });
```
jamescarpinter, line 113: Let's make the SET_OAT_DELETED_MODEL_ID dispatch set the property editor model state to null. So we don't need to do two dispatches here.

jamescarpinter (changes requested):

src/Pages/OATEditorPage/OATEditorPage.types.ts
JoseZarmada: Already deleted everything related to those

src/Components/OATPropertyEditor/PropertyList.tsx
JoseZarmada: The newModel has the property deleted before updating the object again, this is the way Milton implemented the original logic I'm just adding the confirmation dialog before deletion

src/Components/OATModelList/OATModelList.tsx
JoseZarmada: Done

jamescarpinter (approved):

src/Components/OATPropertyEditor/PropertyList.tsx
jamescarpinter: Ah! I see. I thought we were deleting the model. We're just deleting a property from the model.

PR #567, opened 2022-06-17
Fix: Avoid camera issue when scene changes while loading

pasanchMSFT:
### Summary of changes 🔍 
Modified logic after model load promise is successful in SceneView to not do anything in case the user navigated away while it was still loading. Ideally the Promise can be cancelled by an `AbortController` (https://developer.mozilla.org/en-US/docs/Web/API/AbortController), however that proved to be problematic and might be overkill if this component's logic is planned to be re-written.

### Testing 🧪
- Open Mock 3D Scene Page story
- Navigate to Luciana 1 scene
- While the model loads, change to another scene
- Validate the correct model is loading and previous model doesn't show up in screen at all.

Chromatic link: https://pasanch-scene-change--601c6b2fcd385c002100f14c.chromatic.com/?path=/story/pages-adt3dscenepage--mock-3-d-scene-page

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/3DV/SceneView.tsx
+++ b/src/Components/3DV/SceneView.tsx
@@ -919,7 +919,9 @@ function SceneView(props: ISceneViewProps, ref) {
                 }
             );
 
-            if (success) {
+            // TODO: Wrap above promise in an AbortController to cancel in case of changing before the promise resolves
+            // Below line of code works, but it does not cancel the download of the first model, which we already navigated away from
+            if (success && modelUrl === modelUrlRef.current) {
                 sceneRef.current = sc;
 
                 preProcessMeshesOnLoad();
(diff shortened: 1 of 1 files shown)


src/Components/3DV/SceneView.tsx
```diff
@@ -919,7 +919,9 @@ function SceneView(props: ISceneViewProps, ref) {
                 }
             );
 
-            if (success) {
+            // TODO: Wrap above promise in an AbortController to cancel in case of changing before the promise resolves
+            // Below line of code works, but it does not cancel the download of the first model, which we already navigated away from
```
msnyder-msft, line 923: you might mention "what works"

msnyder-msft (approved):

ptallettms (approved):

PR #573, opened 2022-06-21
Switch GUID generation to using a library 

### LINKED ISSUE #572: Switch from using custom GUID generation to UUID package
**Is your feature request related to a problem? Please describe.**
Today we use some custom logic to generate GUIDs in the Utils.ts file. This is harder to mock and also, why should we be in this business?
image

**Describe the solution you'd like**
There is a library called UUID which has been used in all of my previous projects for GUID generation. We should be able to call their `v4` method that returns back a unique GUID.

Then the mocking is as simple as this
image

msnyder-msft:
### Summary of changes 🔍 
Switching to use a library `uuid` instead of generating our own GUID generation logic.

Fixes #572

### Testing 🧪
Create new elements, behaviors, scenes. Anything that would generate a GUID and store it to the config file.

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Models/Services/Utils.ts
+++ b/src/Models/Services/Utils.ts
@@ -23,6 +23,7 @@ import ViewerConfigUtility from '../Classes/ViewerConfigUtility';
 import { createParser, ModelParsingOption } from 'azure-iot-dtdl-parser';
 import { IDropdownOption } from '@fluentui/react';
 import { isConstant, toConstant } from 'constantinople';
+import { v4 } from 'uuid';
 
 let ajv: Ajv = null;
 const parser = createParser(ModelParsingOption.PermitAnyTopLevelElement);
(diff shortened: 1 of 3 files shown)


src/Models/Services/Utils.ts
```diff
@@ -54,6 +55,10 @@ export const validate3DConfigWithSchema = (
 };
 
 export const createGUID = (isWithDashes = false) => {
+    // switch to using a library when we don't need the dashes, but fallback to the original approach when we need it.
+    if (!isWithDashes) {
```
pasanchMSFT, line 59: When will we not need the dashes? Why not call v4 when we need a non-dash guid instead of modifying this method?

msnyder-msft: By default we don't use dashes. I think the only place we use dashes is on one spot taht calls an API.

darsney, line 59: does uuid support adding dashes via a param (e.g. v4(true) or something)

PR #585, opened 2022-06-23
ResourcePicker component and changes in EnvironmentPicker

iremgokce:
### Summary of changes 🔍 
[Azure resource loader for storage container dropdown.docx](https://microsoft.sharepoint.com/:w:/r/teams/Azure_IoT/IoTPlat/Shared%20Documents/Digital%20Twins/36.%20Digital%20Twins%20-%20ExACT%20Team/3D%20Scenes/DevDocs%20(Notion%20cont.)/Feature%20reviews/Azure%20resource%20loader%20for%20storage%20container%20dropdown.docx?d=w2bb53e67bb4f4e258170db431920b3e5&csf=1&web=1&e=P3N0Yv)
[ADO item](https://msazure.visualstudio.com/One/_sprints/taskboard/ADT%20ExACT/One/Custom/IoT/2207?workitem=14600665)

- Added `ResourcePicker` standalone component to pull any type of Azure Resources with a given set of required permissions
   - Added **accountId** to the storage secret file, so please add your own storage account id in your user secret file to test the ResourcePicker with Storage Blob Container resource type. It is required to pull containers more efficiently using **searchParams**, rather than making thousands of calls by pulling all the subscriptions -> storage accounts in all those subscriptions -> storage containers in all those storage accounts -> then checking role assignments for all those?!!
   - Added **take** search parameter to limit the number of resources to check the role definitions for to limit the number of requests being made not to crash the page. The downsize of it is that, when we take first "x" number of resources and check the required permissions against those, the ones that we have access may not be in that batch. _**Please feel free to give any feedback on this.**_ 
- Added `getResourcesByPermissions` method to **AzureManagementAdapter** and revised already existing methods there to clean up the code.
- Consumed ResourcePicker components in the `EnvironmentPicker` (1 for ADT Instances, 2 for Azure Storage Containers)
   - **Using 2 step resource picking for Azure Storage Blob Containers** since there is no other efficient way of fetching containers besides **selecting the storage account first**. Otherwise, we have to make thousands of calls depending on the number of subscriptions and storage accounts of the logged in user. For this purpose, added the storage accounts key to the localstorage to get the account id initially to fetch the containers on mount, otherwise the second resource picker is waiting for the storage accounts resource picker finish and get the id from its results to refetch the containers based on that id.

image
image
image
image

### Testing 🧪
- Make sure you use first party app secrets (clientId and redirectUri) to use Azure Management API
- Add your own storage account id in your user secret file's "storage" section (see the placeholder file) to test the ResourcePicker with Storage Blob Container resource type
- Use ResourcePicker component's stories and EnvironmentPicker in the ADT3DScenePage
- Enable `debugLogging` in AzureManagementAdapter to see the logs (number of fetched resources etc.)

### Checklist ✔️
- [x] Added relevant labels


src/Components/ResourcePicker/ResourcePicker.types.ts
```diff
@@ -0,0 +1,69 @@
...
+    AzureResourceDisplayFields,
+    AzureResourceTypes
+} from '../../Models/Constants/Enums';
+import {
+    IAzureManagementAdapter,
+    IAzureResource,
+    IAzureResourceSearchParams
+} from '../../Models/Constants/Interfaces';
+
+export interface IResourcePickerProps {
```
msnyder-msft, line 20: let's sort the props please.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                            selectedOption.text
+                    );
+                }
+            }
+            return options;
+        },
+        []
+    );
+
+    const optionsFromResources = () => {
```
msnyder-msft, line 217: nit: please add a verb to the start of this name. `getOptionsFromResources`

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+            const resources: Array<IAzureResource> =
+                resourcesState.adapterResult.result?.data;
+            resourcesRef.current = resources;
+
+            if (onResourcesLoaded) {
+                onResourcesLoaded(resources);
+            }
+
+            newOptions = mergeOptionsWithProps(
+                optionsFromResources(),
```
msnyder-msft, line 282: can we please pass in the data to this function taht it is going to use? Setting the data to a ref a few lines before calling a function and then using that internally feels pretty dirty. If you need the ref for some other reason, at least pass in the value of the ref instead of using it internally. It will make it much more clear what dependencies this function has.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                }
+            } else {
+                return resource[AzureResourceDisplayFields[displayField]];
+            }
+        },
+        [displayField]
+    );
+
+    // after fetching the resources merge them with existing options in the dropdown
+    const mergeOptionsWithProps = useCallback(
```
msnyder-msft, line 163: can this get pulled out to module scope? It seems like it can take everything via props and that would simplify this whole flow of nested functions.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                // do the merging with existing options: add the additional options manually entered by user to the end if it is not in the fetched data
+                const existingOptionTexts =
+                    resourcesRef.current?.map((resource) =>
+                        getDisplayFieldValue(resource)
+                    ) || [];
+
+                const optionsToAdd = additionalOptions.filter(
+                    (additionalOption) =>
+                        existingOptionTexts.findIndex((existingOptionText) =>
+                            isValuesEqual(
```
msnyder-msft, line 175: nit: naming. `areEqual` or `areValuesEqual`

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                            option.itemType ===
+                                SelectableOptionMenuItemType.Header &&
+                            option.text === freeformOptionsHeaderText
+                    );
+                    if (!freeFromOptionsHeaderExist) {
+                        options.push(freeformOptionsHeader);
+                    }
+                    options = options.concat([selectedOption]);
+                } else {
+                    setSelectedKey(
```
msnyder-msft, line 206: it is confusing to me that in a function called "mergeOptionsWithProps" that we are setting the selected key. Why do we have to do that? Why would the key change when generating the list of items? Should't it match between the old and new list of options?

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                    if (resourcesRef.current?.length > 0) {
+                        // add the freeform options header if the resources are fetched and exists
+                        options.push(freeformOptionsHeader);
+                    }
+                    options = options.concat(optionsToAdd);
+                }
+            }
+            if (selectedOption) {
+                const selectedOptionInNewOptions = options.find((option) =>
+                    isValuesEqual(option.text, selectedOption.text)
```
msnyder-msft, line 192: why do we find by matching the text instead of using a key/value property? Couldn't that lead to issues where we have manipulated a display value? Ex: adding https: or something to the start.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+        if (selectedOption) {
+            const selectedKey = options?.find(
+                (option) =>
+                    option.text === parsedOptionText(selectedOption.text)
+            )?.key;
+            setSelectedKey(selectedKey);
+        } else {
+            setSelectedKey(null);
+        }
+    }, [selectedOption]);
```
msnyder-msft, line 325: this hsould be dependent on `options` as well.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                    return 'resourcesPicker.select';
+            }
+        } else {
+            if (options.length) {
+                return t('resourcesPicker.select');
+            } else {
+                return t('resourcesPicker.noOption');
+            }
+        }
+    }, [resourceType, t, options.length]);
```
msnyder-msft, line 125: check you dependency list please. Enable the `"react-hooks/exhaustive-deps": "warn",` rule in your `.eslintrc` and fix all of them in this file.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+            additionalOptionsProp?.map((aO) => ({
+                key: aO,
+                text: aO,
+                styles: comboBoxOptionStyles
+            })) || []
+        );
+    }, [additionalOptionsProp]);
+
+    useEffect(() => {
+        setOptions(
```
msnyder-msft, line 338: this nesting of functions feels wrong. They each have many internal side effects and logics. Not sure it can be helped at this point with a lot of churn, but wanted to call it out.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+            mergeOptionsWithProps(
+                optionsFromResources(),
+                additionalOptions,
+                selectedOption
+            )
+        );
+    }, [additionalOptions]);
+
+    const parsedOptionText = useCallback((value) => {
```
msnyder-msft, line 347: naming nit: `parseOptionText` or `sanitizeOptionText`

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+        if (!value) return value;
```
msnyder-msft, line 348: Can we pull this out to a helper function at module scope and just pass in the resource type and display field props?

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                    newVal = 'https://' + newVal;
+                }
+            }
+            return newVal;
+        } else {
+            return value;
+        }
+    }, []);
+
+    const isValidUrlStr = useCallback((urlStr: string) => {
```
msnyder-msft, line 379: let's pull this out to a helper function. It seems self contained enough and will simplify the render loop.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+            }
+            return value1 === value2;
+        } else {
+            return value1 === value2;
+        }
+    };
+
+    const handleOnChange = useCallback(
+        (option, value) => {
+            if (option) {
```
msnyder-msft, line 438: can we add some comments to this `onChange` function? Like why would we ever not have an option? And if we don't how can we make the assumption to parse the text?

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                        const freeFromOptionsHeaderExist = options?.find(
+                            (option) =>
+                                option.itemType ===
+                                    SelectableOptionMenuItemType.Header &&
+                                option.text === freeformOptionsHeaderText
+                        );
+                        newOptions = freeFromOptionsHeaderExist
+                            ? [...newOptions, newOption]
+                            : [...newOptions, freeformOptionsHeader, newOption];
+                        setOptions(newOptions);
```
msnyder-msft, line 481: the fact that we need to set state multiple times in a row screams that we should be using a reducer for this component.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                        setSelectedOption(newOption);
+                        setSelectedKey(newParsedOptionValue);
+                    }
+                    if (onResourceChange)
```
msnyder-msft, line 485: can we pull out the duplicate code that notifies the caller of the change? It seems like both conditionals do the same thing.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+            }
+            setOptions(restOfOptions);
+
+            if (option.key === selectedOption?.key) {
+                setSelectedOption(null);
+                if (onResourceChange)
+                    onResourceChange(
+                        null,
+                        restOfOptions
+                            .filter(
```
msnyder-msft, line 538: we seem to do this filtering a lot, can we pull it out to a helper of some sort to simplify? or have a wrapper function around the onResourceChange that all the places call so they don't have to filter out the headers in every spot?

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                                (option) =>
+                                    option.itemType !==
+                                    SelectableOptionMenuItemType.Header
+                            )
+                            .map((option) => option.data || option.text)
+                    );
+            } else {
+                if (onResourceChange)
```
msnyder-msft, line 546: can we pull out the notification of the parent outside the conditional since it's duplicated in both cases?

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                                    SelectableOptionMenuItemType.Header
+                            )
+                            .map((option) => option.data || option.text)
+                    );
+            }
+        },
+        [options, selectedOption]
+    );
+
+    const handleOnRender = useCallback(
```
msnyder-msft, line 562: naming nit: `onRenderOption`

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,649 @@
...
+                allowFreeform={allowFreeform}
+                autoComplete={'on'}
+                required
+                disabled={disabled}
+                errorMessage={inputError}
+                text={selectedOption?.text} // to show the selectedOption text even there is input error
+                onChange={(_e, option, _idx, value) => {
+                    handleOnChange(option, value);
+                }}
+                onRenderLabel={() => (
```
msnyder-msft, line 608: could pull out the custom render for the label to a callback as well if you want.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -0,0 +1,374 @@
...
+} from '../../Models/Constants/Constants';
+import {
+    ADTEnvironmentInLocalStorage,
+    ADTSelectedEnvironmentInLocalStorage,
+    StorageAccountsInLocalStorage,
+    StorageAccountToContainersMapping
+} from './EnvironmentPicker.types';
+
+class EnvironmentPickerManager {
+    static displayTextForEnvironment(env: string | IAzureResource) {
```
msnyder-msft, line 17: naming nit: start functions wiht a verb `getEnvironmentDisplayName`

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -0,0 +1,374 @@
...
+    StorageAccountsLocalStorageKey
+} from '../../Models/Constants/Constants';
+import {
+    ADTEnvironmentInLocalStorage,
+    ADTSelectedEnvironmentInLocalStorage,
+    StorageAccountsInLocalStorage,
+    StorageAccountToContainersMapping
+} from './EnvironmentPicker.types';
+
+class EnvironmentPickerManager {
```
msnyder-msft, line 16: can we export functions independently like we do for `Utils.ts`? That way we can import individual ones and not always have to prefix them with `EnvironmentPickerManager` which is quite verbose and makes it harder to read.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -0,0 +1,374 @@
...
+                        env,
+                        AzureResourceTypes.DigitalTwinInstance
+                    )
+                );
+                return urlObj.hostname.split('.')[0];
+            } else {
+                return null;
+            }
+        } catch (error) {
+            console.log(error.message);
```
msnyder-msft, line 31: `console.error` please

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -0,0 +1,374 @@
...
+        if (resource) {
+            if (typeof resource === 'string') {
+                // it means the option is manually entered using freeform
+                if (resourceType) {
+                    switch (resourceType) {
+                        case AzureResourceTypes.DigitalTwinInstance:
+                        case AzureResourceTypes.StorageAccount:
+                            return resource;
+                        case AzureResourceTypes.StorageBlobContainer: {
+                            const storageAccountEndpointUrl = EnvironmentPickerManager.getResourceUrl(
```
msnyder-msft, line 67: nit: can we be consistent with either `this` or `EnvironmentPickerManager`? They do the same thing but just makes it easier to read when we are consistent.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -0,0 +1,374 @@
...
+        resourceType: AzureResourceTypes, // always pass this in case the resource is string type
+        parentResource?: IAzureResource | string
+    ) {
+        return resources.map((resource) =>
+            this.getResourceUrl(resource, resourceType, parentResource)
+        );
+    }
+
+    static isResourceUrlsEqual(
+        resourceUrlStr1: string,
```
msnyder-msft, line 130: nit: `areResourceUrlsEqual`

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -0,0 +1,374 @@
...
+        }
+        return resourceUrlStr1 === resourceUrlStr2;
+    }
+
+    static getContainerNameFromUrl(containerUrl: string) {
+        try {
+            const containerUrlObj = new URL(containerUrl);
+            return containerUrlObj.pathname.split('/')[1];
+        } catch (error) {
+            console.log(error.message);
```
msnyder-msft, line 147: `console.error` please

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -0,0 +1,374 @@
...
+            ? environmentsInLocalStorage
+                  .filter((e) => e.config?.appAdtUrl)
+                  .map((e: ADTEnvironmentInLocalStorage) => e.config.appAdtUrl)
+            : [];
+        return environmentUrls;
+    }
+
+    static updateEnvironmentsInLocalStorage(
+        environments: Array<string | IAzureResource>,
+        localStorageKey: string
```
msnyder-msft, line 240: `localStorageKey: string = EnvironmentsLocalStorageKey`

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -122,454 +115,462 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
-            ) {
-                setContainers(containers.concat(newVal));
-            }
-        },
-        [containers, isValidUrlStr]
-    );
-
     const handleOnSave = useCallback(() => {
         if (props.onDismiss) {
             props.onDismiss();
```
msnyder-msft, line 294: we should probably wait to notify the parent until we are done right? They could potentially unmount the EnvironmentPicker before it's done doing it's work.

msnyder-msft (changes requested):
This is a great improvement. I can't wait to see this get in. left a variety of comments, mostly naming nits and cleanup stuff like dependency lists. However, the biggest blocker for me right now is the state management of the selected options. IMO those need to be moved to the Deeplink Context and should not duplicate the state locally. In addition, the account name needs to get added to the context all-up since that would not get transferred to a new session as it stands today.

msnyder-msft:
Can we also sort the results in the list or resources? First by subscription and then by alphabetical order within that subscription.

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
```diff
@@ -59,8 +59,8 @@ export const DeeplinkContextReducer: (
                 draft.sceneId = action.payload.sceneId || '';
                 break;
             }
-            case DeeplinkContextActionType.SET_STORAGE_URL: {
-                draft.storageUrl = action.payload.url || '';
+            case DeeplinkContextActionType.SET_STORAGE_CONTAINER_URL: {
+                draft.storageContainerUrl = action.payload.url || '';
```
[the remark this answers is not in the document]
iremgokce: no we dont need to keep track of the storage account in the context, as the storage account name is in the container url, I extract the account name from container url (e.g. `EnvironmentPickerManager.getStorageAccountUrlFromContainerUrl`)

src/Models/Context/DeeplinkContext/DeeplinkContext.types.ts
```diff
@@ -78,7 +78,7 @@ export type DeeplinkContextAction =
           payload: { sceneId: string };
       }
     | {
-          type: DeeplinkContextActionType.SET_STORAGE_URL;
+          type: DeeplinkContextActionType.SET_STORAGE_CONTAINER_URL;
           payload: { url: string };
```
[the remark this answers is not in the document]
iremgokce: we dont need to keep track of the storage account as I explained in my previous comment, so I didn't add such action or context variable

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
```diff
@@ -647,11 +640,12 @@ const ADT3DScenePageBase: React.FC<IADT3DScenePageProps> = ({
 
 const ADT3DScenePage: React.FC<IADT3DScenePageProps> = (props) => {
     const { adapter } = props;
+    const adtHostUrl = adapter.getAdtHostUrl();
     return (
         <DeeplinkContextProvider
             initialState={{
-                adtUrl: adapter.getAdtHostUrl(),
-                storageUrl: adapter.getBlobContainerURL()
+                adtUrl: adtHostUrl ? `https://${adtHostUrl}` : adtHostUrl,
```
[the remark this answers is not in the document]
iremgokce: I just pushed a change for there. 

The adtHostUrl variable in the adapter is the host part of the url without protocol, but in the context I noticed that you use the protocol part too, that is why I changed it to append https:// as prefix. This variable is used to make requests using the proxy where the protocol was already appended.
image
 
image

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
```diff
@@ -647,11 +640,12 @@ const ADT3DScenePageBase: React.FC<IADT3DScenePageProps> = ({
...
+                storageContainerUrl: adapter.getBlobContainerURL()
```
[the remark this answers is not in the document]
iremgokce: account name is in the container url, so I extract it from the container url in the environment picker. we dont need to expose it as a separate variable in the context

src/Resources/Locales/en.json
```diff
@@ -762,21 +759,18 @@
     },
     "environmentPicker": {
         "environmentUrl": "Azure Digital Twins instance URL",
-        "containerUrl": "Azure Storage container URL",
-        "enterEnvironmentUrl": "e.g. https://example.api.wus2.digitaltwins.azure.net",
-        "enterContainerUrl": "e.g. https://example.blob.core.windows.net/example",
+        "storageAccountName": "Azure Storage account name",
+        "storageAccountUrl": "Azure Storage account URL",
+        "storageContainerName": "Azure Storage container name",
```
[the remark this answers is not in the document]
iremgokce: Azure Storage is the "service" name - a special name, that is why I used capital letter in those words. 
image

src/Components/ResourcePicker/ResourcePicker.types.ts
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * Call to provide customized styling that will layer on top of the variant rules.
+     */
+    styles?: IStyleFunctionOrObject<
+        IResourcePickerStyleProps,
+        IResourcePickerStyles
+    >;
+    adapter: IAzureManagementAdapter | MockAdapter;
```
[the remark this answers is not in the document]
iremgokce: what do you mean? all of those are necessary props. I can group them if it would be better?

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -2,71 +2,109 @@ import axios from 'axios';
...
+    ) {
+        const result = await axios({
+            method: 'get',
+            url: nextLink || url,
+            headers: {
+                'Content-Type': 'application/json',
+                authorization: 'Bearer ' + token
+            },
+            params: {
+                ...(!nextLink && {
```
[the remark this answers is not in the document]
iremgokce: we include api version only when there is 'not' a next link. it is intentional, because nextLink is a full url which includes all the params

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -122,454 +115,462 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
+            ...containerItems,
+            containerToEdit: resource,
+            containers: resources
+        });
+
+        // update mappings in case change is based on addition or removal of a resource
+        const defaultStorageAccountMapping = defaultStorageAccountToContainersMappingRef.current?.find(
+            (mapping) =>
+                EnvironmentPickerManager.isResourceUrlsEqual(
+                    EnvironmentPickerManager.getResourceUrl(
```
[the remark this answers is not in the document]
iremgokce: since EnvironmentPickerManager is an internal manager class I didn't want to export the methods in there to surface (it is not like Utils class where other components or other src files also need those methods). I can export those though. What do you think @msnyder-msft?

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -122,454 +115,462 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
+            environments: resources
+        });
+    };
+
+    const handleOnStorageAccountResourceChange = (
+        resource: IAzureResource | string,
+        resources: Array<IAzureResource | string>
+    ) => {
+        setStorageAccountItems({
+            ...storageAccountItems,
```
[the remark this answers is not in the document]
iremgokce: ohh yeah, produce... :(

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -122,454 +115,462 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
+                        }
+                    }
+                    storageAccountToContainersMapping.push(newPair);
                 }
-            }
-        },
-        []
-    );
+            });
+            defaultStorageAccountToContainersMappingRef.current = storageAccountToContainersMapping;
```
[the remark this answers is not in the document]
iremgokce: I don't want the component to render when this value changes. I am just using it to keep track of a value all over the place in useCallbacks etc.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -122,454 +115,462 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
+        deeplinkDispatch({
+            type: DeeplinkContextActionType.SET_ADT_URL,
+            payload: {
+                url: selectedEnvironmentUrl
             }
+        });
+        setEnvironmentItems({
+            environments: environmentUrls,
+            selectedEnvironment: selectedEnvironmentUrl,
+            environmentToEdit: selectedEnvironmentUrl
```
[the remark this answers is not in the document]
iremgokce: 'selected' values are used to show on the page header, the '...toEdit' values are for state variables where user change as they interact with the dropdowns of which the initial value is set the selected values and not persistent as 'selected' until you 'save' in the modal

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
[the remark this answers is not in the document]
msnyder-msft: I mean I don't think anyone is going to import a function from a utility class of a specific component so I'm not sure there's really a risk there. Are you just trying to hide it from intellisense?

src/Adapters/AzureManagementAdapter.ts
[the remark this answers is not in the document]
msnyder-msft: Gotcha. Ok. makes sense. Maybe just throw a comment in there about that?

src/Components/ResourcePicker/ResourcePicker.types.ts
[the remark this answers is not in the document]
msnyder-msft: Just seems like a crazy complex component to actually consume. :)

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -593,74 +596,175 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
-                            label={t('environmentPicker.containerUrl')}
-                            allowFreeform={true}
-                            autoComplete={'on'}
-                            options={containerOptions}
-                            styles={comboBoxStyles}
-                            required
-                            text={containerUrlToEdit}
-                            onChange={(_e, option, _idx, value) =>
-                                handleOnContainerUrlChange(option, value)
+                {firstTimeVisible && (
```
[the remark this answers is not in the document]
iremgokce: firstTimeVisible is used not to render resource picker components with data fetch requests if the dialog has not opened yet for the first time. We want to trigger the adapter data fetching when we open the dialog for the first time and keep the same component mounted and same data request flow going as we close and open the dialog afterwards.

The issue is when Dialog fluent component is open and closed the content of it [is being unmounted](https://github.com/microsoft/fluentui/issues/11529) (all the ResourcePicker children components) - so it was a problem since we dont want to unmount and mount and refetch the resources again and want to keep the same loading process going while we close and open the dialog by clicking edit icon. That is why I applied the suggested solution for it - just always visible but changing the css. This also answers your other question below.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -593,74 +596,175 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
+                        )}
                     </span>
                 </div>
             )}
 
             <Dialog
-                hidden={isDialogHidden}
+                hidden={false}
+                styles={{
+                    root: { display: !isDialogHidden ? 'flex' : 'none' }
```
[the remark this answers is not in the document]
iremgokce: Please see my other comment above about firstTimeVisible value for more explanation. 

The issue is when Dialog fluent component is open and closed the content of it [is being unmounted](https://github.com/microsoft/fluentui/issues/11529) (all the ResourcePicker children components) - so it was a problem since we dont want to unmount and mount and refetch the resources again and want to keep the same loading process going while we close and open the dialog by clicking edit icon. That is why I applied the suggested solution for it - just always visible but changing the css. This also answers your other question below.

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: oh yeah, it was originally a useMemo variable forgot to rename it

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: it uses resourcesRef inside and one more internal method. will keep it as is

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: `mergeOptionsWithProps` is called after fetching the actual data. it is setting the dropdown options after fetch finishes in this order => fetched resources + additional options using freeform + selected option (if exists and not included in the additional options). The reason why I set the selected key again is because initially the "key" is based on the display value since there is no "id" information of those dropdown options (read from localstorage or passed via props). So I need to update the selected key value with resource ids after fetching is completed and merging options is done, since there is no longer that option with the old key exist.

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: when the fluent combobox's allowfreeform prop is enabled, when onchange two parameters are passed: option and value. Option is referring to an existing dropdown option whereas value is a newly entered freeform value.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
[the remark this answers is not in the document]
msnyder-msft: I understand the reasoning, that makes sense. But we still should not do this. That doesn't fix the accessibility issue. Let's think about alternative solutions to that problem rather than doing this.

src/Components/ResourcePicker/ResourcePicker.types.ts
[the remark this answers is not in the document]
iremgokce: you can just use two props - resource type and required permissions as in stories, other props are mostly for handling our use case, nothing much there to do

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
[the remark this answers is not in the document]
iremgokce: deeplink context only needs to know the selected values, ...ToEdit is the internal state variable for the component. 
image

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: I prefer naming the handler functions to start with handle... where as naming the props start with on.... There was a stack post about it which I adopted

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -122,454 +115,462 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
+                    AzureResourceTypes.DigitalTwinInstance
+                ),
+                EnvironmentPickerManager.getResourceUrls(
+                    environmentItems.environments,
+                    AzureResourceTypes.DigitalTwinInstance
+                )
+            );
         }
+
+        setContainerItems({
```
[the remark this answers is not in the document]
iremgokce: nope, when we change a dropdown option or delete a freeform option, after closing the modal and open, we want to revert the values and ...ToEdit set back to the selected values.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -0,0 +1,374 @@
...
+            AzureResourceTypes.StorageBlobContainer,
+            parentStorageAccount
+        );
+        localStorage.setItem(
+            localStorageKey ?? ContainersLocalStorageKey,
+            JSON.stringify(containerUrls)
+        );
+    }
+
+    static getSelectedContainerUrlFromLocalStorage(localStorageKey: string) {
```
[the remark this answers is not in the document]
iremgokce: hey @msnyder-msft should I add the read from local storage part for selected values in here? 
image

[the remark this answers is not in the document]
msnyder-msft: Hmmm. Somehow those existing values are persisted today. Were those being read from the environment picker before and passed here as the initial state? Somehow when you reload the app today, we are persisting that information, I thought it was in the deeplink context but maybe it wasn't? I'm fine with it being there though. That would make sense to me. This seems like the right spot, we just have to make sure the logic checkouts out with the right prioritizations. After QSP seems ok to me. Let's do the Local storage read above that block though. It's a pretty complicated logic already so let's try to keep it readable.

src/Adapters/MockAdapter.ts
```diff
@@ -719,71 +717,106 @@ export default class MockAdapter
...
+                        Object.keys(AzureResourceDisplayFields).some(
+                            (displayField) =>
+                                !!resource[displayField]?.includes(displayField)
+                        )
+                    );
+                }
+                resources = resources.slice(
+                    0,
+                    searchParams?.take || MAX_RESOURCE_TAKE_LIMIT
+                ); // take the first n number of resources to make sure the browser won't crash with making thousands of requests
```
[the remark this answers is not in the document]
iremgokce: it was for preventing the browser to crash with thousands of calls, by taking a max number of resolved resources right before making another batch of calls for permission check for each resource. But since we use search params for containers based on storage account id, we may no longer need that. (for example, I have ~700 storage account and the whole getResources method takes around 60 seconds with around 1200 requests). I am going to remove take now but will keep it in comment for future use, users probably would have much more less than those number of resources.

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: the key value of the freeform options in the dropdown is the text itself (because there is no fetching resources for them). That is why I am comparing based on the text values we see in the dropdown to check if the options exists or not.

I dont understand what you mean by saying "a manipulated display value". In the dropdown there are options fetched from server, plus the manually added strings using freeform in the combobox. the manually added ones' key values are the text itself, and after adding an option there is no way to manipulate the value of it (you can only remove that option and add a new one)

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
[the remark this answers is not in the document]
iremgokce: yep it was my concern, but revised it as you suggested

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -20,7 +21,7 @@ import {
 } from '../Models/Constants';
 import { createGUID, getDebugLogger } from '../Models/Services/Utils';
 
-const MAX_RESOURCE_TAKE_LIMIT = 1000;
+// const MAX_RESOURCE_TAKE_LIMIT = 1000; // if necessary limit the number of resources to check against permissions
```
msnyder-msft, line 24: do we want to remove this if it's not used?

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -505,25 +480,27 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
...
 
-                    // take the first n number of resources to make sure the browser won't crash with making thousands of requests
+                    // if necessary, take the first n number of resources to make sure the browser won't crash with making thousands of requests
                     // to check permissions for each resource, however, this might cause 0 result after checking the permissions for the first taken n resources
                     // it is hard to limit the number of requests being made and make sure to capture the resources that we might have required access permissions
-                    resources = resources.slice(
-                        0,
-                        searchParams?.take || MAX_RESOURCE_TAKE_LIMIT
-                    );
+                    // resources = resources.slice(
```
msnyder-msft, line 499: wasn't this important? I thought you were seeing it crash your machine without this.

.eslintrc.json
```diff
@@ -1,10 +1,7 @@
 {
     "root": true,
     "parser": "@typescript-eslint/parser",
-    "plugins": [
-        "@typescript-eslint",
-        "react-hooks"
-    ],
+    "plugins": ["@typescript-eslint", "react-hooks"],
```
msnyder-msft, line 4: can we revert the changes to this file?

src/Adapters/AzureManagementAdapter.ts
iremgokce: Since now we are using storageContainerAccountId parameter to pass to the container picker it is not crashing and we don't need to slice the resources result. But I wanted to keep it as comments for the future if any of the other resource pickers crashes, users can try to slice the resources before making another batch of calls for role assignments.

src/Adapters/AzureManagementAdapter.ts
iremgokce: Since now we are using storageContainerAccountId parameter to pass to the container picker it is not crashing and we don't need to slice the resources result. But I wanted to keep it as comments for the future if any of the other resource pickers crashes, users can try to slice the resources before making another batch of calls for role assignments.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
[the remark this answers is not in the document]
msnyder-msft: yeah. so we should only need a single value in state now for each dropdown right? Cause the persisted value will come from the deeplink context.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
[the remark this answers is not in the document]
msnyder-msft: @darsney you have any creative ideas here?

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
```diff
@@ -39,6 +44,9 @@ export const DeeplinkContextReducer: (
         switch (action.type) {
             case DeeplinkContextActionType.SET_ADT_URL: {
                 draft.adtUrl = action.payload.url || '';
+                if (action.payload.url) {
```
msnyder-msft, line 47: why do we care whether there is value? If I call this with null wouldn't we want it to clear the value in local storage? It'd be weird to have it persisted.

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
```diff
@@ -59,8 +67,11 @@ export const DeeplinkContextReducer: (
                 draft.sceneId = action.payload.sceneId || '';
                 break;
             }
-            case DeeplinkContextActionType.SET_STORAGE_URL: {
-                draft.storageUrl = action.payload.url || '';
+            case DeeplinkContextActionType.SET_STORAGE_CONTAINER_URL: {
+                draft.storageContainerUrl = action.payload.url || '';
+                if (action.payload.url) {
```
msnyder-msft, line 72: same here, why do we need the check?

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
```diff
@@ -160,7 +179,7 @@ const buildDeeplink = (
             : undefined,
         mode: currentState.mode,
         adtUrl: currentState.adtUrl,
-        storageUrl: currentState.storageUrl
+        storageContainerUrl: currentState.storageContainerUrl
```
msnyder-msft, line 182: why is this change needed? The public contract should not need to change. THat's why it's decoupled from our internal contracts. Ideally this would not break any existing links that exist in the wild.

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
```diff
@@ -197,3 +216,44 @@ const parseArrayParam = (value: string): string[] => {
...
     return value.split(ARRAY_VALUE_SEPARATOR);
 };
+
+// START of local storage handling
+/**
+ * read the selected environment url from local storage if exists to set the initial value of 'adtUrl' in the initial default state of DeeplinkContext
+ */
+const getSelectedEnvironmentUrlFromLocalStorage = () => {
+    try {
+        return (JSON.parse(
```
msnyder-msft, line 226: can we store these 2 data points in a single object in the storage? it seems like they can be persisted together as part of the app state. 
`environment:{adtUrl:string, storageUrl:string}`

src/Components/EnvironmentPicker/EnvironmentPicker.types.ts
```diff
@@ -54,3 +54,35 @@ export type StorageAccountToContainersMapping = {
     storageAccountUrl: string;
     containerNames: Array<string>;
 };
+
+// START of Actions
+export const SET_ENVIRONMENT_ITEMS = 'SET_ENVIRONMENT_ITEMS';
```
msnyder-msft, line 59: there should be a mapping of action to payload type definitions. Please look at the SceneViewContext, DeeplinkContext or SceneDialog.state for examples.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -114,7 +111,7 @@ export const getResourceUrls = (
     resourceType: AzureResourceTypes, // always pass this in case the resource is string type
     parentResource?: IAzureResource | string
 ) => {
-    return resources.map((resource) =>
+    return resources?.map((resource) =>
```
msnyder-msft, line 114: why did we need to add the ?. Can't we just ensure we are passing valid data in? Ideally any state variables would initialize to [] instead of null.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -250,34 +248,6 @@ export const updateEnvironmentsInLocalStorage = (
...
-        JSON.stringify({
-            appAdtUrl: getResourceUrl(
-                selectedEnvironment,
-                AzureResourceTypes.DigitalTwinInstance
-            )
-        })
-    );
-};
-
 export const getStorageAccountOptionsFromLocalStorage = () => {
```
msnyder-msft, line 251: please define types for anything being returned from local storage. Then use that same type when storing the value to give us some safety.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -143,14 +143,15 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
 
     const areValuesEqual = useCallback(
         (value1: string, value2: string) => {
+            if (!value1 || !value2) return false;
```
msnyder-msft, line 146: I see you had to make the same changes to the EnvironmentPickerManager file as well. It seems like it would make sense to move this to a utils file so we don't end up with bugs in the future if we fix something in one of them and not in the other.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -346,13 +352,18 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
...
+                setSelectedOption(existingOption);
+            } else {
+                setSelectedOption({
+                    key: selectedOptionProp,
+                    text: selectedOptionProp,
+                    styles: comboBoxOptionStyles
+                });
+            }
         }
     }, [selectedOptionProp]);
```
msnyder-msft, line 368: do we expect the prop to change? Could we just store teh values on mount and not need the compleixty of maintaining both the value from prop as well as state?

src/Resources/Locales/en.json
```diff
@@ -1124,5 +1118,23 @@
         "days": "days",
         "year": "year",
         "years": "years"
+    },
+    "resourcesPicker": {
+        "select": "Select a resource",
+        "noOption": "No resources found",
+        "loadingResources": "Loading resources",
```
msnyder-msft, line 1125: please name the resource keys for what they are used for and not for the value they represent. `enterEnvironmentUrl` --> `environmentDropdownPlaceholder`. When we change the text it no longer makes sense. Ex: `select` --> `storageContainerPlaceholderText`

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -1,585 +1,518 @@
 import {
-    ComboBox,
     DefaultButton,
```
msnyder-msft, line 2: it is not possible to clear the value in the dropdown menus. IF you delete the content and click away it reverts back to the previous value which is not what I would expect
cant delete
.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -1,585 +1,518 @@
...
     Dialog,
     DialogFooter,
```
msnyder-msft, line 4: I'm seeing a ton of failed requests when I try to run the environment picker.
image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -1,585 +1,518 @@
...
     DialogType,
     FontIcon,
     FontSizes,
-    IComboBoxOption,
-    IComboBoxStyles,
```
msnyder-msft, line 10: It looks really bad when we show entries without a URL property. I think they should be hidden from teh list instead of showing this error message. 
@darsney thoughts?
image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -1,585 +1,518 @@
...
-    Icon,
     IconButton,
     IDialogContentProps,
```
msnyder-msft, line 9: I'm thinking we should put the storage container name as the header for the list of containers instead of subscription.
image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
msnyder-msft: You can't use it anyways so it just looks bad and doesn't help the user.
image

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -466,20 +479,21 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
...
                         setSelectedKey(newParsedOptionValue);
+
+                        if (onChange)
+                            onChange(
+                                newParsedOptionValue,
+                                getResourcesFromOptions(newOptions) || []
+                            );
+                    } else {
+                        setSelectedOption(existingOption);
+                        setSelectedKey(existingOption.key);
```
msnyder-msft, line 490: why do we need to store both the option and the key? That seems like a bad smell.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,686 @@
...
+    selectedOption: selectedOptionProp,
+    allowFreeform = false,
+    disabled = false
+}) => {
+    const { t } = useTranslation();
+    const classNames = getClassNames(styles, {
+        theme: useTheme()
+    });
+
+    const [options, setOptions] = useState([]);
```
msnyder-msft, line 80: give this a type please. It's considered an any everywhere it gets used since it doesn't know the type of hte array.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -0,0 +1,686 @@
...
+                  text: selectedOptionProp,
+                  styles: comboBoxOptionStyles
+              }
+            : null
+    );
+    const [additionalOptions, setAdditionalOptions] = useState<
+        Array<IComboBoxOption>
+    >([]);
+    const [selectedKey, setSelectedKey] = useState<string>(null); // resource id or the option text if manually entered option
+    const [error, setError] = useState(null);
```
msnyder-msft, line 94: please give this state a type.

msnyder-msft (changes requested):
Left some more comments. I know you're ready to be done with this PR but I don't want to let these in and then have to chase them down later. I think most of them need to be addressed in this PR.

darsney (approved):
Just pulled this down and played with it, felt pretty good overall, but I agree with Mitch's comments that we should remove resources of this nature since they result in an unusable experience anyway.  Aside from that, the additional comments around typings and unit tests make sense, but I am okay with doing those as a separate PR immediately following this one so that the PR changes remain digestable

image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
darsney: yeah we should remove these

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
darsney: agreed, the local storage state should be typed and validated when read and written

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
darsney: i think thats sort of a function of how the ARM APIs work (I use the word "work" very loosely, because the APIs are a major PITA)

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: do you mean storage account name? because these are already the container names. If so, I still dont want to change it for the sake of consistency with other resource pickers in the UI which is showing resources divided by subscription name. Because if we use subscription name header for this resource picker but subscription name for other resource pickers would be confusing.

msnyder-msft: I actually disagree that it's confusing to change it. The pickers are a hierarchy. The first is a list of storage accounts within different subscriptions. The second is a list of contains within the selected storage account. There will NEVER be a container from another subscription so it makes no sense to call out the subscription that container is in. It would make more sense to either have _no_ header OR to have the storage account that was selected in order to convey that we are filtering the containers based on the previous selection.

msnyder-msft (approved):
Please keep a list of all the open comments that we are not closing so that we can be sure to address all of them in the next PR since we are closing this one without addressing most of them.

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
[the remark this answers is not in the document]
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/commit/3b73e823e67ef7c1e984af5c74b1280f1813ceef#diff-58d99eaf30a6daf7142117bf32c1811dbf2790559cfddad0ad1d54b73c1f5c78

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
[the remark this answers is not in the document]
iremgokce: Fixed this by removing "selected" state variables from EnvironmentPicker component: https://github.com/microsoft/iot-cardboard-js/commit/3b73e823e67ef7c1e984af5c74b1280f1813ceef

.eslintrc.json
iremgokce: it is just linting, it automatically made it one line when I was enabling/disabling react-hooks/exhaustive-deps and save the file. Will keep it as is

src/Components/EnvironmentPicker/EnvironmentPicker.state.ts
```diff
@@ -0,0 +1,222 @@
...
+
+export const EnvironmentPickerReducer: (
+    draft: EnvironmentPickerState,
+    action: IAction
+) => EnvironmentPickerState = produce(
+    (draft: EnvironmentPickerState, action: IAction) => {
+        const payload = action.payload;
+
+        switch (action.type) {
+            case SET_ENVIRONMENT_ITEMS:
```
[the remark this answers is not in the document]
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/commit/5480c2f4e4d68e9a1f5f7eb5d95a8c54b1c4bb94#diff-339ca7afc846835ed41cf981fb15ec5df52a13b302731c5e1e8c6f67fed087eeR97

src/Components/EnvironmentPicker/EnvironmentPicker.state.ts
```diff
@@ -0,0 +1,222 @@
...
+        containers: [],
+        selectedContainer: null,
+        containerToEdit: null
+    },
+    firstTimeVisible: false
+};
+
+export const EnvironmentPickerReducer: (
+    draft: EnvironmentPickerState,
+    action: IAction
```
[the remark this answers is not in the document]
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/commit/5480c2f4e4d68e9a1f5f7eb5d95a8c54b1c4bb94#diff-339ca7afc846835ed41cf981fb15ec5df52a13b302731c5e1e8c6f67fed087eeR92

src/Components/EnvironmentPicker/EnvironmentPicker.state.ts
```diff
@@ -0,0 +1,222 @@
...
+    getContainerName,
+    getResourceUrl,
+    getStorageAccountId
+} from './EnvironmentPickerManager';
+
+export const defaultEnvironmentPickerState: EnvironmentPickerState = {
+    environmentItems: {
+        environments: [],
+        selectedEnvironment: null,
+        environmentToEdit: null
```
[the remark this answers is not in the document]
iremgokce: Removed 'selected' state variables from the component, using only '...toEdit' (draft as you referred) of which the initial value comes from the deeplink context: https://github.com/microsoft/iot-cardboard-js/commit/3b73e823e67ef7c1e984af5c74b1280f1813ceef#diff-a570642cc58536abeb789de314c20888652673db23d85927b3a09146c7c7ec42L28

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: because it is a controlled combobox where you can add additional values with freeform or with changes in the prop for selected option I need to be able to set and control the selected key: https://developer.microsoft.com/en-us/fluentui#/controls/web/combobox (Controlled ComboBox)

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/commit/b73eed319a2191ad298d9a707559fee259fa0fac#diff-3bd141423ed5699ed9c540576ed4472a3ed44ba1f9f6c4e73b720705561b84deR81

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/commit/b73eed319a2191ad298d9a707559fee259fa0fac#diff-3bd141423ed5699ed9c540576ed4472a3ed44ba1f9f6c4e73b720705561b84deR100

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/commit/6a3da6ace5c92b93b9f5386ad828f0e9f3326ae4#diff-3bd141423ed5699ed9c540576ed4472a3ed44ba1f9f6c4e73b720705561b84deR499

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/commit/6a3da6ace5c92b93b9f5386ad828f0e9f3326ae4#diff-3bd141423ed5699ed9c540576ed4472a3ed44ba1f9f6c4e73b720705561b84deR499

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: was already fixed: https://github.com/microsoft/iot-cardboard-js/blob/8cf9dcb84ce2ec3f7a62c82c265d7d37c8630aa9/src/Components/EnvironmentPicker/EnvironmentPicker.tsx#L413

iremgokce:
> Can we also sort the results in the list or resources? First by subscription and then by alphabetical order within that subscription.

Added sort: https://github.com/microsoft/iot-cardboard-js/commit/6317d1bcfff78eee2085bdd056b6d232627c5f17

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -2,13 +2,10 @@ import { AzureResourceTypes, IAzureResource } from '../../Models/Constants';
 import {
```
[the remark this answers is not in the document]
iremgokce: Added EnvironmentPickerManager test file: https://github.com/microsoft/iot-cardboard-js/commit/38ce027150df6f1c53a0c5bc7d8e60457c82ab80#diff-59ddafa15a5d3e4c1a0d802605142b1dfebfbd05b989290551b1bda1f9ed61fc

src/Components/EnvironmentPicker/EnvironmentPicker.state.ts
```diff
@@ -0,0 +1,222 @@
+import produce from 'immer';
+import { AzureResourceTypes } from '../..';
+import { IAction, IAzureResource } from '../../Models/Constants/Interfaces';
+import {
+    EnvironmentPickerState,
```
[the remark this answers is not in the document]
iremgokce: Added: https://github.com/microsoft/iot-cardboard-js/commit/0892e957c7ccd22f1286e50bf5f6e6245b384c8d#diff-5c3f47be923e5155cc0d1fb5ae2e18ff8802476aa90aa9d62c076692c2c14509

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
msnyder-msft: can't validate this one since I can't get it to fetch any containers

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
msnyder-msft: Not fixed

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -75,41 +82,34 @@ const modalProps: IModalProps = {
     styles: dialogStyles
 };
 
-const EnvironmentPicker = (props: EnvironmentPickerProps) => {
+const EnvironmentPicker = ({
+    adapter,
+    locale,
```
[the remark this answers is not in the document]
iremgokce: Fixed, [commit1 ](https://github.com/microsoft/iot-cardboard-js/pull/684/commits/3fb398400fb2380a1509e32c79e34ced33369da8)& [commit2](https://github.com/microsoft/iot-cardboard-js/pull/684/commits/ee82aa4b3ff670f351ff159576e6932d5cf9fa48)

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -75,41 +82,34 @@ const modalProps: IModalProps = {
...
-    >([]); // list of storage account and container pairs
-    const [containerItems, setContainerItems] = useState<{
-        containers: Array<IAzureResource | string>; // list of name of container resources or manually entered container names
-        selectedContainer: IAzureResource | string; // either resource itself or manually entered container name
-        containerToEdit: IAzureResource | string; // either resource itself or manually entered container name
-    }>({ containers: [], selectedContainer: null, containerToEdit: null });
+
+    const [state, dispatch] = useReducer(
+        EnvironmentPickerReducer,
+        defaultEnvironmentPickerState
```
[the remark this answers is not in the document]
iremgokce: what do you mean? useReducer hook requires a default state along with the reducer?

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -75,41 +82,34 @@ const modalProps: IModalProps = {
...
-    const defaultStorageAccountToContainersMappingRef = useRef<
-        Array<StorageAccountToContainersMapping>
-    >([]); // list of storage account and container pairs
-    const [containerItems, setContainerItems] = useState<{
-        containers: Array<IAzureResource | string>; // list of name of container resources or manually entered container names
-        selectedContainer: IAzureResource | string; // either resource itself or manually entered container name
-        containerToEdit: IAzureResource | string; // either resource itself or manually entered container name
-    }>({ containers: [], selectedContainer: null, containerToEdit: null });
+
+    const [state, dispatch] = useReducer(
```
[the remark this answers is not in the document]
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/commit/545b661f3bb015b92a33649e30af8e1d8d6db99a

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/5a2a6ad1e8181718ff0f6fa4143c38f6ad4e383d

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
iremgokce: Fixed: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/5a2a6ad1e8181718ff0f6fa4143c38f6ad4e383d

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
iremgokce: "storageUrl" is vague since there is storage account url and storage blob container url - which are different concepts. Even if this is the case, do you think I should revert this since it is going to break the existing deeplinks? @msnyder-msft

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
iremgokce: when we say "environment" it usually refers to the ADT environment (which is an ADT "instance" in technical terms) - at least this was what we were referring to. If we are going to do this change it requires other changes for container url in the local storage. Also, the reason why we kept the adt url in a certain way in the localstorage was to be consistent with the ADT explorer app (they use {appAdtUrl: <selected_adt_instance_url>} in localstorage). 

What do you think @darsney about changing the structure of localstorage for selected url information?

src/Components/EnvironmentPicker/EnvironmentPicker.types.ts
iremgokce: yep changed this as a part of fix for your other comment: https://github.com/microsoft/iot-cardboard-js/commit/0892e957c7ccd22f1286e50bf5f6e6245b384c8d#diff-339ca7afc846835ed41cf981fb15ec5df52a13b302731c5e1e8c6f67fed087eeR89

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
iremgokce: fixed: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/d977cb41f7ee0b9a7ee66c8307366112358fb27c#diff-9a856bccecc07dd21c6fb673efb9e1b33bce3bcbd41c9099d49cad741f8e958eR257

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: fixed: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/bd1b8e1e247bd78cff57d56956330d4e289fe77d#diff-d815c7801f566cf88c3c2129276d79d0ce9cf6c7b3f30519d2ffcff99b9e52b9R673

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: yes we expect, it happens when we change the values in dropdown by selecting another option but not save and click to close the modal. In this case we need to revert the selected options back to the original/initial selected values

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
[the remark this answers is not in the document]
msnyder-msft: ahhh right. forgot about that. I'm used to using the reducer inside a context so you don't have to pass it to the context. Ignore this.

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
msnyder-msft: I think it's fine that the public url has the shorter version since they don't need to discriminate between those 2 concepts. I'd vote to keep it as `storageUrl` if for no other reason than it's a shorter url.

src/Resources/Locales/en.json
iremgokce: changed: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/33de50355aab694cfa0eb40d584c3eaa40f80683#diff-f9ebb0fc4e7c56a13821e99a435e6d222ed423f8aced31d6946e848882c9dc72

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: it is the default behavior, I think it should be as is, what do you think @darsney @azimamansuri ?

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
iremgokce: removed and made sure the array state variables are always set to empty array instead of null but also added default values to method params as guard: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/357ab687abe3ed80be740fa4597775a1e50b1c2c#diff-9a856bccecc07dd21c6fb673efb9e1b33bce3bcbd41c9099d49cad741f8e958eR131

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -122,68 +122,59 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
         ) {
             toggleIsDialogHidden();
         }
-    }, [props.isDialogHidden]);
+    }, [isDialogHiddenProp, previousIsDialogHidden, toggleIsDialogHidden]);
 
     useEffect(() => {
         if (!isDialogHidden) {
-            setFirstTimeVisible(true);
+            dispatch({
```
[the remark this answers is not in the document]
iremgokce: changed: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/8238c0f86f9c68def3fda0830a06923825670d0b

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
msnyder-msft: Gross. Combobox sucks.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -122,68 +122,59 @@ const EnvironmentPicker = (props: EnvironmentPickerProps) => {
...
+    const previousIsDialogHidden = usePrevious(isDialogHidden);
     // Figure out if dialog needs to be open from props
     useEffect(() => {
         // Have undefined checked onMount to avoid an extra render
         // Have is previous check from true to false, to just change dialogHidden on open
         if (
             previousIsDialogHidden !== undefined &&
             previousIsDialogHidden === true &&
-            previousIsDialogHidden !== props.isDialogHidden
+            previousIsDialogHidden !== isDialogHiddenProp
```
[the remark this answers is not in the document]
iremgokce: hmm, this piece of code was from earlier time to prevent rerendering by Pato, now I removed the dialoghidden prop from EnvironmenPicker since I think it should be only handled internally by a state variable now (keep in mind that EnvironmentPicker component has two pieces: 1- is the header environment texts with edit icon to open the dialog, 2- the dialog with ResourcePicker components). Does it make sense? Changes: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/7fd8facb79e2d0e8bdb1507b7ff64b355140bc07

FYI @pasanchMSFT I removed your code block to trigger rerender based on a dialoghidden prop change since now it is only handled internally by the EnvironmentPicker component: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/7fd8facb79e2d0e8bdb1507b7ff64b355140bc07#diff-2e58d2278063e505656fe7c499cb7bb0769943ccd89ac6f49417ac4be03cbea0L125

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -0,0 +1,374 @@
...
+
+    static getSelectedEnvironmentUrlFromLocalStorage(localStorageKey: string) {
+        try {
+            return (JSON.parse(
+                localStorage.getItem(
+                    localStorageKey ?? SelectedEnvironmentLocalStorageKey
+                )
+            ) as ADTSelectedEnvironmentInLocalStorage)?.appAdtUrl;
+        } catch (error) {
+            console.log(error.message);
```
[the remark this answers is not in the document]
iremgokce: changed: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/6b12b6c982089e358e5afcdf87d87c58507639cb#diff-a4ffb93ea69fb34204d6668c5b380f3df74ea592d7d784080076e87f2f16811aR224

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
iremgokce: reverted changes: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/5f8c8f5c9997d7b5e99981cd825512530777a782

PR #595, opened 2022-06-27
feat: switched to using markers for alert badges

samjosephgates:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Components/ADT3DViewer/ADT3DViewer.tsx
```diff
@@ -471,6 +476,25 @@ const ADT3DViewerBase: React.FC<IADT3DViewerProps> = ({
         // // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
+    const createBadge = (badgeGroup: SceneViewBadgeGroup) => {
```
msnyder-msft, line 479: can we pull this out to a function and just pass in the callback for hover? we are close to being self contained.

src/Components/ADT3DViewer/ADT3DViewer.tsx
```diff
@@ -481,12 +505,23 @@ const ADT3DViewerBase: React.FC<IADT3DViewerProps> = ({
...
+        if (sceneAlerts) {
+            const markers: Marker[] = [];
+            sceneAlerts.forEach((alert) => {
+                const badge = createBadge(alert);
+                markers.push(badge);
+            });
+
+            setMarkers(markers);
+        }
+    }, [sceneAlerts, sceneThemeState.sceneBackground]);
```
msnyder-msft, line 522: let's not fake out the dependency list, if something insde depends on the sceneBakcground, please pass it through to the function instead of adding it to the dependency list here without actually using it in this effect directly. It just runs too much of a risk of being cleaned up at some point since we can't see that it's used deeper in the pipeline.

src/Components/AlertBadge/AlertBadge.tsx
```diff
@@ -0,0 +1,60 @@
...
+                    >
+                        <Icon iconName={badge.icon} />
+                    </div>
+                ))
+            )}
+        </div>
+    );
+};
+
+export default React.memo(AlertBadge);
```
msnyder-msft, line 60: can we use the `styled` pattern like we have been doing with all new components? That just makes things more flexible in the long run.

src/Components/ADT3DViewer/ADT3DViewer.tsx
```diff
@@ -650,17 +685,16 @@ const ADT3DViewerBase: React.FC<IADT3DViewerProps> = ({
...
                         modelUrl: modelUrl,
-                        onBadgeGroupHover: onBadgeGroupHover,
                         onMeshClick: meshClick,
                         onMeshHover: meshHover,
                         outlinedMeshitems: outlinedMeshItems,
                         showHoverOnSelected: showHoverOnSelected,
                         showMeshesOnHover: showMeshesOnHover,
                         unzoomedMeshOpacity: unzoomedMeshOpacity,
                         zoomToMeshIds: zoomToMeshIds,
+                        markers: markers,
```
msnyder-msft, line 697: please keep the props alphabetized.

src/Components/3DV/SceneView.tsx
```diff
@@ -1175,75 +1094,84 @@ function SceneView(props: ISceneViewProps, ref) {
...
-                                    ?.length
-                            ) {
-                                element.left = position?.left - 20;
-                                element.top = position?.top - 20;
-                            }
-                            element.marker.GroupedUIElement = groupedUIElement;
-                        } else {
-                            removeGroupedItems(markersAndPositions, marker);
-                            // create new group
+                    if (marker.allowGrouping) {
```
msnyder-msft, line 1097: just out of curiosity, what does this gain us? Why make this change?

src/Components/3DV/SceneView.tsx
```diff
@@ -1254,6 +1182,12 @@ function SceneView(props: ISceneViewProps, ref) {
         }
     }, [markers]);
 
+    useEffect(() => {
+        if (markers && !isLoading) {
+            createMarkersWithPosition();
```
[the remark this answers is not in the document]
samjosephgates: The isLoading flag is not indicating data has been loaded but rather the model itself, it is required as we need the model to be loaded before we can position the markers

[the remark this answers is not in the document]
samjosephgates: Added comment to code

src/Components/ADT3DViewer/ADT3DViewer.tsx
samjosephgates: I have added a comment, we have this as a dependency for the useEffect as we need it to rerun when the background changes to update the badge colors.

src/Components/ADT3DViewer/ADT3DViewer.tsx
samjosephgates: Done

src/Components/AlertBadge/Alertbadges.styles.ts
```diff
@@ -0,0 +1,90 @@
...
+                    padding: '2px',
+                    borderRadius: '10px',
+                    textAlign: 'center',
+                    lineHeight: '32px'
+                } as IStyle
+            ],
+            singleContainer: [
+                classNames.singleContainer,
+                {
+                    background: `${backgroundColor.badgeColor}99`,
```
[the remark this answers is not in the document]
samjosephgates: The 99 here is opacity as sceneview has a helper that passes 8 digit hex codes as rgba. Added comment.

src/Components/AlertBadge/AlertBadge.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const wrapperStyle = { width: 'auto', height: '100px' };
+
+export default {
+    title: 'Components/AlertBadge',
+    component: AlertBadge,
+    decorators: [getDefaultStoryDecorator(wrapperStyle)]
+};
+
+export const SingleAlert = () => {
```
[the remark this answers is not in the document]
samjosephgates: I have added stories for 1, 2, 3, 4 and 5 alerts which cover all scenarios.

ptallettms (approved):

src/Components/3DV/SceneView.tsx
```diff
@@ -1074,8 +1075,8 @@ function SceneView(props: ISceneViewProps, ref) {
             top: number;
             left: number;
         }[] = [];
-        if (markers) {
-            markers.forEach((marker) => {
+        if (markersRef.current) {
```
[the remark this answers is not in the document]
samjosephgates: This was required as part of the resize fix. When the resize calls this code the markers prop is null even if it has already been set, so I store it in a ref when the markers prop is updated and use that when we create the markers

src/Components/ADT3DViewer/ADT3DViewer.tsx
```diff
@@ -476,7 +480,15 @@ const ADT3DViewerBase: React.FC<IADT3DViewerProps> = ({
         // // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
-    const createBadge = (badgeGroup: SceneViewBadgeGroup) => {
+    const createBadge = (
```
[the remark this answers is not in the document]
samjosephgates: Moved to module scope

PR #598, opened 2022-06-27
Zarmada/working/ADTS-178

miltonzarmada:


src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -154,9 +158,10 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
         showRelationships,
         showInheritances,
         showComponents,
+        currentNodeIdRef,
         state
     } = useContext(ElementsContext);
-    const { model } = state;
+    const { model, selectedModelId } = state;
```
jamescarpinter, line 164: Do model and selectedModelId always get updated in tandem? I.e. model[@id] == selectedModelId? If so, we should get rid of the latter.

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -464,28 +469,28 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
...
-            polygons.element.data.comment,
-            polygons.element.data.writable,
-            polygons.element.data.content ? polygons.element.data.content : [],
-            polygons.element.data.target,
-            polygons.element.data.maxMultiplicity
+        execute(
+            () => selectRelationship(),
+            () => {
+                setCurrentNode(currentNodeIdRef.current);
+                // const modelCopy = deepCopy(model);
```
jamescarpinter, line 485: Remove commented code

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -464,28 +469,28 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
...
+                const selectedModelIdCopy = deepCopy(selectedModelId);
```
jamescarpinter, line 486: Isn’t model ID a string? Do we need to deep copy?

src/Components/OATPropertyEditor/FormAddEnumItem.tsx
```diff
@@ -162,35 +164,51 @@ export const FormAddEnumItem = ({
...
-        const modelCopy = deepCopy(model);
-        modelCopy[propertiesKeyName][
-            currentPropertyIndex
-        ].schema.enumValues.push(prop);
-        dispatch({
-            type: SET_OAT_PROPERTY_EDITOR_MODEL,
-            payload: modelCopy
-        });
+        execute(
+            () => update(),
```
jamescarpinter, line 202: This can just be `update,`

src/Components/OATPropertyEditor/PropertyListItemSubMenu.tsx
```diff
@@ -47,6 +47,16 @@ export const PropertyListItemSubMenu = ({
     const propertyInspectorStyles = getPropertyInspectorStyles();
     const subMenuItemStyles = getIconMoreSubMenuItemStyles();
 
+    const handleDelete = () => {
```
jamescarpinter, line 50: I think we’re moving to `onXXX` named handlers.

jamescarpinter (changes requested):
This is off to a good start. I’m not sure if we’re correctly serializing the state to properly undo long stacks of actions though. Can jump on a call to discuss that aspect if you’d like.

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -464,28 +469,28 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
...
-            polygons.element.data.description,
-            polygons.element.data.comment,
-            polygons.element.data.writable,
-            polygons.element.data.content ? polygons.element.data.content : [],
-            polygons.element.data.target,
-            polygons.element.data.maxMultiplicity
+        execute(
+            () => selectRelationship(),
+            () => {
+                setCurrentNode(currentNodeIdRef.current);
```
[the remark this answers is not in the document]
miltonzarmada: From my testing, if we have several nodes and different relationship.
We can select a node, then a relationship or the panel background and the action still undo and redo in proper order.

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -464,28 +469,26 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
...
+                dispatch({
+                    type: SET_OAT_SELECTED_MODEL_ID,
+                    payload: selectedModelId
+                });
+                setNameEditor(false);
```
jamescarpinter, line 489: Let's wrap this inline function into a similarly named function as select relationship. I.e. unselectRelationship. We should do this for all the execute calls.

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -464,28 +469,26 @@ const OATGraphCustomEdge: React.FC<IOATGraphCustomEdgeProps> = ({
...
-            nameText,
-            polygons.element.data.displayName,
-            polygons.element.data.description,
-            polygons.element.data.comment,
-            polygons.element.data.writable,
-            polygons.element.data.content ? polygons.element.data.content : [],
-            polygons.element.data.target,
-            polygons.element.data.maxMultiplicity
+        execute(
+            () => selectRelationship(),
```
jamescarpinter, line 482: You can just pass `selectRelationship,` here.

src/Components/OATGraphViewer/Internal/OATGraphCustomEdge.tsx
```diff
@@ -19,12 +19,15 @@ import {
...
-    SET_OAT_CONFIRM_DELETE_OPEN
+    SET_OAT_CONFIRM_DELETE_OPEN,
+    SET_OAT_SELECTED_MODEL_ID
 } from '../../../Models/Constants/ActionTypes';
 import { DTDLRelationship } from '../../../Models/Classes/DTDL';
 import { getPropertyDisplayName } from '../../OATPropertyEditor/Utils';
 import { IOATGraphCustomEdgeProps } from '../../../Models/Constants';
 import OATTextFieldName from '../../../Pages/OATEditorPage/Internal/Components/OATTextFieldName';
+import { CommandHistoryContext } from '../../../Pages/OATEditorPage/Internal/Context/CommandHistoryContext';
+import { deepCopy } from '../../../Models/Services/Utils';
```
jamescarpinter, line 30: Is this required?

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -501,17 +512,41 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
...
+            }
             dispatch({
                 type: SET_OAT_PROPERTY_EDITOR_MODEL,
                 payload: modelClicked
             });
+        } else {
+            dispatch({
+                type: SET_OAT_PROPERTY_EDITOR_MODEL,
+                payload: null
+            });
```
jamescarpinter, line 549: Maybe we pull this out of the if/else block and have it run at the end of the method with the value of `modelClicked`? And have the default value of that be null?

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -952,33 +1049,27 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
...
+                execute(
+                    () => {
+                        currentNodeIdRef.current = node.id;
+                        dispatch({
+                            type: SET_OAT_SELECTED_MODEL_ID,
+                            payload: node.id
+                        });
+                    },
+                    () => {
+                        const selectedModelIdCopy = selectedModelId;
```
jamescarpinter, line 1066: We probably don't need to duplicate this variable right? It'll be the same reference anyway?

src/Pages/OATEditorPage/Internal/Components/OATTextFieldDisplayName.tsx
```diff
@@ -68,11 +70,24 @@ const OATTextFieldDisplayName = ({
                 payload: modelCopy
             });
             setDisplayName(temporaryValue);
+        };
+
+        if (!displayNameLengthError) {
+            execute(
+                () => commit(),
```
jamescarpinter, line 77: Similarly to above, this can just be `commit,`.

jamescarpinter (changes requested):
This is coming together nicely! A few more comments inline.

PR #600, opened 2022-06-27
Cleanup and prop consolidation

pasanchMSFT:
### Summary of changes 🔍 
After some analysis of the codebase, I noticed several small details that would improve it if cleaned up:
1. Some props were being passed down to components with access to context that contained that same information. I removed that duplication in favor of the context.
2. A ref was unused in the ADT3DGlobe, which I removed.
3. ADT3DViewer had a large number of props that are only included when the component is standalone, so I wrapped those in an object to make a clear distinction of which props are needed for that use case.

### Testing 🧪
3DV (stand-alone Viewer and Builder) stories should all work as intended.

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Models/Constants/Interfaces.ts
+++ b/src/Models/Constants/Interfaces.ts
@@ -733,30 +733,34 @@ export interface IADT3DViewerProps extends BaseComponentProps {
     sceneId: string;
     scenesConfig: I3DScenesConfig;
     pollingInterval: number;
-    title?: string;
-    connectionLineColor?: string;
-    enableMeshSelection?: boolean;
+    externalProps?: ADT3DViewerExternalProps;
     addInProps?: IADT3DAddInProps;
     refetchConfig?: () => any;
-    showMeshesOnHover?: boolean;
-    showHoverOnSelected?: boolean;
-    coloredMeshItems?: CustomMeshItem[];
-    /**
-     * Ids of the elements to zoom the camera to focus on
-     */
-    zoomToElementId?: string;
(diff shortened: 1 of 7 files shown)


src/Models/Constants/Interfaces.ts
```diff
@@ -733,32 +733,36 @@ export interface IADT3DViewerProps extends BaseComponentProps {
...
     showModeToggle?: boolean;
     sceneViewProps?: ISceneViewProps;
-    selectedLayerIds?: string[];
     /**
      * Call to provide customized styling that will layer on top of the variant rules.
      */
     styles?: IStyleFunctionOrObject<IADT3DViewerStyleProps, IADT3DViewerStyles>;
 }
 
+export interface ADT3DViewerExternalProps {
```
msnyder-msft, line 748: "ExternalProps" is an odd name to see as a consumer, I wouldn't know what to expect inside of that object or how it is going to be used. Can we think about a more intuitive name? 
For the props like "hideElementsPanel" I liked the idea of `settings` or `viewerSettings` or `settingsOverride`.
For the props like connectionLineColor, let's think about a name that fits that collection.

msnyder-msft (commented):
I'm fine with everything except the "ExternalProps" concept. I'm not sure I agree with that grouping or naming.

src/Models/Constants/Interfaces.ts
pasanchMSFT: Are you suggesting several groups of props?

PR #602, opened 2022-06-28
ADTS-224

JoseZarmada:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Components/OATPropertyEditor/FormUpdateProperty.tsx
```diff
@@ -653,6 +637,26 @@ export const FormUpdateProperty = ({
...
             </div>
 
+            <div className={propertyInspectorStyles.modalRow}>
+                <Text styles={columnLeftTextStyles}>
+                    {t('OATPropertyEditor.id')}
+                </Text>
+                <TextField
+                    placeholder={t('OATPropertyEditor.id')}
+                    onChange={(_ev, value) =>
+                        handleIdChange(
```
jamescarpinter, line 647: I think we're using the `onXXX` pattern for event handlers now. Can you please update this one?

src/Components/OATPropertyEditor/Utils.ts
```diff
@@ -261,4 +262,7 @@ export const handleIdChange = (
     } else {
         setIdErrorLength(true);
     }
+    if (value.length > 0) {
```
jamescarpinter, line 265: Do we need to check if `setIdWarning` is null or not?

This function is used in a few places... are they all updated to pass this new prop?

src/Components/OATPropertyEditor/FormUpdateProperty.tsx
```diff
@@ -308,11 +309,13 @@ export const FormUpdateProperty = ({
...
 
     const getIdErrorMessage = () => {
-        return idLengthError
+        const idWarn = idWarning ? t('OATPropertyEditor.warningId') : '';
+        const idError = idLengthError
             ? t('OATPropertyEditor.errorIdLength')
             : idValidDTMIError
             ? t('OATPropertyEditor.errorIdValidDTMI')
             : '';
+        return `${idWarn} ${idError}`;
```
jamescarpinter, line 318: I'm not sure if concatting these messages together will make it particularly readable. We might find separate lines are better?

jamescarpinter (changes requested):

src/Components/OATPropertyEditor/FormUpdateProperty.tsx
JoseZarmada: Also this is replicated to all the Utils functions we have in this class

src/Components/OATPropertyEditor/Utils.ts
JoseZarmada: I don't think it will ever be a null since its the callback to set the value of the idWarning state const
image

src/Components/OATPropertyEditor/FormUpdateProperty.tsx
JoseZarmada: We are using the Fluent UI textfield to display this https://developer.microsoft.com/en-us/fluentui#/controls/web/textfield was looking at the doc and google a bit to see if I can make this multiline but didn't find a suitable answer, @jamescarpinter can you take a quick look I really can use your experience here

src/Components/OATPropertyEditor/Utils.ts
jamescarpinter: Right - but this function is used by other components too and they don't invoke this with the setIdWarning parameter? See FormAddEnumItem and FormRootModelDetails.

src/Components/OATPropertyEditor/FormUpdateProperty.tsx
jamescarpinter: Good point. Want to open a task for us to review the naming in the Utils file at a future date?

JoseZarmada: I think it would be better to do so

src/Components/OATPropertyEditor/Utils.ts
JoseZarmada: You are Right and those doesn't need it. Done

src/Components/OATPropertyEditor/FormUpdateProperty.tsx
JoseZarmada: Let me Talk to Jamie about this and will circle back to you

JoseZarmada: @jamescarpinter Committed changes from ADTS-240 Into this PR let me know If anything else need to be done

src/Components/OATPropertyEditor/Utils.ts
```diff
@@ -261,4 +262,13 @@ export const handleIdChange = (
     } else {
         setIdErrorLength(true);
     }
+    if (setIdWarning && value.length > 0) {
+        setIdWarning(true);
+    } else {
+        setIdWarning(false);
```
jamescarpinter, line 268: setIdWarning could be null here?

jamescarpinter (changes requested):

src/Components/OATPropertyEditor/Utils.ts
JoseZarmada: Done

jamescarpinter: I mean the function ref could be null, not the value 😄 

We should probably do something like:
```
if (setIdWarning) {
  setIdWarning(value.length > 0);
}
```

JoseZarmada: Ho ok got it, Change is submited

jamescarpinter (approved):

PR #606, opened 2022-06-29
Zarmada/working/ADTS-235

miltonzarmada:


src/Components/OATHeader/OATHeader.tsx
```diff
@@ -68,9 +124,17 @@ const OATHeader = ({ elements, dispatch, state }: OATHeaderProps) => {
...
+        uploadInputRef.current.click();
     };
 
+    useEffect(() => {
+        if (uploadInputRef.current) {
+            // Add webkitdirectory and attributes to input element
+            uploadInputRef.current.setAttribute('webkitdirectory', '');
+            uploadInputRef.current.setAttribute('mozdirectory', '');
+        }
+    }, [uploadInputRef]);
```
jamescarpinter, line 136: Can we not set these attributes in the HTML?

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -54,12 +50,72 @@ const OATHeader = ({ elements, dispatch, state }: OATHeaderProps) => {
...
+                    : null;
+                directoryPath = modelMetadata.directoryPath
+                    ? modelMetadata.directoryPath
+                    : null;
+            }
+
+            // If fileName or directoryPath are null, generate values from id
+            if (!fileName || !directoryPath) {
+                // Get id path - Get section between last ":" and ";"
+                const idPath = id.substring(
```
jamescarpinter, line 77: Let's pull out all the logic for parsing DTMIs and file paths into a utils file.

We should also do some more validation around the indexOf return values.

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -186,6 +308,23 @@ const OATHeader = ({ elements, dispatch, state }: OATHeaderProps) => {
...
+                            const reader = new FileReader();
+                            reader.onload = () => {
+                                const files = [];
+                                for (const file of uploadInputRef.current
+                                    .files) {
+                                    files.push(file);
+                                }
+                                onFilesUpload(files);
+                            };
+                            reader.readAsDataURL(e.target.files[0]);
```
jamescarpinter, line 325: This should be a standalone function rather than inlined.

src/Components/OATPropertyEditor/FormRootModelDetails.tsx
```diff
@@ -220,6 +242,44 @@ export const FormUpdateProperty = ({
...
+        if (modelsMetadata) {
+            const modelMetadata = modelsMetadata.find(
+                (modelMetadata: any) => modelMetadata['@id'] === model['@id']
+            );
+            if (modelMetadata) {
+                setFileName(modelMetadata.fileName);
+                setDirectoryPath(modelMetadata.directoryPath);
+                return true;
+            }
+            return false;
```
jamescarpinter, line 256: Is this required? Won't it just fall through to the return false on the next line?

src/Components/OATPropertyEditor/FormRootModelDetails.tsx
```diff
@@ -220,6 +242,44 @@ export const FormUpdateProperty = ({
...
+            // Get id path - Get section between last ":" and ";"
+            const idPath = id.substring(
+                id.lastIndexOf(':') + 1,
+                id.lastIndexOf(';')
+            );
+            const idVersion = id.substring(id.lastIndexOf(';') + 1, id.length);
+
+            let scheme = id.substring(id.indexOf(':') + 1, id.lastIndexOf(':'));
+            // Scheme - replace ":" with "\"
+            scheme = scheme.replace(':', '\\');
```
jamescarpinter, line 276: Potentially this logic can leverage utils as well (as mentioned above).

jamescarpinter (changes requested):

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -252,6 +232,18 @@ const OATHeader = ({ elements, dispatch, state }: OATHeaderProps) => {
         return metaDataCopy;
     };
 
+    const onFilesChange = (e) => {
+        const reader = new FileReader();
+        reader.onload = () => {
+            const files = [];
+            for (const file of uploadInputRef.current.files) {
+                files.push(file);
+            }
```
jamescarpinter, line 241: This can probably be a `.map`?

jamescarpinter (approved):
Minor suggestion. Else LGTM!

src/Components/OATHeader/OATHeader.tsx
miltonzarmada: I think we can't because uploadInputRef.current.files is not an array. It's an object
Screen Shot 2022-06-30 at 3 19 17 PM

jamescarpinter: Good point. Leave it as-is then. The alternative would be to use spread `[...uploadInputRef.current.files]` or `Array.from` but that seems less tidy than what you've already got.

miltonzarmada:
I added these tasks to the PR. Since it was all header related.

> ADTS-236  Support both file and folder uplaod 
> ADTS-246 Add spinner when importing - ensure always centered and zoomed to fit 
> ADTS-253 Add delete all button with propmt to delete everything

@jamescarpinter 
@jamiewilbraham

PR #607, opened 2022-06-30
Zarmada/working/ADTS-241

miltonzarmada:
### Summary of changes 🔍 

> ADTS-241 ADTS-242 Property selector popUp fixes. Stay active when reaching diagonally && prevent movement
> ADTS-243 JSON Editor should update and sync with graph viewer

### DIFF
--- a/src/Components/OATPropertyEditor/OATPropertyEditor.styles.ts
+++ b/src/Components/OATPropertyEditor/OATPropertyEditor.styles.ts
@@ -186,9 +186,20 @@ export const getPropertyInspectorStyles = () => {
                 position: 'relative',
                 width: 'fit-content',
                 height: 'fit-content',
-                '& button': {
+                '& div': {
                     height: 'fit-content',
-                    paddingLeft: '10px'
+                    marginLeft: '10px',
+                    display: 'flex',
+                    alignItems: 'center',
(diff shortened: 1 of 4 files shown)


src/Components/OATPropertyEditor/OATPropertyEditor.styles.ts
```diff
@@ -186,9 +186,20 @@ export const getPropertyInspectorStyles = () => {
                 position: 'relative',
                 width: 'fit-content',
                 height: 'fit-content',
-                '& button': {
+                '& div': {
```
jamescarpinter, line 189: What is the ampersand doing here?

src/Components/OATPropertyEditor/PropertyList.tsx
```diff
@@ -277,15 +277,15 @@ export const PropertyList = ({
                         }}
                         ref={addPropertyLabelRef}
                     >
-                        <ActionButton>
+                        <div>
```
jamescarpinter, line 280: I think we wanted to make this a button so it was correct from an accessibility perspective? Do we need to change it to a div?

jamescarpinter (changes requested):

jamescarpinter (approved):

PR #611, opened 2022-06-30
SceneViewContext phase 1

pasanchMSFT:
### Summary of changes 🔍 
**Overview**
Created a POC to showcase how `SceneView` relevant values can be handled with a context. In this proof of concept, `outlinedMeshItems` are managed with a context and ultimately passed down to `SceneView` through props. The idea behind this pattern is to have a context that handles all `SceneView` relevant values in both the viewer and the builder component trees, but ultimately not change how `SceneView` works, just how data is managed before it reaches the component. 

Currently on the builder side some `SceneView` relevant values are part of the `SceneBuilderContext` and on the viewer side no context holds these values. What `SceneViewContext` tries to achieve is mirror the way these two component trees handle `SceneView` data and simplify development around it.

In this PR each tree (builder/viewer) create the context at their top-most parent component `ADT3DViewer` and `ADT3DSceneBuilder`. This is done so the viewer and builder components can still be stand-alone. Meaning 3DV stories should not be affected by this change.

**Future**
In the future more props/values can be added to `SceneViewContext` to improve state management, remove prop duplication and simplify maintenance of data relevant to `SceneView` component.

### Testing 🧪
1. Review stories that change the colors of items by hovering over elements. (Builder: Elements list, Elements list inside Behavior form, Element form. Viewer: Hover over left panel elements).
2. Review 3DV stories (stand-alone builder and viewer components).

### Checklist ✔️
- [x] Added relevant labels


BrennaCourtney (approved):
With the exception of the 3DV for Builder still being broken (an existing bug), UX side looks good; also I remember going over the code in dev time so the adaptable "setSceneViewAttributes" is v cool

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.tsx
```diff
@@ -436,10 +432,18 @@ const ADT3DSceneBuilderBase: React.FC<IADT3DSceneBuilderCardProps> = (
             }
 
             if (outlinedElements.length > 0) {
-                setSceneViewAttributes({ outlinedMeshItems: outlinedElements });
+                // See if this makes sense or if it is better to have a setter
+                sceneViewDispatch({
```
msnyder-msft, line 462: is it really a reset in this case? I do have some resets that take the value to revert too, so it's really more of a semantic decision than a "properness" question.

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.tsx
```diff
@@ -481,39 +485,27 @@ const ADT3DSceneBuilderBase: React.FC<IADT3DSceneBuilderCardProps> = (
...
-                        }
-
-                        // colored meshes that are in the elements contained in the hovered behavior
-                        setSceneViewAttributes({
-                            outlinedMeshItems: createCustomMeshItems(
-                                ids,
-                                state.objectColor.outlinedMeshHoverColor
-                            )
+                        // Grab the selected elements from behavior and outline the meshes from them
+                        sceneViewDispatch({
```
msnyder-msft, line 515: love how much simpler this is. Very cool.

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.tsx
msnyder-msft: I would expect it to use the "previously outlined items" ref if it is really a 'reset'

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.tsx
```diff
@@ -952,7 +965,9 @@ const ADT3DSceneBuilder: React.FC<IADT3DSceneBuilderCardProps> = (props) => {
     return (
         <DeeplinkContextProvider>
             <SceneThemeContextProvider>
-                <ADT3DSceneBuilderBase {...props} />
+                <SceneViewContextProvider outlinedMeshItems={[]}>
```
msnyder-msft, line 998: does it  need to take in the outlinedMeshItems initial state if it's always empty array?

src/Models/Context/SceneViewContext/SceneViewContext.tsx
```diff
@@ -29,16 +32,92 @@ export const SceneViewContextReducer: (
...
+                    .forEach((ds) => {
+                        ds.elementIDs.forEach((elementId) => {
+                            const element = action.payload.elements.find(
+                                (el) => el.id === elementId
+                            );
+                            element && selectedElements.push(element);
+                        });
+                    });
+
+                for (const element of selectedElements) {
```
msnyder-msft, line 74: I haven't looked in detail at what htis actually does, would it make any difference if there are duplicate meshIds? would it make sense to add the items to a set instead of an array?

msnyder-msft (commented):
This looks really good. I haven't played with the functionality but I'm totally behind the new reducer!!! 🥳🪅

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.tsx
pasanchMSFT: Wording here is tricky, I have a SET and RESET action, but I used the set to 'build' the outlined items out of meshIds and colors. I didn't want to make SET more complicated or just create a new reducer action that does the same thing as RESET. Do you think an extra action would be warranted here?

src/Models/Context/SceneViewContext/SceneViewContext.tsx
pasanchMSFT: changed this to a set

src/Models/Context/SceneViewContext/SceneViewContext.types.ts
```diff
@@ -10,7 +10,7 @@ export interface ISceneViewContext {
 }
 
 export interface ISceneViewContextProviderProps {
```
msnyder-msft, line 12: got a warning here for unused imports in this file

src/Components/ADT3DSceneBuilder/Internal/Elements/Elements.tsx
```diff
@@ -33,6 +33,12 @@ import PanelFooter from '../Shared/PanelFooter';
 import { IADTObjectColor } from '../../../../Models/Constants';
 import { deepCopy } from '../../../../Models/Services/Utils';
 import IllustrationMessage from '../../../IllustrationMessage/IllustrationMessage';
+import { useSceneViewContext } from '../../../../Models/Context/SceneViewContext/SceneViewContext';
+import {
+    ISceneViewContextState,
```
darsney, line 38: unused importsss

src/Components/3DV/SceneViewWrapper.tsx
```diff
@@ -183,6 +185,7 @@ const SceneViewWrapper: React.FC<ISceneViewWrapperProps> = (props) => {
                 objectStyle={sceneThemeState.objectStyle}
                 onCameraMove={addInProps?.onCameraMove ? cameraMove : undefined}
                 {...svp}
+                outlinedMeshitems={sceneViewState.outlinedMeshItems}
```
darsney, line 188: It would be good to add a story that illustrates how to use the properties that are set by the context.  In fact add the context to the existing stories that contains no properties just so we can add the context based properties to it as we remove them from props as implicit documentation of how the context works for consumers of SceneViewWrapper

msnyder-msft (approved):

PR #614, opened 2022-07-03
Zarmada/working/ADTS-252

miltonzarmada:

### DIFF
--- a/src/Components/OATPropertyEditor/PropertyListItemNest.tsx
+++ b/src/Components/OATPropertyEditor/PropertyListItemNest.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useMemo } from 'react';
 import { TextField, Text, IconButton } from '@fluentui/react';
 import {
     getPropertyInspectorStyles,
@@ -112,7 +112,7 @@ export const PropertyListItemNest = ({
         }
 
         switch (lastPropertyFocused.item.schema['@type']) {
-            case DTDLSchemaType.Object:
+            case 'object':
                 setPropertySelectorVisible(true);
                 return;
             case DTDLSchemaType.Enum:
@@ -159,7 +159,7 @@ export const PropertyListItemNest = ({
             );
         } else if (
             newModel[propertiesKeyName][parentIndex].schema['@type'] ===
-            DTDLSchemaType.Object
+            'object'
         ) {
             newModel[propertiesKeyName][parentIndex].schema.fields.splice(
                 index,
@@ -207,6 +207,16 @@ export const PropertyListItemNest = ({
         });
     };
 
+    const showObjectPropertySelector = useMemo(() => {
+        if (item.schema['@type'] === 'object') {
+            if (item.schema.fields.length === 0) {
+                return true;
+            } else if (item.schema.fields.length > 0 && collapsed) {
+                return true;
+            }
+        }
+    }, [collapsed]);
+
     return (
         <div
             className={propertyInspectorStyles.propertyListRelativeWrap}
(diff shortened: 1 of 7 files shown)


src/Components/OATPropertyEditor/PropertyListItemNest.tsx
```diff
@@ -159,7 +159,7 @@ export const PropertyListItemNest = ({
             );
         } else if (
             newModel[propertiesKeyName][parentIndex].schema['@type'] ===
-            DTDLSchemaType.Object
+            'object'
```
jamescarpinter, line 162: Should we be doing a case insensitive comparison here?

Does the DTDL spec require a specific casing? Seems weird we can't use the official enum here.

src/Components/OATPropertyEditor/PropertyListItemNest.tsx
```diff
@@ -207,6 +207,16 @@ export const PropertyListItemNest = ({
         });
     };
 
+    const showObjectPropertySelector = useMemo(() => {
+        if (item.schema['@type'] === 'object') {
+            if (item.schema.fields.length === 0) {
+                return true;
+            } else if (item.schema.fields.length > 0 && collapsed) {
+                return true;
+            }
```
jamescarpinter, line 216: We should be able to collapse this whole block into a return statement...

`return item.schema['@type'] === 'object' && (item.schema.fields.length === 0 || (item.schema.fields.length > 0 && collapsed));` 

Or something like that?

src/Components/OATPropertyEditor/PropertySelector.tsx
```diff
@@ -161,43 +161,41 @@ const PropertySelector = ({
...
-                                            className={
-                                                propertyInspectorStyles.propertyTag
-                                            }
-                                            src={tag.icon}
-                                            title={t(tag.title)}
-                                        ></Svg>
-                                    </ActionButton>
-                                );
-                            }
+                    {propertySelectorData.propertyTags.complex.map((tag, i) => {
```
jamescarpinter, line 164: This is a lot of logic to have inline.

Maybe we do a `.filter` before we do a `.map`? That way we don't need the `return <></>;` case?

src/Components/OATPropertyEditor/PropertySelector.tsx
```diff
@@ -161,43 +161,41 @@ const PropertySelector = ({
...
+                                <ActionButton
+                                    onClick={() => {
+                                        onTagClick(tag.name);
+                                    }}
+                                    onKeyPress={() => {
+                                        onTagClick(tag.name);
+                                    }}
+                                >
+                                    <Svg
+                                        tabIndex={0}
```
jamescarpinter, line 183: Should we be hardcoding a zero tab index here? Or should we be using `i`? Not sure tbh.

src/Models/Constants/Constants.ts
```diff
@@ -242,79 +242,79 @@ export const globeUrl =
...
+            },
+            {
+                name: DTDLSchemaType.Enum,
+                title: 'OATPropertyEditor.enum',
+                icon: IconEnum,
+                complex: true
             }
         ],
-        sectionThird: [
+        geoSpatial: [
```
jamescarpinter, line 317: The `S` shouldn't be a capital in geospatial.

jamescarpinter (changes requested):

src/Components/OATPropertyEditor/PropertySelector.tsx
```diff
@@ -161,77 +161,61 @@ const PropertySelector = ({
...
-                                        title={t(tag.title)}
-                                    ></Svg>
-                                </ActionButton>
-                            );
-                        }
-                    })}
+                            lastPropertyFocused.item.schema['@type'] ===
+                                'object'
+                                ? tag.type === 'object'
+                                : tag.type !== 'object'
```
jamescarpinter, line 170: Should this use the enum const?

jamescarpinter (approved):
One question inline, else looks good to me.

PR #615, opened 2022-07-03
Zarmada/working/ADTS-254

miltonzarmada:


src/Components/OATModelList/OATModelList.tsx
```diff
@@ -116,6 +116,16 @@ const OATModelList = ({
         setFilter(evt.target.value);
     };
 
+    const getDisplayNameText = (item) => {
+        return typeof item['displayName'] === 'string'
```
jamescarpinter, line 120: Rather than indexing into item to get display name each time, maybe pull it out first?

E.g. `const displayName = item['displayName'];`.

jamescarpinter (changes requested):

jamescarpinter (approved):

PR #620, opened 2022-07-05
zarmada/working/ADTS-257

miltonzarmada:

### DIFF
--- a/src/Components/OATGraphViewer/OATGraphViewer.tsx
+++ b/src/Components/OATGraphViewer/OATGraphViewer.tsx
@@ -253,7 +253,6 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
     const { t } = useTranslation();
     const theme = useTheme();
     const reactFlowWrapperRef = useRef(null);
-    const newModelButtonRef = useRef(null);
     const [elements, setElements] = useState(
         getGraphViewerElementsFromModels(models, modelPositions)
     );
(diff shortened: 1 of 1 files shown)


src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -637,15 +639,45 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
...
+        return getNewNodePosition(newCoordinates);
+    };
+
     const onNewModelClick = () => {
         if (!state.modified) {
             // Create a new floating node
             const name = `Model${newModelId}`;
             const id = `${idClassBase}model${newModelId};${versionClassBase}`;
+            let startPositionCoordinates = null;
+            if (newModelButtonRef.current) {
```
jamescarpinter, line 670: If current is null, then startPositionCoordinates will be null too. What happens if we then call getNewNodePosition with a null value for the param?

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -1101,6 +1113,7 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
                             styles={buttonStyles}
                             onClick={onNewModelClick}
                             text={t('OATGraphViewer.newModel')}
+                            elementRef={newModelButtonRef}
```
jamescarpinter, line 1116: Rather than add another ref, could we get the element reference from the event args passed to onClick?

jamescarpinter (changes requested):

jamescarpinter (approved):

PR #623, opened 2022-07-06
Story demonstrating gizmo action using SceneView

BrennaCourtney:
### Summary of changes 🔍 
Part 1 of integrating transforms exploration: this chunk creates a story showing off manipulating elements with meshes in SceneView. 
image

The bulk of the code here is a useEffect for a new SceneView prop, gizmoElementItem of type TransformElementItem, that determines which element in a scene gets a gizmo attached to it. 
image

There is also a useEffect for another SceneView prop, gizmoTransformItem, that is meant to capture position/rotation data to be consumed by a builder panel or other React component outside of SceneView -- here, it is only used to print position/rotation data.
image

In the story, gizmoElementItem is hardcoded, but when "actually" used, it should be determined by the elements attached to the behavior being edited. It is still unclear how to manipulate multiple elements in a single view (at the moment, only a single element -- the first one -- is supported). Options include point and click or selecting from a drop down. 

### Testing 🧪
https://t-brcourtney-gizmo-sceneview-story--601c6b2fcd385c002100f14c.chromatic.com/?path=/story/components-sceneview--gizmo

Head to Components -> SceneView -> Gizmo. The second tank from the left should be manipulatable (it consists of two meshes). Open the inspector to see the changed position and rotation on drag end (when you let go of the gizmo).

You should see the X Y Z values in the textfields being updated on drag end. You can also edit these values and see the element manipulated in real time. 

There is also a checkbox that toggles "transform mode." This is meant to simulate exiting the transforms tab and returning to it. When unchecked, the X Y Z text fields and gizmo should disappear and the element return to its initial position. When checked again, the old position/rotation should resume.

### Checklist ✔️
- [x] Added relevant labels


src/Components/3DV/SceneView.tsx
```diff
@@ -1653,6 +1660,205 @@ function SceneView(props: ISceneViewProps, ref) {
...
+            } catch {
+                console.warn('unable to add gizmo to mesh');
+            }
+            if (debugLogging) {
+                console.timeEnd('adding gizmo to meshes');
+            }
+        }
+
+        return () => {
+            debugLog('debug', 'Mesh gizmo cleanup');
```
ptallettms, line 1823: A common problem with SceneView is cacheing objects when SceneView is recycled. I suspect if you switch models (scenes) you will get a crash as the Babylon context will be reset, but the code will try and use the gizmo stuff from the previous model. See the ADTViewer/ZoomAndColor story where you can switch scenes to test this (add some gizmo stuff to this story to test).

I think you need to add some cleanup code to this effect (eg clear gizmoManagerRef and gizmoTransformItemRef so that they get recreated)

BrennaCourtney: I checked it out, and clearing gizmoManagerRef and gizmoTransformItemRef actually causes issues with switching in and out of the tab that enables the transform. Based on our earlier conversation and on the ADTViewer story, though, I know that the user's application doesn't always consume the builder -- what if I added the scene as a dependency to the useEffect?

src/Components/3DV/SceneView.tsx
```diff
@@ -1653,6 +1660,205 @@ function SceneView(props: ISceneViewProps, ref) {
...
 
+    // Handle gizmoElementItem
+    useEffect(() => {
+        debugLog(
+            'debug',
+            'adding gizmo to element based on gizmoElementItem prop' +
+                (scene ? ' with scene' : ' no scene')
+        );
+
+        if (scene && !isLoading) {
```
pasanchMSFT, line 1671: could this effect be run only if gizmoElementItem is defined?

src/Components/3DV/SceneView.tsx
```diff
@@ -1653,6 +1660,205 @@ function SceneView(props: ISceneViewProps, ref) {
...
+                if (!gizmoManagerRef.current) {
+                    gizmoManagerRef.current = new BABYLON.GizmoManager(
+                        scene,
+                        1,
+                        utilLayer.current
+                    );
+                }
+                const gizmoManager = gizmoManagerRef.current;
+
+                if (!gizmoElementItem) {
```
pasanchMSFT, line 1686: The contents of this if block are a good candidate for the cleanup + removing the gizmo manager refs.

pasanchMSFT (commented):
Left some comments, overall great work!

src/Components/3DV/SceneView.tsx
```diff
@@ -1653,6 +1660,205 @@ function SceneView(props: ISceneViewProps, ref) {
...
+                            };
+                        }
+
+                        setGizmoTransformItem(
+                            gizmoTransformItemRef.current.transform
+                        );
+                    });
+
+                    // update the gizmoTransformItem (allows builder panel to display new position/rotation)
+                    const updateTransform = () => {
```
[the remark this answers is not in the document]
BrennaCourtney: done

src/Components/3DV/SceneView.tsx
BrennaCourtney: done

src/Components/3DV/SceneView.tsx
```diff
@@ -1653,6 +1660,205 @@ function SceneView(props: ISceneViewProps, ref) {
...
+            } catch {
+                console.warn('unable to add gizmo to mesh');
+            }
+            if (debugLogging) {
+                console.timeEnd('adding gizmo to meshes');
+            }
+        }
+
+        return () => {
+            debugLog('debug', 'Mesh gizmo cleanup');
```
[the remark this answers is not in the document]
BrennaCourtney: done

src/Components/3DV/SceneView.tsx
BrennaCourtney: still need this logic for switching in and out of transforms mode, but we discussed the clean up together - done

src/Components/3DV/SceneView.tsx
BrennaCourtney: done

abalogun-dev (approved):

BrennaCourtney:
> This is a nit comment, but the table that displays the transforms is quite low on the screen. Move it up a bit so at least the table header is peeking before scrolling down so ppl know the table is below

done!

src/Components/3DV/SceneView.stories.tsx
```diff
@@ -16,199 +22,202 @@ export default {
...
-            id: 'id' + createGUID(),
-            attachedMeshIds: ['Cube.003'],
-            UIElement: <ModelLabel label={'Marker 1'} />,
-            showIfOccluded: true
-        },
+export const Globe = Template.bind({}) as SceneViewStory;
+const globeMarkers: Marker[] = [
+    {
+        id: 'ID' + createGUID(),
+        name: 'Ibhayi',
```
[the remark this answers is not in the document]
BrennaCourtney: ah that code actually isn't mine for the globe -- decided not to refactor any of the existing SceneView stories so those should all be the same

src/Components/3DV/SceneView.stories.tsx
```diff
@@ -16,199 +22,202 @@ export default {
...
+        id: 'ID' + createGUID(),
+        name: 'Uberlandia',
+        UIElement: <ModelLabel label="Uberlandia" />,
+        latitude: -18.9231,
+        longitude: -48.2886
+    }
+];
+const meshClick = (mesh: any, e: any) => {
+    if (!mesh && !e) {
+        console.log('Hello');
```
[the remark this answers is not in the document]
BrennaCourtney: Globe story should just be showing markers -- things look fine on my end? again tho globe isn't mine so if there are bugs there maybe it should be opened as a separate issue

src/Components/3DV/SceneView.stories.tsx
```diff
@@ -16,6 +35,231 @@ export default {
...
+                gizmoTransformItem.rotation
+            );
+        },
+        []
+    );
+
+    const setGizmoElementItem = useCallback(
+        (gizmoElementItem: TransformedElementItem) => {
+            dispatch({
+                type: SET_ELEMENT_TO_GIZMO,
```
iremgokce, line 69: nit: just to be consistent with the action names with the one above maybe we can rename this?
```suggestion
                type: SET_GIZMO_ELEMENT_ITEM,
```

src/Components/3DV/SceneView.stories.tsx
```diff
@@ -16,6 +35,231 @@ export default {
     }
 };
 
+const defaultGizmoElementItems: TransformedElementItem = {
```
iremgokce, line 38: nit
```suggestion
const defaultGizmoElementItem: TransformedElementItem = {
```

src/Components/3DV/SceneView.stories.tsx
```diff
@@ -16,6 +35,231 @@ export default {
...
+    meshIds: ['tank6_LOD0.003_primitive0', 'tank6_LOD0.003_primitive1'],
+    parentMeshId: 'tank6_LOD0.003_primitive0'
+};
+
+export const Gizmo = () => {
+    const [state, dispatch] = useReducer(
```
iremgokce, line 44: Hmm, maybe we can move all the logic and the returned dom here to a separate standalone wrapper component called like `SceneViewWithGizmo` or something? Having lots of things happening in a story may not be ideal. What do you think?

src/Components/3DV/SceneView.stories.tsx
```diff
@@ -16,6 +35,231 @@ export default {
...
+    );
+};
+
+interface IFakePanelProps {
+    gizmoTransformItem: TransformInfo;
+    setGizmoTransformItem?: (gizmoTransformItem: TransformInfo) => void;
+}
+
+const FakePanel = (props: IFakePanelProps) => {
+    const { gizmoTransformItem, setGizmoTransformItem } = props;
```
iremgokce, line 122: and moving all those internal components and interface to that separate wrapper component as in my comment above?

src/Components/3DV/SceneView.stories.tsx
BrennaCourtney: Yeah that might be good -- would I just generate a new component (complete with types, stories, etc.) or just move it all the a standalone .tsx file?

src/Components/3DV/SceneView.stories.tsx
```diff
@@ -16,6 +35,231 @@ export default {
...
+    }, [gizmoTransformItem]);
+
+    const onTransformChange = useCallback(
+        (event) => {
+            if (gizmoTransformItem) {
+                const { name, value } = event.target;
+                gizmoTransformItemDraftRef.current = deepCopy(
+                    gizmoTransformItem
+                );
+                const valueAsNumber = value;
```
iremgokce, line 138: what is the reason for this new variable declaration? can't we just use the `value` for below assignments?

src/Components/3DV/SceneView.stories.tsx
BrennaCourtney: done

src/Components/3DV/SceneView.stories.tsx
BrennaCourtney: done

src/Components/3DV/SceneView.stories.tsx
```diff
@@ -16,6 +35,231 @@ export default {
...
+                z: Number(gizmoTransformItem.rotation.z)
+            }
+        });
+    }, [gizmoTransformItem]);
+
+    return (
+        <div style={{ padding: '20px' }}>
+            <Stack tokens={{ childrenGap: 12 }} horizontal>
+                <TextField
+                    label="X Position: "
```
iremgokce, line 184: just to keep in mind, we might need to localize the labels when you consume this component (maybe you already did :))

src/Components/3DV/SceneView.stories.tsx
BrennaCourtney: ah that was leftover from something i was doing before -- will fix

src/Components/3DV/SceneView.stories.tsx
BrennaCourtney: oh yeah I actually wasn't even thinking this component would be consumed; just used as a placeholder for now -- in the actual tab it will probably be more like X: Y: Z: with a "Position" label on top or something

src/Components/3DV/SceneView.stories.tsx
```diff
@@ -16,6 +35,231 @@ export default {
...
+                    name="xPos"
+                    type="number"
+                    onChange={onTransformChange}
+                    onBlur={onTransformBlur}
+                    value={
+                        gizmoTransformItem
+                            ? '' + gizmoTransformItem.position.x
```
iremgokce, line 191: hmm we should not need to cast it to string, the number values should be working in a text field value, no?

BrennaCourtney: image
I get a type error :( says number cannot be cast to string

src/Components/3DV/SceneView.stories.tsx
BrennaCourtney: done

src/Components/3DV/SceneView.stories.tsx
BrennaCourtney: done

iremgokce:
looks good to me, great work!

iremgokce (approved):

src/Models/Classes/SceneView.types.ts
```diff
@@ -107,6 +131,8 @@ export interface ISceneViewProps {
     cameraPosition?: ICameraPosition;
     coloredMeshItems?: CustomMeshItem[];
     getToken?: () => Promise<string>;
+    gizmoElementItem?: TransformedElementItem;
```
msnyder-msft, line 134: if one of these gizmo props is provided, are all of them required? or are each independent?

src/Components/3DV/SceneViewWithGizmoWrapper.tsx
```diff
@@ -0,0 +1,252 @@
+import React, {
```
msnyder-msft, line 1: does this get used somewhere in the production app? or is it only for stories?

src/Components/3DV/SceneView.tsx
```diff
@@ -1673,6 +1680,212 @@ function SceneView(props: ISceneViewProps, ref) {
...
+            // only cleanup if exiting edit behavior, not when switching tabs
+            if (!gizmoElementItem) {
+                gizmoTransformItemDraftRef.current = null;
+            }
+            debugLog('debug', 'Mesh gizmo cleanup');
+        };
+    }, [scene, gizmoElementItem, isLoading]);
+
+    // Handle gizmoTransformItem
+    useEffect(() => {
```
msnyder-msft, line 1859: so if I understand correctly, this is for specifically handling the text box inputs in the left panel. Is there a reason that this requires a separate set of state? Could we pass in the position of the object the same way for the manual inputs as well as the click and drag?

PR #628, opened 2022-07-06
Added/updated tests for markers with and without grouping

samjosephgates:
### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/3DV/SceneView.stories.tsx
+++ b/src/Components/3DV/SceneView.stories.tsx
@@ -23,63 +23,72 @@ export const Globe = () => {
             name: 'Ibhayi',
             UIElement: <ModelLabel label="Ibhayi" />,
             latitude: -33.872,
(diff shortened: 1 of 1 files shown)


BrennaCourtney:
image
Clicking the marker when the globe is still spinning causes the tooltip to appear in the top left corner -- not a huge issue and probably not relevant to this PR but might just have to do with the way the location is calculated only after stop?

pasanchMSFT:
> image Clicking the marker when the globe is still spinning causes the tooltip to appear in the top left corner -- not a huge issue and probably not relevant to this PR but might just have to do with the way the location is calculated only after stop?

@BrennaCourtney I think this needs to be a separate issue. Could you create a Github issue so we can track that visual bug and fix it later?

ptallettms (approved):

PR #629, opened 2022-07-06
Zarmada/working/ADTS-268

miltonzarmada:
> 267: Fix:Importing an untargeted relationship fails DTDL parser 
> 269: Fix: Sometime untargeted relationships wont get deleted

### DIFF
--- a/src/Components/OATHeader/OATHeader.tsx
+++ b/src/Components/OATHeader/OATHeader.tsx
@@ -66,7 +66,6 @@ const OATHeader = ({ elements, dispatch, state }: OATHeaderProps) => {
         namespace
     } = state;
     const uploadInputRef = useRef(null);
-    const keyDown = useRef(false);
     const redoButtonRef = useRef(null);
     const undoButtonRef = useRef(null);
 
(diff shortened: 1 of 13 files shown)


src/Components/OATPropertyEditor/PropertySelector.tsx
```diff
@@ -191,17 +191,17 @@ const PropertySelector = ({
                 <Separator styles={propertySelectorSeparatorStyles} vertical />
                 <div className={propertyInspectorStyles.propertyTagsWrapFirst}>
                     {propertySelectorData.propertyTags.primitive.map(
-                        (tag, i) => (
+                        (tag, index) => (
```
jamescarpinter, line 194: Index isn't a great key. Ideally we use an actual identifying property on the tag.

jamescarpinter (changes requested):

miltonzarmada:
Added ability to undo/redo via keyboard @jamie @jamescarpinter

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -394,6 +399,48 @@ const OATHeader = ({ elements, dispatch, state }: OATHeaderProps) => {
         onFilesUpload(acceptedFiles);
     }, [acceptedFiles]);
 
+    const onKeyDown = (e) => {
+        if (keyDown.current) {
+            // Prevent keydown from being triggered twice
+            return;
```
jamescarpinter, line 405: Could we use this property on the event to avoid having to use a ref to hold this state? https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -394,6 +399,48 @@ const OATHeader = ({ elements, dispatch, state }: OATHeaderProps) => {
...
+            keyDown.current = true;
+            redoButtonRef.current['_onClick']();
+        }
+    };
+
+    useEffect(() => {
+        // Set listener to undo/redo buttons on key press
+        document.addEventListener('keydown', (e) => onKeyDown(e));
+        return () => {
+            document.removeEventListener('keydown', (e) => onKeyDown(e));
```
jamescarpinter, line 427: Does this unsubscribe the event handler properly? I.e. we're redefining the handler function here. Does the redefined function === the function we create when we subscribe? To solve this I think we can just pass `onKeyDown` rather than create an inline function.

src/Components/OATPropertyEditor/Utils.ts
```diff
@@ -25,6 +25,14 @@ export const getModelPropertyListItemName = (name) => {
     return '';
 };
 
+export const isDisplayNameDefined = (name) => {
+    return name && typeof name === 'string' && name.length > 0
+        ? true
+        : name && Object.values(name)[0].length > 0
+        ? true
+        : false;
```
jamescarpinter, line 33: I think we can simplify this further... I don't think ternary expressions should really be used to return true/false.

E.g. `return (name && typeof name === 'string' && name.length > 0) || (name && Object.values(name)[0].length > 0);` should be equivalent?

jamescarpinter (changes requested):

jamescarpinter (approved):

PR #633, opened 2022-07-08
Make meshes tab appear required

abalogun-dev:
### Summary of changes 🔍 
When no meshes are assigned to an element, i made it so there is a little red dot next to meshes tab to show that it is required. I also moved the custom rendering function that makes that happen into its own component under the shared folder

### Testing 🧪
Try to make a new element and see if the meshes tab has a red dot next to it, then add some meshes to it and make sure it no longer appears.

Also, do the same steps for behaviors (this is to ensure that my refactor didn't break anything) 

https://msazure.visualstudio.com/One/_workitems/edit/14903848/

### Checklist ✔️
- [ x ] Linked associated issue (if present)
- [ ] Added relevant labels


BrennaCourtney (approved):

src/Components/ADT3DSceneBuilder/Internal/Shared/MakePivotRequired.tsx
```diff
@@ -0,0 +1,22 @@
+import { IPivotItemProps } from '@fluentui/react';
+import React from 'react';
+import { customPivotItemStyles } from '../Behaviors/BehaviorsForm.styles';
+
+export function setPivotToRequired(
+    isValid: boolean | undefined,
+    link?: IPivotItemProps,
```
pasanchMSFT, line 7: Could you change the name of this argument to linkProps? To clarify its purpose.

src/Components/ADT3DSceneBuilder/Internal/Shared/MakePivotRequired.tsx
```diff
@@ -0,0 +1,22 @@
...
+    defaultRenderer?: (link?: IPivotItemProps) => JSX.Element | null
```
pasanchMSFT, line 8: same thing as above for this one.

src/Components/ADT3DSceneBuilder/Internal/Shared/MakePivotRequired.tsx
abalogun-dev: that's the naming convention for the customRendering function for Pivot components. Just wanted to stick to that so it isn't confusing for the next person

pasanchMSFT (approved):

src/Components/ADT3DSceneBuilder/Internal/Shared/MakePivotRequired.tsx
```diff
@@ -1,21 +1,25 @@
...
         return null;
     }
     return (
         <span className={customPivotItemStyles.root}>
             {defaultRenderer({ ...link, itemIcon: undefined })}
-            {/* TODO: Add an aria label of some kind here for screen readers to see this error state */}
             {isValid === false && (
-                <span className={customPivotItemStyles.alert} />
+                <span
+                    aria-label={t('meshIsRequired')}
```
darsney, line 20: This actually isnt a correct aria label since this can be used in contexts other than the elements tab.  I think a more appropriate label is probably "This field is required" or something, but if we feel that people should be able to customize it, they should use a prop to pass in the aria label

src/Components/ADT3DSceneBuilder/Internal/Shared/MakePivotRequired.tsx
```diff
@@ -0,0 +1,26 @@
+import { IPivotItemProps } from '@fluentui/react';
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { customPivotItemStyles } from '../Behaviors/BehaviorsForm.styles';
+
+export function setPivotToRequired(
```
darsney, line 6: It may be better to place this function in a place outside of ADT3DSceneBuilder just because the Pivot is a component that we stand to use in places other than just 3D Scenes.  That is, a UI for making a pivot appear required is a concept that is something we are likely to want to use across cardboard, not just in 3D Scenes.  My personal preference is src/Theming/FluentComponentStyles/Pivot.styles.ts, but open to other options.  Regardless, I think it should probably move to a place not restricted to 3D Scenes

src/Components/ADT3DSceneBuilder/Internal/Shared/MakePivotRequired.tsx
```diff
@@ -0,0 +1,26 @@
...
+    isValid: boolean | undefined,
+    link?: IPivotItemProps,
+    defaultRenderer?: (link?: IPivotItemProps) => JSX.Element | null
+): JSX.Element | null {
+    const { t } = useTranslation();
+    if (!link || !defaultRenderer) {
+        return null;
+    }
+    return (
+        <span className={customPivotItemStyles.root}>
```
darsney, line 16: customPivotItemStyles should be part of this file, not imported from BehaviorsForm

PR #637, opened 2022-07-12
Zarmada/working/ADTS-271

miltonzarmada:

### DIFF
--- a/src/Components/OATModelList/OATModelList.tsx
+++ b/src/Components/OATModelList/OATModelList.tsx
@@ -82,11 +82,10 @@ const OATModelList = ({
     useEffect(() => {
         if (model) {
             currentNodeId.current = model['@id'];
-            setItems([...elements]);
         } else {
             currentNodeId.current = null;
-            setItems([...elements]);
         }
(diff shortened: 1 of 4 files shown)


src/Components/OATModelList/OATModelList.tsx
```diff
@@ -79,6 +79,15 @@ const OATModelList = ({
         );
     }, [filter]);
 
+    useEffect(() => {
+        if (model) {
+            currentNodeId.current = model['@id'];
+        } else {
+            currentNodeId.current = null;
```
jamescarpinter, line 86: Could we do `currentNodeId.current = model ? model['@id'] : null;`?

src/Components/OATModelList/OATModelList.tsx
```diff
@@ -79,6 +79,15 @@ const OATModelList = ({
...
+        }
+        setItems([...elements]);
```
jamescarpinter, line 88: Why does items change on model change?

jamescarpinter (changes requested):

PR #639, opened 2022-07-12
Zarmada/working/ADTS-271

miltonzarmada:

### DIFF
--- a/src/Components/OATGraphViewer/OATGraphViewer.tsx
+++ b/src/Components/OATGraphViewer/OATGraphViewer.tsx
@@ -865,7 +865,9 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
                     const untargetedRelationship = {
                         '@type': OATRelationshipHandleName,
                         '@id': id,
-                        name: '',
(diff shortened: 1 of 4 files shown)


miltonzarmada:
`Why does items change on model change?`
@jamescarpinter items are changed on model change, to trigger all ModelList items to be re-rendered. So that it applies border styles changes to the selected node/model list item. Without it when selecting the graphviewer's background (setting model to null), no changes appear on MoldelList, making it look like there is still a Model selected.

Perhaps there is a better approach to trigger this re-render

jamescarpinter:
> `Why does items change on model change?` @jamescarpinter items are changed on model change, to trigger all ModelList items to be re-rendered. So that it applies border styles changes to the selected node/model list item. Without it when selecting the graphviewer's background (setting model to null), no changes appear on MoldelList, making it look like there is still a Model selected.
> 
> Perhaps there is a better approach to trigger this re-render

Got it, thanks Milton. I think we have a code comment somewhere else in the component explaining this issue. Can you copy the comment here as well?

jamescarpinter (approved):

PR #641, opened 2022-07-14
Fix: Model file upload correctly creates new scene

pasanchMSFT:
### Summary of changes 🔍 
In this PR changes were made that fix an issue which caused the creation of a new scene to fail in the case of uploading a file. Files were being added to config file as scenes, not as part of a scene. With this change, that should not occur anymore.

### Testing 🧪
**Test this by running this branch locally**
- Create a new scene, upload a file to create the new scene.
- Your new scene should be shown in the scene list and in your config file.

### Checklist ✔️
- [ ] Added relevant labels


BrennaCourtney (approved):
works for me! (both mock & adt 3d scenes)

darsney (approved):
LGTM, nice fixes!

PR #647, opened 2022-07-22
fixed error where PI could not read fields of properties that were enums

BrennaCourtney:
### Summary of changes 🔍 
Error in Property Inspector that showed up in the Explorer:
image

After investigation, discovered that it was caused by fields of properties that are enums not being read properly. Here is the same twin in cardboard after the fix: 
image
tempsnip

After this fix goes in, Explorer will need to be updated to consume the most recent version of cardboard.

### Testing 🧪
*** IF you would prefer not to clone these models yourself, just ping me and we can walk through using my ADT instead
Repro error in Explorer: 
1. git clone this repo: https://github.com/RealEstateCore/BRECK-DTDL.git
2. do git checkout 737a1c8b63 (will revert to dtdl v2)
3. go to https://explorer.digitaltwins.azure.net/
4. Upload folder "Point" (or, if you don't want to upload all 1000 models, upload the following files: Point.json, Sensor.json, Power_Sensor.json, Electric_Power_Sensor.json, and Active_Power_Sensor.json)
5. Select "Active Power Sensor" under models and create a twin; call it "ActivePowerSensorTwin" 
image
6. Click twin in graph and see error as above

Repro error in cardboard:
1. Go to PropertyInspector.stories.local.tsx and change twin to "ActivePowerSensorTwin" 
image
2. Go to Storybook -> Property Inspector -> Adt Twin
3. You should see a similar error as in Explorer

Test this fix:
1. Pull this branch and go to Storybook -> Property Inspector -> Adt Twin
2. You should see the fields showing up properly as above; try editing and saving
3. Property Inspector in all other locations (ADT Explorer, Mock) should work the same

This PR also fixes error where Property Inspector dropdown has same color for text and background when property is unedited: 
image
Should now look like:
image
(And working with all themes)

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
+++ b/src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
@@ -43,7 +43,7 @@ export const AdtTwin = (args, { globals: { theme, locale } }) => {
 AdtTwin.argTypes = {
     twinId: {
         control: { type: 'text' },
-        defaultValue: 'ActivePowerSensorTwin'
+        defaultValue: 'PasteurizationMachine_A01'
     }
 };
 
--- a/src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
+++ b/src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
@@ -43,7 +43,7 @@ export const AdtTwin = (args, { globals: { theme, locale } }) => {
 AdtTwin.argTypes = {
     twinId: {
         control: { type: 'text' },
-        defaultValue: 'PasteurizationMachine_A01'
+        defaultValue: 'ActivePowerSensorTwin'
     }
 };
 
--- a/src/Components/PropertyInspector/PropertyInspectoryModel.ts
+++ b/src/Components/PropertyInspector/PropertyInspectoryModel.ts
@@ -656,8 +656,8 @@ abstract class PropertyInspectorModel {
 
     static conformSyntax = (inputEl) => {
         // Update syntax of all object keys
-        if (!inputEl) return;
         if (typeof inputEl === 'object') {
+            if (!inputEl) return;
             Object.keys(inputEl).forEach((key) => {
                 // Update to simplified DTDL syntax (conform to our interface)
                 if (key in dtdlSyntaxMap) {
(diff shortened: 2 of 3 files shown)


src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
```diff
@@ -43,7 +43,7 @@ export const AdtTwin = (args, { globals: { theme, locale } }) => {
 AdtTwin.argTypes = {
     twinId: {
         control: { type: 'text' },
-        defaultValue: 'PasteurizationMachine_A01'
+        defaultValue: 'ActivePowerSensorTwin'
```
abalogun-dev, line 46: wouldn't you want to revert the value back. The ActivePowerSensor was just put there just so you can repro the error locally, correct?

abalogun-dev (approved):

src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
BrennaCourtney: done!

msnyder-msft (approved):
Seems like the logic makes sense to me.

pasanchMSFT (approved):
LGTM

src/Components/PropertyInspector/PropertyInspectoryModel.ts
```diff
@@ -646,53 +650,61 @@ abstract class PropertyInspectorModel {
...
+    static replaceKeyInObj = (obj, oldKey, newKey) => {
+        delete Object.assign(obj, { [newKey]: obj[oldKey] })[oldKey];
+    };
 
-        const replaceKeyInObj = (obj, oldKey, newKey) => {
-            delete Object.assign(obj, { [newKey]: obj[oldKey] })[oldKey];
-        };
+    static conformSyntax = (inputEl) => {
+        // Update syntax of all object keys
+        if (typeof inputEl === 'object') {
```
iremgokce, line 659: inputEl might be `null` which is also an object type. not sure if that would be a case but maybe handling null might be good before passing it to Object.keys.

iremgokce (changes requested):
only one comment about handling null case in type checking for 'object'

src/Components/PropertyInspector/PropertyInspectoryModel.ts
BrennaCourtney: done!

PR #648, opened 2022-07-26
zarmada/working/ADTS-276

miltonzarmada:
-- 
ADTS-276
> Enable valueSchema edit 
> Enable nested enum values edit
> Remove unnecesary attibutes for enum values
> Remove unused modal component 

--
ADTS-275
> Validate silge file upload relationships
--

ADTS-274
> Fix: Graph Controls move when open dialog


src/Components/OATHeader/OATHeader.tsx
```diff
@@ -219,21 +225,66 @@ const OATHeader = ({ dispatch, state }: OATHeaderProps) => {
...
+    const validateSingleFileRelationships = async (
+        file,
+        newFilesErrors,
+        newFiles
+    ) => {
+        if (file) {
+            let errorFound = false;
+            // Validate contents
+            let model = await file.text();
+            model = JSON.parse(model);
```
jamescarpinter, line 237: Do we need a try/catch here? I.e. in case it's not valid JSON?

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -219,21 +225,66 @@ const OATHeader = ({ dispatch, state }: OATHeaderProps) => {
...
+                        content['@type'] &&
+                        content['@type'] === OATRelationshipHandleName
+                    ) {
+                        newFilesErrors.push(
+                            t('OATHeader.errorSingleFileHoldsRelationships', {
+                                fileName: file.name
+                            })
+                        );
+                        errorFound = true;
+                        break;
```
jamescarpinter, line 251: I think the issue @jamiewilbraham reported is more subtle than "error on relationship found". I.e. if the graph already has the target model then this isn't an error. Same goes for untargeted relationships.

We also have to consider similar errors for components and extends.

I think the best option would be to use the DTDL validator. I.e. take the models that are currently in the graph, add the model we're importing, and run validate. This suggestion assumes that the DTDL validator checks relationships are valid (I'm not sure). Otherwise we might need to write more extensive validation code.

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -219,21 +225,66 @@ const OATHeader = ({ dispatch, state }: OATHeaderProps) => {
...
                 newFilesErrors.push(
                     t('OATHeader.errorFileFormatNotSupported', {
-                        fileName: sF.name
+                        fileName: file.name
                     })
                 );
             }
-            sF = new File([], '');
-        });
+            file = new File([], '');
```
jamescarpinter, line 285: What does this do?

src/Components/OATPropertyEditor/FormUpdateProperty.tsx
```diff
@@ -220,16 +231,25 @@ export const FormUpdateProperty = ({
...
                 schema: activeNestedProperty.schema,
                 name: activeNestedProperty.name
             };
 
+            // Update nested enum value
+            if (
+                activeNestedProperty &&
+                activeProperty.schema['@type'] === 'Enum'
+            ) {
+                delete prop.displayName;
```
jamescarpinter, line 243: Why is display name being removed? It's a valid DTDL prop right?

src/Components/OATPropertyEditor/Utils.ts
```diff
@@ -26,6 +26,13 @@ export const getModelPropertyCollectionName = (type: string | string[]) => {
     return 'contents';
 };
 
+export const getNestedPropertyCollectionName = (type: string | string[]) => {
+    if (type && type === 'Enum') {
+        return 'enumValues';
+    }
+    return 'fields';
```
jamescarpinter, line 33: Let's turn this into a ternary expression.

src/Resources/Locales/cs.json
```diff
@@ -1,915 +1,917 @@
 {
-    "languageName": "Czech",
```
jamescarpinter, line 2: It looks like the change has altered all of the tab indents for these files? Was this intentional?

jamescarpinter (changes requested):

src/Components/OATPropertyEditor/FormUpdateProperty.tsx
miltonzarmada: It is a valid prop, it's an optional one. From the card, I understood we would take it off. 
Let me know if I should re add it for the update modal? @jamescarpinter @jamiewilbraham

src/Resources/Locales/cs.json
miltonzarmada: @jamescarpinter I believe it now has the same indent as the other language files (except english)

PR #649, opened 2022-07-26
Zarmada/working/ADTS-273

miltonzarmada:


src/Models/Constants/Interfaces.ts
```diff
@@ -1,4 +1,4 @@
-import { AbstractMesh, Scene } from '@babylonjs/core';
+import { AbstractMesh, int, Scene } from '@babylonjs/core';
```
jamescarpinter, line 1: I don't think this is referenced anywhere in the file. So I don't think we need to add it here?

jamescarpinter (changes requested):

jamescarpinter (approved):

PR #651, opened 2022-07-27
Zarmada/working/ADTS-275

miltonzarmada:
### Summary of changes 🔍


src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -113,24 +113,34 @@ const OATGraphViewer = ({ state, dispatch }: OATGraphProps) => {
             if (input.contents) {
                 // Get the relationships
                 input.contents.forEach((content) => {
+                    const foundComponentTarget = models.find(
+                        (model) => model['@id'] === content.schema
+                    );
```
jamescarpinter, line 118: We should probably only do this if we're dealing with a component? I.e. under the `case` statement for component.

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -316,34 +313,55 @@ const OATHeader = ({ dispatch, state }: OATHeaderProps) => {
...
-        const items = [];
+        const newModels = [];
         if (files.length > 0) {
             const filesErrors = [];
             let modelsMetadataReference = null;
             for (const current of files) {
                 const content = await current.text();
-                const error = await parseModel(content);
-                if (!error) {
+                const JSONContent = JSON.parse(content);
```
jamescarpinter, line 322: I think `jsonContent` would be the right capitalization here.

jamescarpinter: Also, what happens if the file isn't valid JSON? Do we catch it somewhere?

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -316,34 +313,55 @@ const OATHeader = ({ dispatch, state }: OATHeaderProps) => {
...
+                newModels.push(JSONContent);
+            }
+
+            const combinedModels = [...models, ...newModels];
+            const error = await parseModels(combinedModels);
+
+            if (!error) {
+                console.log('No error');
```
jamescarpinter, line 330: Remove console log.

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -316,34 +313,55 @@ const OATHeader = ({ dispatch, state }: OATHeaderProps) => {
...
+                let loopIndex = 0;
+                for (const current of files) {
```
jamescarpinter, line 332: It might be cleaning to use a traditional `for (let i = 0; i <...` rather than have the `loopIndex` as well.

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -316,34 +313,55 @@ const OATHeader = ({ dispatch, state }: OATHeaderProps) => {
...
+
             if (filesErrors.length === 0) {
+                const modelsCopy = deepCopy(models);
+                // Add new models to models copy
+                for (const model of newModels) {
+                    // Check if model already exists
+                    const modelExists = modelsCopy.find(
+                        (m) => m['@id'] === model['@id']
+                    );
+                    if (!modelExists) {
```
jamescarpinter, line 357: I think we should do this check after parsing. I.e. this is actually an error and we should add a message to the errors collection and stop.

src/Models/Services/Utils.ts
```diff
@@ -390,6 +390,23 @@ export async function parseModel(modelJson: string) {
     }
 }
 
+export async function parseModels(models) {
```
jamescarpinter, line 393: Can we remove the old `parseModel` function from this file?

jamescarpinter: Can we type `models`? It should be the same type as what's on the reducer. Array of `IDtdlInterface` from memory?

jamescarpinter (changes requested):

src/Components/OATHeader/OATHeader.tsx
miltonzarmada: We are checking if `type === 'application/json'` on the previous method `onFilesUpload`

jamescarpinter: Hmmm. Is that driven by a `.json` extension? Or does that do a deep inspection of the file?

src/Models/Services/Utils.ts
miltonzarmada: I believe `JSONEditor` file also uses `parseModel`

jamescarpinter: Hmmm. What happens if you use the JSONEditor with a model that has an extends/component? Will it throw an error given the references won't resolve?

miltonzarmada: In the case of only having Model0, when updating it via JSONEditor, by adding an extends
` "extends": [
    "dtmi:com:example:model2;1"
  ]`
  
We get the error `No getModels provided to resolve requisite reference(s): dtmi:com:example:model2;1`

jamescarpinter: Ah. So I think we probably need to update JSONEditor to use `parseModels` too (i.e. pass in [...models, ...freshlyEditedModelFromJsonEditor]`).

src/Components/OATHeader/OATHeader.tsx
miltonzarmada: True, I added a validation

src/Components/OATHeader/OATHeader.tsx
```diff
@@ -315,35 +312,78 @@ const OATHeader = ({ dispatch, state }: OATHeaderProps) => {
...
                 const content = await current.text();
-                const error = await parseModel(content);
-                if (!error) {
-                    modelsMetadataReference = populateMetadata(
-                        current,
-                        content,
-                        modelsMetadataReference
+                const validJson = isJsonValid(content);
+                if (validJson) {
+                    const jsonContent = JSON.parse(content);
```
jamescarpinter, line 333: Can we avoid parsing the JSON twice? I.e. we parse it once in the `isJsonValid` function and then again here.

Could we do something like a `safeJsonParse` function that just returns null if it's invalid? And if it's null, then add the error message?

jamescarpinter (approved):

Issue #524, opened 2022-06-03
Cannot navigate to 'Learn more'


rploeg:
**Describe the bug**
Cannot click on 'Learn more' link in the 'New behavior' window under Property expression i

**To Reproduce**
1. Go to New behavior
2. Click on status tab
3. Click on Property expression i (information box)
4. I see a link to Learn more, but when I try to move to the link the pop-up gets away

**Expected behavior**
I can click the link

**Screenshots**
![image](https://user-images.githubusercontent.com/49752333/171806061-1a756d31-898e-49b4-a86c-5a20870c9997.png)

**Browser (please complete the following information):**
- Browser; Edge 102.0.1245.30 64 bits
- Windows 11

PR #660, opened 2022-08-02
tooltip hover fix

BrennaCourtney:
### Summary of changes 🔍 
See original issue here: https://github.com/microsoft/iot-cardboard-js/issues/524

image

The TooltipCallout component could not be interacted with. This PR addresses the bug by replacing the component's Fluent Callout with Fluent Tooltip. 

As a result, there are a few cosmetic differences (such as a change in animation), but otherwise the functionality has improved/remained the same.

### Testing 🧪
Ensure that hovering/clicking behavior now works in the TooltipCallout stories.

See the fix in the original error site by going to BehaviorForm -> Edit Status Tab -> hover over Property expression "i" and ensure that you are able to click on "Learn more."

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/TooltipCallout/TooltipCallout.tsx
+++ b/src/Components/TooltipCallout/TooltipCallout.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React from 'react';
 import {
     ITooltipCalloutProps,
     ITooltipCalloutStyleProps,
@@ -10,10 +10,11 @@ import {
     classNamesFunction,
     useTheme,
     styled,
-    Callout,
     DirectionalHint,
     IconButton,
-    Link
+    Link,
+    TooltipHost,
+    TooltipDelay
 } from '@fluentui/react';
 import { useTranslation } from 'react-i18next';
 
@@ -25,8 +26,6 @@ const getClassNames = classNamesFunction<
 const TooltipCallout: React.FC<ITooltipCalloutProps> = (props) => {
     const { content, calloutProps, dataTestId, styles } = props;
     const { buttonAriaLabel, calloutContent, iconName, link } = content;
-    // state
-    const [flyoutVisible, setFlyoutVisible] = useState(false);
 
     // hooks
     const id = useId();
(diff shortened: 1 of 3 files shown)


src/Components/TooltipCallout/TooltipCallout.tsx
```diff
@@ -38,35 +37,33 @@ const TooltipCallout: React.FC<ITooltipCalloutProps> = (props) => {
...
-                    {calloutContent}
-                    {link && ' '}
-                    {link && (
-                        <Link target="_blank" href={link.url}>
-                            {link.text || t('learnMore')}
-                        </Link>
-                    )}
-                </Callout>
-            )}
+        <div>
```
msnyder-msft, line 40: do we need the wrapping div?

src/Components/TooltipCallout/TooltipCallout.tsx
```diff
@@ -38,35 +37,33 @@ const TooltipCallout: React.FC<ITooltipCalloutProps> = (props) => {
...
+            <TooltipHost
+                directionalHint={DirectionalHint.rightCenter}
+                {...calloutProps}
+                delay={TooltipDelay.zero}
+                styles={classNames.subComponentStyles.callout}
+                content={
+                    <div style={{ fontSize: '14px' }}>
```
msnyder-msft, line 47: do we need the wrapping div? could we set the style on the tooltip styles object instead?

msnyder-msft (approved):
Simple, effective. Love it. Seems good to me.

src/Components/TooltipCallout/TooltipCallout.tsx
BrennaCourtney: div here is needed b/c content prop is of type string | JSX Element -- I also tried moving the font size to the tooltip styles object but no dice

src/Components/TooltipCallout/TooltipCallout.tsx
BrennaCourtney: nope, good point -- done

iremgokce (approved):
fluent tooltiphost <3

PR #662, opened 2022-08-02
Zarmada/working/ADTS-278-

miltonzarmada:
### Summary of changes 🔍


src/Components/OATHeader/internal/FileSubMenu.tsx
```diff
@@ -205,7 +206,11 @@ export const FileSubMenu = ({
                 setIsFileStored(true);
             }
         }
-    }, []);
+    };
+
+    useEffect(() => {
+        onProjectChange();
+    }, [projectName, isActive]);
```
jamescarpinter, line 213: Why is `isActive` a dependency here? It doesn't appear to be used in the `onProjectChange` function?

jamescarpinter (changes requested):

jamescarpinter (approved):

PR #665, opened 2022-08-03
Add ability to view dynamic refresh frequency for a scene

msnyder-msft:
### Summary of changes 🔍 
Adding a new schema entry to configure refresh frequency for a scene. 
[Product Backlog Item 14861095](https://msazure.visualstudio.com/One/_workitems/edit/14861095): Domain Experts can configure the refresh rate

This is done by adding an optional `PollingConfiguration` to a scene.
image
image
image

Users will be able to configure the type of refresh (in a future PR). For now it's just direct in the config file. They can pick between `Realtime` and `Limited`. If they selected `Limited` they will provide a minimum duration between polls. The idea being that they can slow down the refresh rate if they don't need updates as often.

We also exposed a "refresh" action in the Viewer. Clicking it will force an immediate refresh
image

Hovering the button shows details about the polling configuration
image

[Figma Designs](https://www.figma.com/file/LsmZeVOZJgwZs75HPTQmgU/3D-Scenes-Mock-Screens?node-id=46172%3A192479)

### Testing 🧪
- [x] Configure a scene with `Limited` polling configuration (in the config file) and see the limited polling interval respected
- [x] Configure a scene with `Realtime` polling configuration (in the config file) and see the polling interval adjust higher as you get past 10 twins in a scene

### Checklist ✔️
- [x] Added relevant labels


src/Models/Hooks/useRuntimeSceneData.ts
```diff
@@ -10,33 +10,47 @@ import ViewerConfigUtility from '../Classes/ViewerConfigUtility';
...
     const sceneData = useAdapter({
-        adapterMethod: () => adapter.getSceneData(sceneId, scenesConfig),
-        refetchDependencies: [sceneId, scenesConfig],
+        adapterMethod: (args?: { isManualRefresh: boolean }) => {
+            setLastRefreshTime(Date.now());
+            return adapter.getSceneData(
+                sceneId,
+                scenesConfig,
+                selectedLayerIds,
+                args?.isManualRefresh ?? false
```
pasanchMSFT, line 50: Instead of this null check why not make `getSceneData` default to false in case it doesn't receive this parameter or if it receives a null/undefined one as well.

pasanchMSFT (commented):
In general looks good, just wanted to understand some decisions you made before approving.

src/Adapters/__mockData__/3DScenesConfiguration.jsonc
```diff
@@ -66,18 +70,21 @@
                         "url": "https://cardboardresources.blob.core.windows.net/cardboard-mock-files/TruckBoxesEnginesPastmachineReduced.gltf",
                         "extensionProperties": {}
                     }
-                ]
+                ],
+                /* the polling configuration will fetch data as quickly as allowed given the number of twins on the screen */
+                "pollingConfiguration": {
+                    "pollingStrategy": "Realtime",
+                    "minimumPollingFrequency": 0
```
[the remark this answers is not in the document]
msnyder-msft: It will only read this value when the strategy is "Limited" so we shouldn't see any impact of this being 0.

src/Adapters/MockAdapter.ts
```diff
@@ -567,90 +572,88 @@ export default class MockAdapter
...
+                            DatasourceType.ElementTwinToObjectMappingDataSource
+                        ) {
+                            dataSource.elementIDs.forEach((mappingId) => {
+                                mappingIds.push(mappingId);
+                            });
+                        }
 
-                                // TODO get FilteredTwinDatasources
-                            }
+                        // TODO get FilteredTwinDatasources
```
iremgokce, line 598: remove comment if already implemented?

src/Adapters/MockAdapter.ts
```diff
@@ -567,90 +572,88 @@ export default class MockAdapter
...
-                                if (element) {
-                                    // get primary twin
-                                    twins[
-                                        PRIMARY_TWIN_NAME
-                                    ] = this.mockTwins.find(
-                                        (t) => t.$dtId === element.primaryTwinID
+                    // cycle through mapping ids to get twins for behavior and scene
+                    for (const id of mappingIds) {
+                        const twins = {};
+                        const element: ITwinToObjectMapping = scene.elements.find(
```
iremgokce, line 604: if there is no elements in scene it is always an empty array, right? Otherwise:
```suggestion
                        const element: ITwinToObjectMapping = scene.elements?.find(
```

src/Adapters/__mockData__/3DScenesConfiguration.json
```diff
@@ -199,7 +201,11 @@
                         "url": "https://cardboardresources.blob.core.windows.net/cardboard-mock-files/refinery_scene_textured_1m_2.glb",
                         "extensionProperties": {}
                     }
-                ]
+                ],
+                "pollingConfiguration": {
+                    "pollingStrategy": "Limited",
```
iremgokce, line 206: are those ()e.g. "Limited" reserved strings? If so, can we define a enum type in the 3d config schema and use it here?

src/Adapters/__mockData__/3DScenesConfiguration.jsonc
```diff
@@ -66,18 +70,21 @@
                         "url": "https://cardboardresources.blob.core.windows.net/cardboard-mock-files/TruckBoxesEnginesPastmachineReduced.gltf",
                         "extensionProperties": {}
                     }
-                ]
+                ],
+                /* the polling configuration will fetch data as quickly as allowed given the number of twins on the screen */
+                "pollingConfiguration": {
+                    "pollingStrategy": "Realtime",
```
iremgokce, line 76: Can we define an enum type in the config schema so that the polling strategy is no longer a type string?

src/Models/Hooks/useRuntimeSceneData.ts
```diff
@@ -166,11 +184,44 @@ export const useRuntimeSceneData = (
...
+                twinCount: number,
+                pollingConfig: IPollingConfiguration
+            ) => {
+                const MIN_INTERVAL = 10000;
+                const fastestPossibleRefreshRateSeconds = Math.max(
+                    twinCount * 1000, // 1s per twin
+                    MIN_INTERVAL
+                );
+                const actualRefreshRateSeconds =
+                    pollingConfig.pollingStrategy === 'Limited' &&
```
iremgokce, line 203: ahh you already defined a type for it, lets use it
```suggestion
                    pollingConfig.pollingStrategy === IPollingStrategy.Limited &&
```

src/Adapters/__mockData__/3DScenesConfiguration.json
iremgokce: ahh I just noticed that, in fact there is a type for it, please ignore my comment

src/Components/ADT3DViewer/ADT3DViewer.tsx
```diff
@@ -725,12 +726,18 @@ const ADT3DViewerBase: React.FC<IADT3DViewerProps> = ({
...
+                {/* Header controls */}
                 <Stack
                     horizontal
                     styles={classNames.subComponentStyles.headerStack}
                     tokens={{ childrenGap: 8 }}
                 >
+                    <SceneRefreshButton
+                        isRefreshing={isLoading}
+                        lastRefreshTimeInMs={lastRefreshTime}
+                        refreshFrequency={nextRefreshTime - lastRefreshTime}
```
iremgokce, line 738: shouldn't this "refresh frequency" a static number? can't we do this calculation internally?

src/Components/SceneRefreshButton/SceneRefreshButton.tsx
```diff
@@ -0,0 +1,151 @@
...
+            setTimeout(() => {
+                setIsRefreshInProgress(false);
+            }, ANIMATION_DURATION_SECONDS * 1000 + 0.1); // give it enough time to finish the animation, then remove the style
+        }
+    }, [isRefreshing]);
+
+    // to get live updating we have to trigger renders and recalculate on a regular cadence so set a timer to keep checking
+    setTimeout(() => {
+        const timeSince =
+            lastRefreshTimeInMs > 0 ? Date.now() - lastRefreshTimeInMs : 0;
```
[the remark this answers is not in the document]
iremgokce: lastRefreshTimeInMs is a number type

src/Models/Classes/ViewerConfigUtility.ts
```diff
@@ -73,6 +77,25 @@ abstract class ViewerConfigUtility {
         return updatedConfig;
     }
 
+    /** Scene polling configuration */
+    static getPollingConfig(
+        config: I3DScenesConfig,
+        sceneId: string
+    ): IPollingConfiguration {
+        const defaultConfig: IPollingConfiguration = {
+            pollingStrategy: 'Realtime',
```
iremgokce, line 86:
```suggestion
            pollingStrategy: IPollingStrategy.Realtime,
```

src/Models/Classes/ViewerConfigUtility.ts
```diff
@@ -637,19 +660,17 @@ abstract class ViewerConfigUtility {
         sceneId: string
     ) {
         if (!config) return [];
+        const selectedLayers = deepCopy(selectedLayerIds);
```
iremgokce, line 663: I think `selectedLayerIds` was better for naming

src/Models/Classes/ViewerConfigUtility.ts
```diff
@@ -663,7 +684,7 @@ abstract class ViewerConfigUtility {
 
         // Add behavior Ids from selected scene layers to Id dict
         config?.configuration.layers.forEach((layer) => {
-            if (selectedLayerIds.includes(layer.id)) {
+            if (selectedLayers.includes(layer.id)) {
                 layer.behaviorIDs.forEach((behaviorId) => {
                     uniqueBehaviorIds.set(behaviorId, '');
```
iremgokce, line 689: unrelated to this PR but why do we set the value as an empty string here in the map? should this be a Map then? or should we set it to null or something?

src/Models/Hooks/useRuntimeSceneData.ts
```diff
@@ -10,33 +10,47 @@ import ViewerConfigUtility from '../Classes/ViewerConfigUtility';
...
 export const useRuntimeSceneData = (
     adapter: IADT3DViewerAdapter,
     sceneId: string,
     scenesConfig: I3DScenesConfig,
-    pollingInterval: number,
     /** Optional array of layer Ids to apply SceneVisual behavior filtering */
     selectedLayerIds: string[] = null
 ) => {
     const [modelUrl, setModelUrl] = useState('');
+    const [pollingInterval, setPollingInterval] = useState(10);
```
iremgokce, line 36: so it is no longer a prop, should we move the value 10 to constants file?

src/Models/Constants/Types.ts
```diff
@@ -171,6 +171,15 @@ export type CardboardIconNames =
     | 'SpeedHigh'
     | 'View';
 
+export enum DurationUnits {
```
iremgokce, line 174: I think you set the values for enums (0,1,2,3,4,5) to make comparison in another file (`formatTimeInRelevantUnits` method), but still we don't need to set the values for enum keys as you can still make < > comparisons

src/Models/Constants/Types.ts
```diff
@@ -171,6 +171,15 @@ export type CardboardIconNames =
...
+    milliseconds = 0,
+    seconds = 1,
+    minutes = 2,
+    hours = 3,
+    days = 4,
+    years = 5
```
iremgokce, line 180:
```suggestion
    milliseconds,
    seconds,
    minutes,
    hours,
    days,
    years
```

src/Components/SceneRefreshButton/SceneRefreshButton.types.ts
```diff
@@ -0,0 +1,43 @@
...
+    IStyle,
+    IStyleFunctionOrObject,
+    ITheme
+} from '@fluentui/react';
+import { IHeaderControlButtonStyles } from '../HeaderControlButton/HeaderControlButton.types';
+import { IHeaderControlGroupStyles } from '../HeaderControlGroup/HeaderControlGroup.types';
+
+export interface ISceneRefreshButtonProps {
+    isRefreshing: boolean;
+    lastRefreshTimeInMs: number;
```
iremgokce, line 12: nit: it is bothering me seeing millisecond with Ms abbreviation :p maybe using 'ms' or 'msec' by ignoring the camel casing?

src/Components/SceneRefreshButton/SceneRefreshButton.tsx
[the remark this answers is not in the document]
pasanchMSFT: Date.now() is also a number, the milliseconds since Jan 1, 1970. If `lastRefreshTimeInMs` is calculated with Date it will never be 0.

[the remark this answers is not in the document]
msnyder-msft: it's just a safety net if it doesn't get the prop. It won't ever actually execute. I probably added it when my story blew up up for this control because I wasn't passing the prop correctly.

src/Models/Constants/Types.ts
iremgokce: no downside, just the values being a number for "DurationUnits" might be confusing

src/Components/SceneRefreshButton/SceneRefreshButton.types.ts
iremgokce: yeah, it is fine then

src/Components/ADT3DViewer/ADT3DViewer.tsx
iremgokce: no need then

iremgokce (approved):

pasanchMSFT (approved):
LGTM

PR #674, opened 2022-08-10
Various badge fixes

ptallettms:
The problems occured in the Space Station demo which has 20px padding round the viewer. This was causing the popover to be displayed in the wrong position which meant you could move off the badge without dismissing the popover (mouseout).

Also the flag to show the popover is sometimes set, even when no popover is showing, so setting it again did nothing.

The text cursor was showing when over the icon as that is a text element => set to no pointer events.

### Testing 🧪
 > To test this, modify one of the stories to add a 20px padding => check popovers are positioned correctly
> To check that the popover always appears, mouse into the badge very slowly from the bottom right corner => popover should still appear
> The popover will dismiss correctly as long as it is positioned correctly (mouseout)

### Checklist ✔️
- [ ] Added relevant labels


src/Components/AlertBadge/AlertBadge.tsx
```diff
@@ -29,12 +29,14 @@ const AlertBadge: React.FC<IAlertBadgeProps> = ({
                     ? classNames.groupContainer
                     : classNames.singleContainer
             }
-            onMouseOver={(element) => {
+            onMouseMove={(element) => {
```
msnyder-msft, line 32: won't onMouseMove fire super frequently? Do we really want it to keep firing for every movement even when we are just hovering the same badge?

msnyder-msft: can we use `onMouseEnter` instead? We have had better luck with that in other places since it only fires a single time on entry.

msnyder-msft:
Can you add a screenshot for the padding issue? I'm not sure I fully understand that one.

msnyder-msft (commented):
left a couple of comments

src/Components/AlertBadge/AlertBadge.tsx
ptallettms: No, mouseMove is fine because the popover is over the top of it.

src/Components/ADT3DViewer/ADT3DViewer.tsx
```diff
@@ -623,15 +623,18 @@ const ADT3DViewerBase: React.FC<IADT3DViewerProps> = ({
...
-            setAlertPanelItems({
-                element: badgeGroup.element,
-                behaviors: badgeGroup.behaviors,
-                twins: badgeGroup.twins
-            });
-            // Adding offsets to ensure the popover covers the alerts badges as per the designs
-            setAlertPopoverPosition({ left: left - 50, top: top - 30 });
             setIsAlertPopoverVisible(true);
+        } else {
+            setIsAlertPopoverVisible(false); // Hide old popover
```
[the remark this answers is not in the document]
ptallettms: It doesn;t really matter what order they are done in. The point of these two lines is to force a refresh.

ptallettms:
image

src/Components/AlertBadge/AlertBadge.tsx
ptallettms: We could possibly revert this change as I'm pretty sure its the true/false flag that was causing the issue

src/Components/AlertBadge/AlertBadge.tsx
```diff
@@ -29,12 +29,14 @@ const AlertBadge: React.FC<IAlertBadgeProps> = ({
...
+                        element.target['parentElement'];
                     onBadgeGroupHover(
                         badgeGroup,
-                        element.clientX,
-                        element.clientY
+                        container?.offsetLeft || 0,
+                        container?.offsetTop || 0
                     );
                 }
             }}
```
ptallettms, line 42: clientLeft/top are incorrect if there's padding on the viewer - use offsetLeft/top of the container

src/Components/AlertBadge/AlertBadge.tsx
ptallettms: reverted

src/Components/ADT3DViewer/ADT3DViewer.tsx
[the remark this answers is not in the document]
ptallettms: I was trying to avoid root causing this, but decided to do that - the set visible should be unconditional and then the issue does not arise. No need for false/true.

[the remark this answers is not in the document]
ptallettms: The diffs are a bit hard to follow. The code is now:

    const onBadgeGroupHover = (
        badgeGroup: SceneViewBadgeGroup,
        left: number,
        top: number
    ) => {
        setAlertPanelItems({
            element: badgeGroup.element,
            behaviors: badgeGroup.behaviors,
            twins: badgeGroup.twins
        });
        // Adding offsets to ensure the popover covers the alerts badges as per the designs
        setAlertPopoverPosition({ left: left - 50, top: top - 30 });
        setIsAlertPopoverVisible(true);
    };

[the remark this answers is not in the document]
ptallettms: The issue was caused if the adapter polled whilst the popover was visible - some stale state going on

src/Components/OATHeader/internal/FormOpen.types.ts
```diff
@@ -0,0 +1 @@
+//
```
msnyder-msft, line 1: can we just delete this file instead please?

msnyder-msft (approved):

pasanchMSFT (approved):

PR #679, opened 2022-08-15
feat: enable dynamic cache length for resources

msnyder-msft:
### Summary of changes 🔍 
Allow the cache length for an entity to be updated dynamically. In particular, allow this when the user is selecting a refresh rate faster than the default cache length of 10 seconds.

### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Models/Classes/AdapterEntityCache.ts
+++ b/src/Models/Classes/AdapterEntityCache.ts
@@ -18,6 +18,11 @@ class AdapterEntityCache<T extends IAdapterData> {
         this.maxAgeMs = maxAgeMs;
     }
 
+    setMaxAgeMs(maxAge: number): void {
+        this.maxAgeMs = maxAge;
+        this.cachedEntities = new Map();
+    }
+
     /**
      * Retrieves an entity from the cache.  If the entity is stale, getEntityData will be executed to resolve the entity
      * @param key The key for the desired entity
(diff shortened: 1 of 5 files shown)


src/Models/Hooks/useRuntimeSceneData.ts
```diff
@@ -58,6 +58,12 @@ export const useRuntimeSceneData = (
         pollingIntervalMillis: pollingInterval
     });
 
+    // notify the adapter of the updating polling interval so the cache length can be adjusted
+    useEffect(() => {
+        // use a value slightly lower than the interval so that there is not a race between the polling and the cache expiring
+        adapter.setTwinCacheLength(pollingInterval - 0.2 * pollingInterval);
```
iremgokce, line 64: why this number -  0.2? can we keep it in a const?

src/Models/Classes/AdapterEntityCache.ts
```diff
@@ -18,6 +18,11 @@ class AdapterEntityCache<T extends IAdapterData> {
         this.maxAgeMs = maxAgeMs;
     }
 
+    setMaxAgeMs(maxAge: number): void {
+        this.maxAgeMs = maxAge;
+        this.cachedEntities = new Map();
```
iremgokce, line 23: so does this mean whenever we change the polling interval we will lose the cachedEntities, this is correct, right?

iremgokce (approved):
left couple of small comments, lgtm

PR #680, opened 2022-08-16
Add metrics telemetry event and rename existing types

msnyder-msft:
### Summary of changes 🔍 
Adds a telemetry event type for metrics and adds some logging to see telemetry even when running locally.

[Product Backlog Item 14663266](https://msazure.visualstudio.com/One/_workitems/edit/14663266): How do we do feature level app insights telemetry logging?

### Testing 🧪
- [x] Enable logging in the `TelemetryServices.ts` file and then take an action on the scene list like create, edit, delete. Look in the console to see the events firing.

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/SceneList/SceneList.tsx
+++ b/src/Components/SceneList/SceneList.tsx
@@ -331,6 +331,9 @@ const SceneList: React.FC<SceneListProps> = ({
                         <ActionButton
                             iconProps={{ iconName: 'Add' }}
                             onClick={() => {
+                                useTelemetry().sendEventTelemetry({
+                                    name: 'Create scene - open'
+                                });
                                 setIsSceneDialogOpen(true);
                             }}
                             disabled={
@@ -442,7 +445,7 @@ const SceneList: React.FC<SceneListProps> = ({
                     sceneToEdit={selectedScene}
                     onEditScene={(updatedScene) => {
                         useTelemetry().sendEventTelemetry({
-                            name: 'Edit scene confirm'
+                            name: 'Edit scene - confirm'
                         });
                         editScene.callAdapter({
                             config: config,
@@ -452,7 +455,7 @@ const SceneList: React.FC<SceneListProps> = ({
                     }}
                     onAddScene={(newScene) => {
                         useTelemetry().sendEventTelemetry({
-                            name: 'Create scene confirm'
+                            name: 'Create scene - confirm'
                         });
                         let newId = createGUID();
                         const existingIds = sceneList.map((s) => s.id);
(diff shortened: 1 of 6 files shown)


src/Components/SceneList/SceneList.tsx
```diff
@@ -207,6 +208,9 @@ const SceneList: React.FC<SceneListProps> = ({
                                 event.stopPropagation();
                                 setSelectedScene(item);
                                 setIsSceneDialogOpen(true);
+                                useTelemetry().sendEventTelemetry({
```
iremgokce, line 211: instead of calling useTelemetry().sendEventTelemetry multiple times as below can we do as we do for useTranslation?:
`const { sendEventTelemetry } = useTelemetry();` and then calling just sendEventTelemetry?

src/Models/Services/TelemetryService/TelemetryService.ts
```diff
@@ -1,17 +1,26 @@
...
+const logDebugConsole = getDebugLogger('TelemetryService', debugLogging);
 
 class TelemetryService {
-    static telemetryCallback;
+    static telemetryCallback: (telemetry: ITelemetryEvent) => void;
 
     // Attach telemetry processing callback from consuming application
     static registerTelemetryCallback(
-        telemetryCallback: (telemetry: Telemetry) => void
+        telemetryCallback: (telemetry: ITelemetryEvent) => void
```
iremgokce, line 12: Since there is a specific type 'event' telemetry down in `TelemetryType`, it is confusing to see ITelemetryEvent here, can we change it to `ITelemetry` or something else?

src/Models/Services/TelemetryService/Telemetry.ts
```diff
@@ -2,12 +2,20 @@ import {
...
     IBaseTelemetryParams,
     IExceptionTelemetryParams,
+    IMetricTelemetryParams,
     IRequestTelemetryParams,
     ITraceTelemetryParams,
     SeverityLevel,
     TelemetryType
 } from './TelemetryService.types';
 
+export type ITelemetryEvent =
```
iremgokce, line 12: Yeah, can we rename this? Since there is type EventTelemetry, seeing ITelemetryEvent might be confusing, I know it is the "event" object but maybe we can rename it to `ITelemetryObject` or something?

src/Components/SceneList/SceneList.tsx
msnyder-msft: Sure. Done.

src/Models/Services/TelemetryService/Telemetry.ts
msnyder-msft: Ok. renamed to `ITelemetryItem`. I'd rather avoid ITelemetry as that doesn't really tell you whether it's the service, a single item, arguments, or whatnot. Let me know if you think `ITelemetryObject` is better.

iremgokce (approved):
Looks good to me!

abalogun-dev (approved):

PR #681, opened 2022-08-17
Advanced Search Result Component

abalogun-dev:
### Testing 🧪
 > You can view the work i did in 3 different stories, the AdvancedSearchResultDetailsList as a standalone component is viewable under the SearchResults folder in the AdvanceSearch folder. You can view the experience of the results table under the parent component, in the Base story of  AdvancedSearch. And lastly, you can view PropertyInspectorCallout by itself under the Property Inspector folder. 

https://msazure.visualstudio.com/One/_workitems/edit/15105160/
Chromatic link: https://601c6b2fcd385c002100f14c-fhnkdjnxaa.chromatic.com/?path=/story/components-advancedsearch--base

### Checklist ✔️
- [ ] Added relevant labels


src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.styles.ts
```diff
@@ -2,16 +2,23 @@ import {
...
+    root: `${classPrefix}-root`,
+    headerCorrection: `${classPrefix}-headerCorrection`
 };
 export const getStyles = (
     _props: IAdvancedSearchResultDetailsListStyleProps
 ): IAdvancedSearchResultDetailsListStyles => {
     return {
         root: [classNames.root],
-        subComponentStyles: {}
+        headerCorrection: [
```
msnyder-msft, line 17: nit: let's just call it `header`

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -71,39 +77,46 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
...
+            </h3>
             <DetailsList
                 items={twins}
                 columns={
                     additionalColumns.length > 0
                         ? columns.concat(additionalColumns)
                         : columns
                 }
-                layoutMode={DetailsListLayoutMode.fixedColumns}
+                layoutMode={DetailsListLayoutMode.justified}
```
msnyder-msft, line 111: I havne't used this prop before. What made you change to justified?

src/Components/PropertyInspector/PropertyInspector.scss
```diff
@@ -2,11 +2,15 @@
...
+    padding: 8px 20px;
     width: 100%;
 }
 
 .cb-property-inspector-loading {
     min-height: 200px;
     padding-top: 20px;
 }
+
+.cb-property-inspector-button {
```
msnyder-msft, line 14: FYI: I'm removing this class as part of the fix for the button padding.

src/Components/PropertyInspector/PropertyInspector.scss
```diff
@@ -2,11 +2,15 @@
 
 .cb-property-inspector-container {
     height: 100%;
-    padding-left: 20px;
+    padding: 8px 20px;
```
msnyder-msft, line 5: is this a class only used by your flyout?

src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
```diff
@@ -149,3 +151,13 @@ ModeToggleWithResolvedData.argTypes = {
         defaultValue: '4690c125-aac8-4456-9203-298c93f5fcf0'
     }
 };
+
+export const PropertyInspectorCallout = (
+    _args,
+    { globals: { theme, locale } }
```
msnyder-msft, line 157: please add _ (ex: _theme) to unused vars here or remove them.

src/Components/PropertyInspector/PropertyInspectorCallout.tsx
```diff
@@ -0,0 +1,52 @@
...
+                }}
+                className={'cb-property-inspector-button'}
+                id={buttonId}
+            />
+            {isVisible && (
+                <Callout
+                    calloutMinWidth={200}
+                    calloutMaxWidth={380}
+                    onDismiss={setIsVisible}
+                    directionalHint={11}
```
msnyder-msft, line 38: can we use the `DirectionalHint` enum instead of a number please?

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -71,39 +77,46 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
...
-                    <IconButton
-                        iconProps={{ iconName: 'EntryView' }}
-                        title={t('advancedSearch.inspectProperty')}
-                        ariaLabel={t('advancedSearch.inspectProperty')}
-                        onClick={(event) => {
-                            event.stopPropagation();
-                        }}
-                        className={'cb-scenes-action-button'}
+                    <PropertyInspectorCalloutButton
+                        twinId={`${currentTwin?.$dtId}`}
```
msnyder-msft, line 81: this does not pass the right twinId. It's showing the data as undefined. We should be able to use the `item` passed into this function instead of managing another state variable.
image

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -71,39 +77,46 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
...
                 return <span>{fieldContent}</span>;
         }
     };
+    const selection = new Selection({
+        getKey(item, index?) {
+            return item + index;
+        },
+        onSelectionChanged: () => {
+            setTwin(selection.getSelection());
+            onTwinSelection(currentTwin);
```
msnyder-msft, line 95: this is undefined in your stories which throws a console error when you select a row. Can you please pass the callback to prevent this?

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -71,39 +77,46 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
...
             default:
                 return <span>{fieldContent}</span>;
         }
     };
+    const selection = new Selection({
+        getKey(item, index?) {
+            return item + index;
+        },
+        onSelectionChanged: () => {
+            setTwin(selection.getSelection());
```
msnyder-msft, line 94: What would you think of isntead of storing it to state, just passing the value of `selection.getSelection` directly to the `onTwinSelection` callback. Maybe I'm missing something, but I am not seeing that that state is used for anything else and you can't rely on this "setState" happening before the next line executes (in fact I don't think it ever will since set state triggers a render and it will finish the current render loop first) This is likely the cause of the other bug I mentioned where the inspector does not have the right value.

src/Components/AdvancedSearch/AdvancedSearch.types.ts
```diff
@@ -5,6 +5,7 @@ import {
     IStyleFunctionOrObject,
     ITheme
 } from '@fluentui/react';
+import { IPropertyInspectorAdapter } from '../../Models/Constants';
```
msnyder-msft, line 8: we can remove this import, it's unused.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.stories.tsx
```diff
@@ -24,7 +25,17 @@ type AdvancedSearchResultDetailsListStory = ComponentStory<
 >;
 
 const Template: AdvancedSearchResultDetailsListStory = (args) => {
-    return <AdvancedSearchResultDetailsList {...args} />;
+    const authenticationParameters = useAuthParams();
```
msnyder-msft, line 28: why do we need this for a mock story?

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -26,9 +27,12 @@ const getClassNames = classNamesFunction<
 const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsListProps> = ({
     twins,
     searchedProperties,
-    styles
+    onTwinSelection,
+    styles,
+    adapter
```
msnyder-msft, line 32: nit: let's keep these in alphabetical order please.

msnyder-msft (commented):
Left a few comments, overall, looking good but the property inspector doesn't show any data for the selected twin.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
darsney: is it possible to have a linting rule that formats this on save somehow?

src/Components/PropertyInspector/PropertyInspector.scss
msnyder-msft: This class is used in the normal property inspector. Are we sure this doesn't have any impact where we use it today?

msnyder-msft: Ok. so then let's remove this padding on top then cause it's pushing the inspector down now. 
image

msnyder-msft: we will also need to make sure this still looks good in the ADT Explorer since that's primarily where this control is used.

msnyder-msft (commented):
Looking good. Just a few minor comments.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
abalogun-dev: This occurs when it's fixed width

image

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
abalogun-dev: yeah you're right, the currentTwin state was not needed so yeah selection.getSelection() can be passed directly to onTwinSelection

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.stories.tsx
abalogun-dev: That's my mistake, no need for it. removing...

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.styles.ts
```diff
@@ -7,15 +7,15 @@ import { CardboardClassNamePrefix } from '../../../../Models/Constants';
...
+    listHeaderCorrection: `${classPrefix}-headerCorrection`
 };
 export const getStyles = (
     _props: IAdvancedSearchResultDetailsListStyleProps
 ): IAdvancedSearchResultDetailsListStyles => {
     return {
         root: [classNames.root],
-        headerCorrection: [
-            classNames.headerCorrection,
+        listHeaderCorrection: [
```
msnyder-msft, line 17: nit: can we remove the "correction" part of the name?

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.types.ts
```diff
@@ -11,26 +11,25 @@ import {
...
 
 export interface IAdvancedSearchResultDetailsListProps {
-    twins: IADTTwin[];
-    searchedProperties: string[]; //string of fieldnames, for specific properties that were used in the search
-    // callback function from parent on what to do once the user selects or deselects a twin
-    onTwinSelection?: (IADTTwin) => void; //not meant to be optional, just did that for now so the stories work
-    /**
-     * Call to provide customized styling that will layer on top of the variant rules.
-     */
+    adapter?: IPropertyInspectorAdapter;
```
msnyder-msft, line 14: adapter is required right? Cause the property inspector won't work without it.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.types.ts
```diff
@@ -11,26 +11,25 @@ import {
...
+    /* Callback function from parent on what to do once the user selects or deselects a twin. */
+    onTwinSelection?: (IADTTwin) => void;
```
msnyder-msft, line 16: let's make the callback required as well. This component needs that.

msnyder-msft (approved):
Looking pretty good.

msnyder-msft:
are we supposed to be able to sort the list by clicking the column headers? Can be a different PR but I think that was expected to work. @darsney thoughts?

msnyder-msft:
let's default sort the list by the twinId just to have some kind of consistency in the ordering.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.stories.tsx
```diff
@@ -41,5 +41,6 @@ const Template: AdvancedSearchResultDetailsListStory = (args) => {
 export const Base = Template.bind({}) as AdvancedSearchResultDetailsListStory;
 Base.args = {
     twins: filteredTwins,
-    searchedProperties: cols
+    searchedProperties: cols,
+    onTwinSelection: null
```
msnyder-msft, line 45: maybe have it log to console or something or pop an alert box with the id to see that it's working?

msnyder-msft (approved):

pasanchMSFT (approved):
LGTM

PR #683, opened 2022-08-20
Query builder UI + field validation

pasanchMSFT:
### Summary of changes 🔍 
Initial PR for the UI of the query builder in Advanced Search modal. This new component will allow users to build queries based on properties, operators and values. This can be achieved by adding/removing query rows and updating the inputs in the respective fields. The scope of this PR includes property combo-box, operators changing based on property type and basic value field validation and row to string query creation.

This does not accomplish the following tasks for advanced search:
- Value auto-complete
- Query search and results integration
- SQL display of query

### Testing 🧪
- Open Advanced search base story
- Click on Add/Remove query row buttons
  -   Validate limit of 10 rows
  -   Validate first row not removable
- Select properties in property dropdown to see how operators change
  -   Validate string/numeric/boolean types to have different operators
- Add values in the respective input field to see how validation works:
  -   Validate empty string
  -   Validate no string values in numeric types

### Checklist ✔️
- [ ] Added relevant labels


msnyder-msft (commented):
This is looking really really good. Mostly naming things and nits right now. Structure looks solid. A reducer might have made sense in the query builder since there's a lot of state properties that get updated together, I'll let you decide if it's worth the refactor.
Only other thing is more comments on the hooks and utils. The components have good comments but those other parts could use some more.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
```diff
@@ -56,8 +56,11 @@ export const buildModelledProperties = async ({
 
     try {
         // Get ADT Models
```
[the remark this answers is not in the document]
pasanchMSFT: Where would you create common utils? That was an issue when I thought about doing this. There is currently no common folder iirc.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.tsx
```diff
@@ -355,7 +355,7 @@ const choiceGroupOptions = [
     }
 ];
 
-const getDropdownOptions = (
+export const getDropdownOptions = (
```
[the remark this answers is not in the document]
pasanchMSFT: not using it, will remove the export.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -0,0 +1,247 @@
...
+                                    text: 'Or'
+                                }
+                            ]}
+                            onChange={onChangeCombinator}
+                            styles={classNames.subComponentStyles.andDropdown}
+                        />
+                    )}
+                </div>
+                {/* Property */}
+                <div className={classNames.propertyContainer}>
```
msnyder-msft, line 212: Can we style the dropdown instead of creating a container?

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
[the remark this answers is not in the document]
msnyder-msft: I'd vote to create a folder/file in the hooks folder I think. It seems fine to have hookUtilities or even a file specific to AdtModelUtilities

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
pasanchMSFT: Updated to not use dropdown, now styling is in the component not the container

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
```diff
@@ -59,52 +73,110 @@ export const getStyles = (
...
+        input: (provided) => ({
+            ...provided,
+            color: 'var(--cb-color-text-secondary)',
+            fontSize: '14px'
+        }),
+        menuList: (provided) => ({
+            ...provided,
+            backgroundColor: 'var(--cb-color-bg-canvas)',
+            maxHeight: '300px',
+            maxWidth: isOnlyFirstRow ? 258 : 218,
```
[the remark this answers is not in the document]
msnyder-msft: also, can we even these numbers out? 258 is very strange. Can we get to an even 260?

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -28,37 +29,88 @@ const QueryBuilder: React.FC<IQueryBuilderProps> = (props) => {
...
+
+    const onChangeValue = useCallback((rowId: string, newValue: string) => {
+        // Reset validity on value change
+        if (newValue.length) {
+            validityMap.current.set(rowId, true);
+        } else {
+            validityMap.current.set(rowId, false);
+        }
+        setIsSearchDisabled(!checkIsValidQuery());
+    }, []);
```
msnyder-msft, line 85: if you are going to `useCallback` on these, make sure your dependency lists are updated. I'd suggest just not using `useCallback` here unless you suspect that it will break a memoization lower in the DOM

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -71,67 +123,46 @@ const QueryBuilder: React.FC<IQueryBuilderProps> = (props) => {
...
-            validityMap.current.set(rowId, false);
-        }
-        setIsSearchDisabled(!checkIsValidQuery());
-    };
+    }, [rows]);
 
-    const onSearch = () => {
+    const onSearch = useCallback(() => {
         const query = buildQuery(Array.from(querySnippets.current.values()));
         executeQuery(query);
```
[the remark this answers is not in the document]
msnyder-msft: I think the full modal probably needs to be blocked with a spinner of some kind so users don't interact while the query is running.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -71,67 +123,46 @@ const QueryBuilder: React.FC<IQueryBuilderProps> = (props) => {
...
-            <Stack horizontal={true}>
-                <h4 className={classNames.columnHeader}>And/Or</h4>
-                <h4 className={classNames.firstColumnHeader}>Property</h4>
-                <h4 className={classNames.columnHeader}>Operator</h4>
-                <h4 className={classNames.columnHeader}>Value</h4>
-            </Stack>
-            <Stack style={{ maxHeight: 150, overflow: 'auto' }}>
+            <div className={classNames.headerGrid}>
+                {!(rows.length === 1) && (
+                    <p className={classNames.headerText}>
```
msnyder-msft, line 151: do we need to put the class on every time? Can't we just style the container? I suspect you're overriding spacing with this class which wouldn't be necessary if these were just `span` instead of `p` tags.

src/Resources/Locales/en.json
```diff
@@ -1148,6 +1148,14 @@
         "toggleSelection": "Toggle to Select Twin",
         "selectRow": "Twin Selected",
         "inspectProperty": "Inspect Property",
-        "results": "Results ({{twinCount}})"
+        "results": "Results ({{twinCount}})",
```
msnyder-msft, line 1151: we are trying to keep the key names reflecting where the value is used rather than the actual text it represents. ex: `ResultsListTitle`, `addRowText`, `AndOrColumnHeader`

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.types.ts
```diff
@@ -56,9 +58,11 @@ export interface IQueryBuilderRowProps {
         propertyName: string,
         propertyType: PropertyValueType
     ) => void;
-    onChangeValue: (rowId: string, rowValue: QueryRowType) => void;
+    onChangeValue: (rowId: string, newValue: string) => void;
+    updateSnippet: (rowId: string, rowValue: QueryRowType) => void;
```
msnyder-msft, line 62: naming nit: `onUpdateSnippet`

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -39,61 +46,95 @@ const QueryBuilderRow: React.FC<IQueryBuilderRowProps> = (props) => {
         onChangeValue,
         position,
         removeRow,
-        styles
+        styles,
+        theme,
+        updateSnippet
     } = props;
+    const propertySelectorId = useId('cb-advanced-search-property-select');
```
msnyder-msft, line 53: can this just be a const at module scope?

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -39,61 +46,95 @@ const QueryBuilderRow: React.FC<IQueryBuilderRowProps> = (props) => {
...
-        []
+    /** React select values */
+    const { isLoading, flattenedModelProperties } = useFlattenedModelProperties(
+        {
+            adapter,
+            allowedPropertyValueTypes
+        }
     );
+
+    const [propertyOptions, setPropertyOptions] = useState([]);
```
msnyder-msft, line 68: give this a type please, otherwise it's considered an `any[]` which isn't very safe.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -39,61 +46,95 @@ const QueryBuilderRow: React.FC<IQueryBuilderRowProps> = (props) => {
...
 
+    // Update operator options
     useEffect(() => {
-        const operatorOptions = getOperators(selectedProperty?.data.type);
-        setOperatorOptions(operatorOptions);
-        if (operatorOptions.length) {
-            setSelectedOperator(operatorOptions[0]);
+        if (selectedProperty) {
+            const operatorOptions = getOperators(selectedProperty.data.type);
+            console.log('options', operatorOptions);
```
msnyder-msft: remove the console log

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
```diff
@@ -42,23 +43,24 @@ export enum RESERVED_WORDS {
...
+    } else {
+        return '0';
+    }
+};
 
-interface OperatorFunction {
-    operatorType: 'Function';
-    operatorFunction: (property: string, value: string) => string;
-}
+export const getDefaultOperator = (): OperatorData => {
```
msnyder-msft, line 54: why is this a funciton? Can it be a const? `DEFAULT_OPERATOR = `

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
```diff
@@ -42,23 +43,24 @@ export enum RESERVED_WORDS {
...
+    };
+};
 
-export interface QueryRowType {
-    property: string;
-    operatorData: OperatorData;
-    value: string;
-    combinator: string;
-}
+export const getDefaultCombinator = (): string => {
```
msnyder-msft, line 61: same here, no need for the function call overhead.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -39,61 +46,95 @@ const QueryBuilderRow: React.FC<IQueryBuilderRowProps> = (props) => {
...
-                    comboBoxOptions.push({
-                        key: property.key,
-                        text: property.localPath,
-                        data: {
-                            name: property.name,
-                            type: property.propertyType
-                        }
+        const propertyComboboxOptions = [];
+        if (flattenedModelProperties) {
+            Object.keys(flattenedModelProperties).forEach(
```
msnyder-msft, line 86: suggestion: could pull the creation logic out to a module scope helper that returns the list of options to keep render a bit more condensed

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -110,51 +151,66 @@ const QueryBuilderRow: React.FC<IQueryBuilderRowProps> = (props) => {
...
-            combinator: selectedCombinator,
-            operatorData: selectedOperator.data,
-            property: selectedProperty.text, // MAYBE CHANGE TO DATA VALUE?
-            value: newValue
-        });
+        setSelectedValue(newValue);
+        onChangeValue(rowId, newValue);
     };
 
     const onChangeDropdownValue = (
```
msnyder-msft, line 169: nit: `onChangeValueDropdown` just to match with `onChangeField`

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -110,51 +151,66 @@ const QueryBuilderRow: React.FC<IQueryBuilderRowProps> = (props) => {
...
-            value: option.text
-        });
+        setSelectedValue(option.text);
+        onChangeValue(rowId, option.text);
     };
 
+    const propertySelectorStyles = reactSelectStyles(isRemoveDisabled);
+
+    const Group = (props) => (
+        <div>
```
msnyder-msft, line 181: curious, why do we need the extra div?

src/Models/Hooks/useFlattenedModelProperties.ts
```diff
@@ -24,10 +24,10 @@ const getPropertyModel = (propertyId: string) => {
     return modelName;
 };
 
-const groupFlattenedProperties = (flattenedProperties) => {
+const groupFlattenedModelProperties = (flattenedModelProperties) => {
```
msnyder-msft, line 27: can we give the argument a type please? it's an `any` right now.

src/Models/Hooks/useFlattenedModelProperties.ts
```diff
@@ -83,25 +83,29 @@ const buildFlattenedProperties = async (
...
         const updateModelProperties = async () => {
-            setIsLoading(true);
-            const newFlattenedProperties = await buildFlattenedProperties(
+            const newFlattenedModelProperties = await buildFlattenedModelProperties(
                 adapter,
                 allowedPropertyValueTypes
             );
             if (isMounted) {
-                setFlattenedProperties(newFlattenedProperties);
+                console.log('IS MOUNTED');
```
msnyder-msft: extra console logs.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
```diff
@@ -0,0 +1,182 @@
...
+            minHeight: '32px',
+            outline: '1px solid',
+            outlineColor: state.isFocused
+                ? 'var(--cb-color-theme-primary)'
+                : 'var(--cb-color-input-border)'
+        }),
+        indicatorSeparator: () => ({
+            display: 'none'
+        }),
+        dropdownIndicator: (provided) => ({
```
msnyder-msft, line 138: I feel like list item text shouldn't be bold. It feels pretty heavy.
image

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
```diff
@@ -0,0 +1,182 @@
...
+            borderRadius: '2px',
+            fontSize: '14px',
+            fontWeight: 400,
+            minHeight: '32px',
+            outline: '1px solid',
+            outlineColor: state.isFocused
+                ? 'var(--cb-color-theme-primary)'
+                : 'var(--cb-color-input-border)'
+        }),
+        indicatorSeparator: () => ({
```
msnyder-msft, line 135: the Add new row button needs it's text updated and it's styling fixed to match the designs. It feels off to be floating in from the side like that right above another button which is NOT indented.
Actual
image
Designs 
image

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
```diff
@@ -0,0 +1,182 @@
...
+            },
+            props.rowCount === 1
+                ? {
+                      gridTemplateColumns: '4fr 4fr 4fr 1fr'
+                  }
+                : {
+                      gridTemplateColumns: '2fr 4fr 4fr 4fr 1fr'
+                  }
+        ],
+        headerText: [
```
msnyder-msft, line 40: The header text should not be bright white. Let's make it match the designs please. It should be more of a softer gray

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
```diff
@@ -0,0 +1,182 @@
...
+                : 'var(--cb-color-input-border)'
+        }),
+        indicatorSeparator: () => ({
+            display: 'none'
+        }),
+        dropdownIndicator: (provided) => ({
+            ...provided,
+            padding: '0px 4px'
+        }),
+        input: (provided) => ({
```
msnyder-msft, line 142: the text color for selected value seems a bit hot, can we align this with the semantic colors? I think it should be a gray instead of white.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -0,0 +1,310 @@
...
+    getDefaultCombinator,
+    getDefaultOperator,
+    getDefaultPropertyValues,
+    getOperators
+} from './QueryBuilderUtils';
+import { useFlattenedModelProperties } from '../../../../Models/Hooks/useFlattenedModelProperties';
+import Select, { components, SelectOptionActionMeta } from 'react-select';
+import { useTranslation } from 'react-i18next';
+import BaseComponent from '../../../BaseComponent/BaseComponent';
+import { PropertyValueType } from '../../../ModelledPropertyBuilder/ModelledPropertyBuilder.types';
```
msnyder-msft, line 32: unused import.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -0,0 +1,310 @@
...
+    );
+
+    const Menu = (props) => (
+        <Callout
+            target={`#${propertySelectorId}`}
+            styles={classNames.subComponentStyles.propertyCallout}
+            isBeakVisible={false}
+        >
+            <BaseComponent theme={theme}>
+                <components.MenuList
```
msnyder-msft, line 193: can we add a tooltip on each menu item hover? long values get cut off and you can't read them. 
image

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -0,0 +1,203 @@
...
+                <p className={classNames.headerText}>
+                    {t('advancedSearch.property')}
+                </p>
+                <p className={classNames.headerText}>
+                    {t('advancedSearch.operator')}
+                </p>
+                <p className={classNames.headerText}>
+                    {t('advancedSearch.value')}
+                </p>
+            </div>
```
msnyder-msft, line 164: the results list should scroll internally but not scroll the column headers NOR the list title.
image

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -0,0 +1,203 @@
...
+            <div className={classNames.rowContainer}>
+                {rows.map((row, index) => (
```
msnyder-msft, line 166: Is it correct that the query does not include the parent property name for WheelInformation as is shown in the dropdown?
image

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
```diff
@@ -0,0 +1,182 @@
...
+                      gridTemplateColumns: '4fr 4fr 4fr 1fr'
+                  }
+                : {
+                      gridTemplateColumns: '2fr 4fr 4fr 4fr 1fr'
+                  }
+        ],
+        headerText: [
+            classNames.headerText,
+            {
+                fontSize: '16px',
```
msnyder-msft, line 43: the designs call for 14px headers, not 16.

src/Components/AdvancedSearch/AdvancedSearch.stories.tsx
```diff
@@ -12,13 +16,29 @@ export default {
     decorators: [getDefaultStoryDecorator<IAdvancedSearchProps>(wrapperStyle)]
```
msnyder-msft, line 16: can we add an interactive test that adds a property of each type so that when we get back Chromatic it will snapshot the different visual styles of each editor?

msnyder-msft (changes requested):
Left a lot of styling comments. Let's try to use semantic colors as much as possible, especially for the react-select control. Let's try to avoid putting margin on every single element, we can leverage `<Stack>` instead which also helps us to easily extend components and get the spacing automatically in the future without having to add more styles.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
pasanchMSFT: react-select doesn't have this built-in I'll include this functionality later if there is feedback that we need it.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
pasanchMSFT: not sure what you mean by this one?

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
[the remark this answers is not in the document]
pasanchMSFT: Good catch, will include that when I add the execute query functionality

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
pasanchMSFT: span tags don't interact well with grid, they ignore it.

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
```diff
@@ -27,11 +27,24 @@ const filteredTwins: IADTTwin[] = twinData;
...
+    } = props;
     const classNames = getClassNames(styles, {
         theme: useTheme()
     });
     const titleId = useId('advanced-search-modal-title');
+    const additionalProperties = useRef(new Set<string>());
+    const executeQuery = (query: string) => {
+        alert(query);
+    };
+    const updateColumns = (properties: Set<string>) => {
```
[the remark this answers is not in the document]
pasanchMSFT: yep, the prop change will cause the child component to re-render

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -0,0 +1,171 @@
...
+    };
+
+    const onChangeValue = (rowId: string, rowValue: QueryRowType) => {
+        if (rowValue.value.length) {
+            validityMap.current.set(rowId, true);
+        } else {
+            validityMap.current.set(rowId, false);
+        }
+        querySnippets.current.set(rowId, rowValue);
+        setIsSearchDisabled(!checkIsValidQuery());
```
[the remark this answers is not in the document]
pasanchMSFT: A reducer seems a bit of an overkill here, but I can try the pattern in the next feature I work on. If I find that the reducer pattern is more light-weight I'll refactor this.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -0,0 +1,171 @@
...
+    };
+
+    const onSearch = () => {
+        const query = buildQuery(Array.from(querySnippets.current.values()));
+        executeQuery(query);
+    };
+
+    return (
+        <div className={classNames.root}>
+            <Stack horizontal={true}>
```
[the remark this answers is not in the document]
pasanchMSFT: I know it is possible to add boolean props with this shorthand, however I like the explicit ={true} and it is not very cumbersome.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
```diff
@@ -0,0 +1,177 @@
...
+    propertyType?: PropertyValueType
+): IDropdownOption<OperatorData>[] => {
+    if (!propertyType) {
+        return [];
+    }
+
+    const operators: IDropdownOption[] = [
+        {
+            key: '1',
+            text: 'Equals',
```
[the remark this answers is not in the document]
pasanchMSFT: I believe we settled on all the inputs to be in English because they are part of the query.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+    return operators;
+};
+
+export const buildQuery = (querySnippets: QueryRowType[]) => {
+    let fullQuery = `SELECT *\nFROM DIGITALTWINS T\n`;
+    querySnippets.forEach((snippet, index) => {
+        if (snippet.operatorData.operatorType === 'Function') {
+            if (index !== 0) {
+                fullQuery = fullQuery.concat(snippet.combinator);
```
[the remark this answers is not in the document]
pasanchMSFT: non-first element has AND/OR

src/Components/AdvancedSearch/Internal/QueryBuilder/useFlattenedProperties.ts
```diff
@@ -0,0 +1,116 @@
...
+            const newFlattenedProperties = await buildFlattenedProperties(
+                adapter,
+                allowedPropertyValueTypes
+            );
+            if (isMounted) {
+                setFlattenedProperties(newFlattenedProperties);
+                setIsLoading(false);
+            }
+        };
+        updateModelProperties();
```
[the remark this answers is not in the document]
pasanchMSFT: It is cleaner to do that with async functions, I've seen the pattern in several places.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
pasanchMSFT: It seems like this is creating a unique id for every row with the string provided as a prefix. I'll leave it like this because it is required for the callout to target the correct react-select dropdown.

src/Components/AdvancedSearch/Internal/QueryBuilder/useFlattenedProperties.ts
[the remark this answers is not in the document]
msnyder-msft: Ok. But why not just make the callback you give to the useEffect async?

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
msnyder-msft: neat. good to know. thanks.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -0,0 +1,171 @@
...
+        // Remove row component
+        const newRows = [...rows];
+        if (index === 0) {
+            newRows.shift();
+        } else {
+            newRows.splice(index, 1);
+        }
+        updateRows(newRows);
+        // Remove row values in query and property
+        querySnippets.current.delete(rowId);
```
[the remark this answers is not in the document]
pasanchMSFT: I'll explore a reducer approach in the next feature, I might refactor this if the pattern feels better. For the moment I'll leave it as is.

src/Components/AdvancedSearch/AdvancedSearch.tsx
```diff
@@ -51,6 +53,10 @@ const AdvancedSearchModal: React.FC<IAdvancedSearchProps> = (props) => {
     const updateColumns = (properties: Set<string>) => {
         additionalProperties.current = properties;
     };
+    const stackTokens: IStackTokens = {
```
msnyder-msft, line 56: nit: can get pulled out to module scope so we don't recreate each render.

src/Components/AdvancedSearch/AdvancedSearch.tsx
```diff
@@ -51,6 +53,10 @@ const AdvancedSearchModal: React.FC<IAdvancedSearchProps> = (props) => {
...
+        childrenGap: '20px',
```
msnyder-msft, line 57: nit: children gap can just be a number too. That's what we do elsewhere in the app. `childrenGap: 20`

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -146,45 +148,51 @@ const QueryBuilder: React.FC<IQueryBuilderProps> = (props) => {
 
     return (
         <div className={classNames.root}>
-            <div className={classNames.headerGrid}>
-                {!(rows.length === 1) && (
+            <Stack tokens={{ childrenGap: 10 }}>
+                <div className={classNames.headerGrid}>
+                    {!(rows.length === 1) && (
```
msnyder-msft, line 153: nit: instead of inverting the full condition can we do `rows.length !== 1`? Just easier to read IMO

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
msnyder-msft: if they are styled by the grid, a `div` would work well then I think no? Then you don't have to override the spacing that comes with `p` tags.

msnyder-msft (approved):
Getting pretty close. Let's close out the last comments and send it.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -0,0 +1,307 @@
...
+                    }
+                    isSearchable={true}
+                    components={{ Group: Group, Menu: Menu }}
+                    onChange={onChangePropertySelected}
+                    styles={propertySelectorStyles}
+                />
+            </div>
+            <div className={classNames.inputColumn}>
+                {/* Operator */}
+                <Dropdown
```
msnyder-msft, line 279: This should be disabled if we haven't selected a property yet.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -0,0 +1,307 @@
...
+                            onChange={onChangeCombinator}
+                        />
+                    )}
+                </div>
+            )}
+            <div className={classNames.inputColumn}>
+                {/* Property */}
+                <Select<PropertyOption>
+                    id={propertySelectorId}
+                    options={propertyOptions}
```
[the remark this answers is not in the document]
msnyder-msft: it also doesn't loc today, so that value likely comes from the library and we should set our own anyways.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -0,0 +1,307 @@
...
+                    components={{ Group: Group, Menu: Menu }}
+                    onChange={onChangePropertySelected}
+                    styles={propertySelectorStyles}
+                />
+            </div>
+            <div className={classNames.inputColumn}>
+                {/* Operator */}
+                <Dropdown
+                    options={operatorOptions}
+                    selectedKey={selectedOperator ? selectedOperator.key : null}
```
msnyder-msft, line 281: this should have placeholder values. `Select an operator` or `Select operator` and if a property is not selected it should say `No property selected`

darsney (approved):
Component feels fire to me, any chance we could get a stories.local for this bad boy to test it against a real instance?  Might be in a subsequent PR though, which is fine, lets just make sure we got one

I defer code comments to folks who code regularly

src/Components/AdvancedSearch/AdvancedSearch.stories.tsx
```diff
@@ -1,24 +1,44 @@
...
 
 export default {
     title: 'Components/AdvancedSearch',
-    component: AdvancedSearchModal,
+    component: AdvancedSearch,
     decorators: [getDefaultStoryDecorator<IAdvancedSearchProps>(wrapperStyle)]
 };
 
-type AdvancedSearchStory = ComponentStory<typeof AdvancedSearchModal>;
+type AdvancedSearchStory = ComponentStory<any>;
```
iremgokce, line 19:
```suggestion
type AdvancedSearchStory = ComponentStory<AdvancedSearch>;
```

src/Components/AdvancedSearch/AdvancedSearch.stories.tsx
```diff
@@ -1,24 +1,44 @@
...
 
-const Template: AdvancedSearchStory = (args) => {
-    return <AdvancedSearchModal {...args} />;
+const Template: AdvancedSearchStory = (
+    args: IAdvancedSearchProps,
+    context: IStoryContext<IAdvancedSearchProps>
+) => {
+    return <AdvancedSearch {...args} theme={args.theme || context.theme} />;
```
iremgokce, line 25: lets use the default decorator which already wraps the component with the context to be consistent in our stories so you can remove theme prop here: 
image

abalogun-dev (approved):

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
```diff
@@ -42,24 +55,35 @@ const AdvancedSearchModal: React.FC<IAdvancedSearchProps> = (props) => {
...
+                    Link the right property by narrowing down your search
+                </p>
             </div>
             <div className={classNames.content}>
                 <div className={classNames.queryContainer}>
-                    <QueryBuilder />
+                    <QueryBuilder
+                        adapter={adapter}
+                        allowedPropertyValueTypes={allowedPropertyValueTypes}
+                        executeQuery={executeQuery}
```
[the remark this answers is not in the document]
pasanchMSFT: I'd rather keep them separate for single responsibility.

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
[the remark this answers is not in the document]
msnyder-msft: I still think we should move this to state. Refs are intended for things that won't be reflected in the UI.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
pasanchMSFT: using them to avoid re-renders of input fields. Should not be a huge performance upgrade but at least it should save us some renders.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
```diff
@@ -46,7 +59,8 @@ export const getStyles = (
             },
             addButton: {
                 root: {
```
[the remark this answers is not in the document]
pasanchMSFT: underline on hover/focus is not done in other action buttons in the app. i.e. behaviors list in element form.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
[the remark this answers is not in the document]
pasanchMSFT: I'm still not convinced on moving these, I re-use ~2-3 functions and don't think there will be other scenarios for this. If this shows up again. I think then would be a good time to create a utils file

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
pasanchMSFT: header text in the designs is white

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
```diff
@@ -0,0 +1,174 @@
...
+                }
+            });
+            break;
+        default:
+    }
+    return operators;
+};
+
+export const buildQuery = (querySnippets: QueryRowData[]) => {
+    let fullQuery = `SELECT *\nFROM DIGITALTWINS T\n`;
```
iremgokce, line 152: should we add the model id into the query as well? Because in the property dropdown it is sectioned by models and if there are same named properties from two different models I would expect from the result to show me only the twins of that model which has that property, is that right?

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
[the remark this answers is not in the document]
msnyder-msft: Then we need a hover/focus state for the button. Right now there is no difference between rest and hover states.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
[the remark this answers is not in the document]
msnyder-msft: Can we link the two using a constant? so compute it here as `DIALOG_WIDTH - 2`

src/Components/AdvancedSearch/AdvancedSearch.stories.tsx
pasanchMSFT: I get type errors when I try to do interactive stories, I'll leave the any since I see other stories doing the same thing as well.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
```diff
@@ -0,0 +1,174 @@
...
+    if (!propertyType) {
+        return [];
+    }
+
+    const operators: IDropdownOption[] = [
+        {
+            key: '1',
+            text: 'Equals',
+            data: {
+                operatorType: 'Simple',
```
iremgokce, line 73: can we define enum for operator type?

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
```diff
@@ -0,0 +1,174 @@
...
+    propertyType?: PropertyValueType
+): IDropdownOption<OperatorData>[] => {
+    if (!propertyType) {
+        return [];
+    }
+
+    const operators: IDropdownOption[] = [
+        {
+            key: '1',
+            text: 'Equals',
```
iremgokce, line 71: can we define enum for these texts as well?

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
```diff
@@ -0,0 +1,174 @@
...
+                    snippet.property,
+                    snippet.value
+                )}\n`
+            );
+        } else {
+            if (index !== 0) {
+                fullQuery = fullQuery.concat(`${snippet.combinator} `);
+            }
+            fullQuery = fullQuery.concat(
+                `WHERE ${snippet.property} ${snippet.operatorData.operatorSymbol} ${snippet.value}\n`
```
iremgokce, line 169: yeah, as we discussed, just wanted to highlight that there are two "WHERE" clause appended but it is fixed in the subsequent PR: https://github.com/microsoft/iot-cardboard-js/pull/688

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
```diff
@@ -58,6 +58,7 @@ export const buildModelledProperties = async ({
...
         const modelDict = (await adapter.getAllAdtModels()).getData()
             .parsedModels;
+        console.log('modelDict', modelDict);
```
iremgokce, line 61: can we use `useLogger` instead for all console.logs around? @msnyder-msft this is the hook that we should use for any type of logging purpose from now on right? Not the utils method you created before this hook `getDebugLogger`?

src/Models/Hooks/useFlattenedModelProperties.ts
```diff
@@ -0,0 +1,117 @@
...
+            const newFlattenedModelProperties = await buildFlattenedModelProperties(
+                adapter,
+                allowedPropertyValueTypes
+            );
+            if (isMounted) {
+                setFlattenedModelProperties(newFlattenedModelProperties);
+                setIsLoading(false);
+            }
+        };
+        updateModelProperties();
```
iremgokce, line 108: so if you just move the updateModelProperties method out of the useEffect and call it in the useEffect with empty dependency list is going to call it onmount anyhow right? I mean no need to define a isMounted in useeffect and check it with if condition in the updateModelProperties method? maybe I am missing something :)

iremgokce:
left couple of small comments, nice job! :)

iremgokce (approved):

src/Components/AdvancedSearch/AdvancedSearch.types.ts
```diff
@@ -17,17 +21,18 @@ export interface IAdvancedSearchProps {
         IAdvancedSearchStyleProps,
         IAdvancedSearchStyles
     >;
+    theme: Theme;
```
iremgokce, line 24: I think we shouldn't need theme here

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -0,0 +1,307 @@
...
+            <BaseComponent theme={theme} disableDefaultStyles={true}>
+                <components.MenuList
+                    {...props}
+                    styles={propertySelectorStyles.menuList}
+                />
+            </BaseComponent>
+        </Callout>
+    );
+
+    const renderValueField = () => {
```
iremgokce, line 198: Can we prepend the type of the property with icons? We did it in the past for `ValueWidgetBuilder` using `DTDLPropertyIconographyMap`
image

image

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
pasanchMSFT: Isn't `p` a better html element semantically speaking?

src/Components/AdvancedSearch/AdvancedSearch.types.ts
pasanchMSFT: I had it to be able to resolve `var( )` css values but now I don't need it. good catch

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
pasanchMSFT: I'll fix this in a following PR, thanks for the reminder.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.model.ts
pasanchMSFT: I didn't need these console.logs thought I had removed them, nice catch

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
pasanchMSFT: I'll look into this on the next PR, that actually queries for data

msnyder-msft (approved):

PR #684, opened 2022-08-22
ResourcePicker cleanup

iremgokce:
### Summary of changes 🔍 
Closing most of the comments on that original [ResourcePicker PR](https://github.com/microsoft/iot-cardboard-js/pull/585). **The ones still open or keep as is on the original PR are:**

- How to keep the ResourcePicker components mounted as the dialog is getting opened/closed: [Comment-1](https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r935832778), [Comment-2](https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r935833292)
- Finding the selected option among options by comparing the text value (it is because for freeform added options has key of the added text whereas the fetched options have the key of the resource id, so it is safer to compare the displayed text to check if exists): [Comment](https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r935848207)
- Updating the options state variable if any of the other selected or additional options state variable changes: [Comment](https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r935852656)
- Naming convention for callback function: [Comment](https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r935859901)
- The structure of how we keep the selected adt environment and container url in the local storage: [Comment](https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r944859439)
- With the selected option prop change in ResourcePicker we should update the internal state variable: [Comment](https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r944865305)
- The expected behavior when clearing the combobox and blur from it: [Comment](https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r944869971)
- The default header for ResourcePicker component which is currently subscription name: [Comment](https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r944873844)

### Testing 🧪
Same as the original PR

### Checklist ✔️
- [x] Added relevant labels


src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -241,12 +247,30 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
         [areValuesEqual, getDisplayFieldValue]
     );
 
+    const sortResources = (r1, r2) => {
```
msnyder-msft, line 250: please give these params a type so we get the type safety.

src/Components/EnvironmentPicker/EnvironmentPicker.test.tsx
```diff
@@ -0,0 +1,216 @@
...
+            };
+            initialState.containerItems = {
+                containers: [MOCK_STORAGE_CONTAINER, MOCK_STORAGE_CONTAINER2],
+                containerToEdit: MOCK_STORAGE_CONTAINER2
+            };
+
+            const action: EnvironmentPickerAction = {
+                type: EnvironmentPickerActionType.RESET_ITEMS_ON_DISMISS,
+                payload: {
+                    selectedEnvironmentUrl: MOCK_ENVIRONMENT_URL,
```
msnyder-msft, line 148: why does a reset operation require the consumer to pass in all the original data? Shouldn't it know the previous state internally and manage that for the consumer? Seems like it's just a massive "Set" action rather than a "Reset"

src/Components/EnvironmentPicker/EnvironmentPickerManager.test.ts
```diff
@@ -0,0 +1,153 @@
...
+        const displayText2 = getContainerDisplayText(
+            MOCK_STORAGE_CONTAINER_NAME,
+            MOCK_STORAGE_ACCOUNT_URL
+        );
+
+        expect(displayText1).toBe('teststorageaccount/teststoragecontainer');
+        expect(displayText2).toBe('teststorageaccount/teststoragecontainer');
+    });
+
+    test('Getting resource url properly', () => {
```
msnyder-msft, line 48: in my ideal world, each test validates a single operation/functionality rather than multiple variations. I find that it helps to self document the scenario since each gets it's own test name and if one condition starts failing you know exactly what was being validated rather than that _something_ about this function has changed. 
What are your thoughts?

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -13,10 +13,15 @@ import {
 export const getEnvironmentDisplayText = (env: string | IAzureResource) => {
     try {
         if (env) {
-            const urlObj = new URL(
-                getResourceUrl(env, AzureResourceTypes.DigitalTwinInstance)
-            );
-            return urlObj.hostname.split('.')[0];
+            if (typeof env === 'string') {
+                if (new URL(env)) {
+                    return env.split('.')[0].split('://')[1]; // to respect casing in the name of the instance
```
msnyder-msft, line 18: I'm curious, why can't we use the object that gets created from the URL class now?

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -13,10 +13,15 @@ import {
 export const getEnvironmentDisplayText = (env: string | IAzureResource) => {
     try {
         if (env) {
-            const urlObj = new URL(
-                getResourceUrl(env, AzureResourceTypes.DigitalTwinInstance)
-            );
-            return urlObj.hostname.split('.')[0];
+            if (typeof env === 'string') {
+                if (new URL(env)) {
```
msnyder-msft, line 17: I know in the past we have seen many exceptions thrown by this URL class. Do we have sufficient safeguards in place now to handle those conditions gracefully?

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -405,7 +405,7 @@ const EnvironmentPicker = ({
         resources: Array<IAzureResource | string>
```
msnyder-msft, line 405: It says the containers are loaded but I only see a single result and I know there are many more in that storage account.
image
image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -405,7 +405,7 @@ const EnvironmentPicker = ({
...
     ) => {
```
msnyder-msft, line 406: Exception thrown if you delete teh value in the STorage account dropdown field. 
1. Select a value
2. Select a container
3. Clear the container
4. Clear the storage account url
5. Boom
image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -468,7 +468,7 @@ const EnvironmentPicker = ({
...
```
msnyder-msft, line 469: Did we give types to the local storage read/write operations?

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -468,7 +468,7 @@ const EnvironmentPicker = ({
...
         dispatch({
```
msnyder-msft, line 470: I had left a comment about combining the local storage values into a single "Environment" object. I don't see that here.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -468,7 +468,7 @@ const EnvironmentPicker = ({
...
-            type: EnvironmentPickerActionType.HANDLE_CONTAINER_CHANGE,
```
msnyder-msft: Have not been able to validate the header in the container list shows the storage account instead of subscription since that dropdown doesn't seem to populate with remote values now. It only shows the single manual entry.
image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -468,7 +468,7 @@ const EnvironmentPicker = ({
...
+            type: EnvironmentPickerActionType.SET_CONTAINER_ITEMS,
             payload: {
```
msnyder-msft, line 472: Still unable to clear values in the instance dropdown.
cant delete

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -107,7 +98,7 @@ const EnvironmentPicker = ({
         Boolean(isDialogHiddenProp)
```
msnyder-msft, line 98: image

msnyder-msft, line 98: image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -152,7 +143,7 @@ const EnvironmentPicker = ({
     useEffect(() => {
```
msnyder-msft, line 143: image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -152,7 +143,7 @@ const EnvironmentPicker = ({
...
         if (!isDialogHidden) {
             dispatch({
-                type: SET_FIRST_TIME_VISIBLE,
```
msnyder-msft, line 155: image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -526,9 +518,7 @@ const EnvironmentPicker = ({
                 <div className="cb-environment-picker-container">
                     <FontIcon iconName={'Database'} />
                     <span className="cb-environment-picker-container-title">
-                        {getContainerText(
-                            state.containerItems.selectedContainer
-                        )}
+                        {containerDisplayText}
                     </span>
                 </div>
             )}
```
msnyder-msft, line 524: you still need to fix this accessibility issue with the dialog being in the DOM when not visible.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -141,6 +146,7 @@ export const getContainerNameFromUrl = (containerUrl: string) => {
 };
```
msnyder-msft, line 146: image

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -141,6 +146,7 @@ export const getContainerNameFromUrl = (containerUrl: string) => {
...
```
msnyder-msft, line 147: image

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -141,6 +146,7 @@ export const getContainerNameFromUrl = (containerUrl: string) => {
...
 
 export const getContainerName = (container: string | IAzureResource) => {
+    // container can either be the name string of the container directly or the azure resource object
     return typeof container === 'string' ? container : container?.name;
 };
```
msnyder-msft, line 151: image

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -32,6 +32,7 @@ import {
     AzureResourceDisplayFields,
     AzureResourceTypes,
     IAzureResource,
+    IComponentError,
```
msnyder-msft, line 35: image

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
```diff
@@ -13,10 +13,15 @@ import {
 export const getEnvironmentDisplayText = (env: string | IAzureResource) => {
```
msnyder-msft, line 13: image

msnyder-msft, line 13: image

msnyder-msft, line 13: image

msnyder-msft, line 13: image

msnyder-msft (changes requested):
There seem to still be a lot of unresolved comments from the previous PR. I tried to copy them all forward but it's hard to find them all. Please ensure we don't miss any from the previous PR.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
iremgokce: URL class is lowercase all the letters in domain name

src/Components/EnvironmentPicker/EnvironmentPicker.test.tsx
iremgokce: what do you mean by "consumer"? The component knows the original selected data through its prop and I just send the value of it to the reducer to update the internal state variables. Yes it is basically "re-setting" the options back to its originals by selected values passed by props. In internal state we only keep the draft values which is subject to change (e.g. including deleting the currently selected freeform option), but when we dismiss the dialog we want to reset the selected options back to the selected values passed thought props. Since there are many things to check and reset it should be a separate reducer action to reset the variables in state.

src/Components/EnvironmentPicker/EnvironmentPickerManager.test.ts
iremgokce: I am not a testing expert but to my mind this test method is testing the `getResourceUrl` method. I just tried to exemplify different inputs for this method to use its parameters as different as possible (e.g. for different resource types like resource url of adt instace/storage account/container). So it is testing a single functionality with different inputs.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
iremgokce: yes, now in the codebase whenever we use URL class it is wrapped in try/catch block and should be handled gracefully.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: yes, this is expected behavior. you should wait until the storage account resource picker one above finishes fetching the data, until then we only show the passed values from props/local storage if exists. This was a requirement from @darsney, originally I was disabling the containers ResourcePicker (3rd combobox) until the storage container finishes fetching the data as I need the resource if of the storage account to fetch containers.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: yes

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: it is outside of the scope of this original PR. We can further discuss how to keep the data in local storage for a separate PR.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: I am not going to change it now as I still think all the resource pickers should be consistent with headers to show the resources divided by subscription. We can change it later if this is what customers want to see.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: As in my other reply this is expected behavior: https://github.com/microsoft/iot-cardboard-js/pull/585#discussion_r944869971

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: had commented and marked as "resolved" this on the original PR

iremgokce: had commented and marked as "resolved" this on the original PR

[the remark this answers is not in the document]
iremgokce: I also commented on the original PR, useReducer expects an initialState, didn't understand what you want me to do here

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: had commented and marked as "resolved" this on the original PR

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: had commented and marked as "resolved" this on the original PR

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
iremgokce: had commented and marked as "resolved" this on the original PR

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
iremgokce: had commented and marked as "resolved" this on the original PR

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
iremgokce: had commented and marked as "resolved" this on the original PR

iremgokce: had commented and marked as "resolved" this on the original PR

iremgokce: had commented and marked as "resolved" this on the original PR (basically reverted name changes)

iremgokce: this is something we can discuss and implement in a separate PR

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: had commented on this on the original PR

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
iremgokce: had commented and marked as "resolved" this on the original PR

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
msnyder-msft: even once it finished loading I never saw a list of containers. Also, it showed a loading spinner for the container picker which implies that it has finished loading the data for this picker. :/ That's pretty confusing.

src/Components/EnvironmentPicker/EnvironmentPickerManager.ts
msnyder-msft: this is the second PR. You already punted it from the PR where you added this. :)

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
msnyder-msft: subscription doesn't tell them anything in this case. It's better to hide the header so that it's not more confusing. This would imply that they might see data from another subscription in some case but that won't happen.

msnyder-msft (changes requested):
I still can't get containers to load when running locally. The storage accounts load, but never get any container names.
image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
msnyder-msft: image

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: fixed: https://github.com/microsoft/iot-cardboard-js/pull/684/commits/8d5b111d93af7c315c728ece23ba4fbff2f310df

msnyder-msft (approved):
Looks like things are working now. Let's send it.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -638,24 +620,36 @@ const EnvironmentPicker = ({
...
                                     }
                                     requiredAccessRoles={{
-                                        enforcedRoleIds: [
-                                            AzureAccessPermissionRoles['Reader']
-                                        ],
+                                        enforcedRoleIds: [],
                                         interchangeableRoleIds: [
-                                            AzureAccessPermissionRoles[
-                                                'Storage Blob Data Owner'
+                                            [
```
pasanchMSFT, line 625: Could you move the arrays out of the props and create constants or refs so the component doesn't create the array on every render?

pasanchMSFT (approved):
LGTM

msnyder-msft (approved):

src/Models/Constants/Types.ts
```diff
@@ -188,9 +188,9 @@ export type IConsoleLogFunction = (
     ...args: unknown[]
 ) => void;
 
-export type MissingAzureRoleDefinitionAssignments = {
-    enforced?: Array<AzureAccessPermissionRoles>;
-    alternated?: Array<AzureAccessPermissionRoles>;
+export type AzureAccessPermissionRoleGroups = {
+    enforced: Array<AzureAccessPermissionRoles>; // roles must exist/required (and)
+    interchangeables: Array<Array<AzureAccessPermissionRoles>>; // within each group one or the other role has to exist (either/or)
```
msnyder-msft, line 193: super nit: would make it singular `interchangable`

src/Models/Constants/Constants.ts
```diff
@@ -439,3 +443,38 @@ export const BlobStorageServiceCorsAllowedHeaders = [
     'x-ms-version',
     'x-ms-blob-type'
 ];
+
+export const EnforcedStorageAccountAccessRoleIds: Array<AzureAccessPermissionRoles> = [];
```
msnyder-msft, line 447: Personally I would have left these where they are used since they don't seem to be used across multiple places.

src/Models/Classes/AdapterDataClasses/AzureManagementData.ts
```diff
@@ -30,9 +30,9 @@ export class AzureResourcesData implements IAdapterData {
 }
 
 export class AzureMissingRoleDefinitionsData implements IAdapterData {
-    data: MissingAzureRoleDefinitionAssignments;
+    data: AzureAccessPermissionRoleGroups;
```
msnyder-msft, line 33: curious why this change was needed. It seems like it's larger than just fixing the type mismatch now that we have the grouped array.

src/Models/Services/Utils.ts
```diff
@@ -688,3 +691,51 @@ export function areResourceValuesEqual(
         return value1 === value2;
     }
 }
+
+export function getRoleIdsFromRoleAssignments(
```
msnyder-msft, line 695: it'd be ideal if we could add some UTs for anythign going into the utils file since it's already standalone and easy to test. Maybe something for another rev.

msnyder-msft (approved):

src/Models/Constants/Types.ts
iremgokce: yeah, I thought about that but decided to keep as plural to make it clear it is array of interchangeable arrays

src/Models/Constants/Constants.ts
iremgokce: Not this particular one, but, I am using other consts that I defined there (`RequiredAccessRoleGroupFor...`) across many files including stories, that is why I moved them here

src/Models/Classes/AdapterDataClasses/AzureManagementData.ts
iremgokce: I renamed it to refer to a "group" including "enforced" and "interchangeables" instead - not to frame it just for "missing" roles, now it is referring just a group of roles I am interested in basically, it can be missing roles or it can refer to a group that I want to check against. Also, I moved the missing role checking related methods to the Utils file to make it consolidated (single place to be consumed by many components/adapters) and kept the internal logic same that I used as a fix last week

src/Models/Services/Utils.ts
iremgokce: okay adding it now for the methods I added, then we can merge

PR #686, opened 2022-08-30
Enable sorting Advanced Search Results List

abalogun-dev:
### Testing 🧪
 > Click on twin id column and see if the items and arrow icon are correct
### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
+++ b/src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
@@ -20,7 +20,7 @@ import {
 import { IADTTwin } from '../../../../Models/Constants';
 import { useTranslation } from 'react-i18next';
 import PropertyInspectorCallout from '../../../PropertyInspector/PropertyInspectorCallout/PropertyInspectorCallout';
-import { sortAlphabetically } from '../../../../Models/Services/Utils';
+import { ascendOrDescendSort } from '../../../../Models/Services/Utils';
 const getClassNames = classNamesFunction<
     IAdvancedSearchResultDetailsListStyleProps,
     IAdvancedSearchResultDetailsListStyles
@@ -38,10 +38,10 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
     const classNames = getClassNames(styles, {
         theme: useTheme()
     });
-    const sortKey = '$dtId';
+    const [sortKey, setSortKey] = useState<keyof IADTTwin>('$dtId');
     const [isSortedDescending, setSortDescending] = useState<boolean>(false);
     const listItems = twins.sort(
-        sortAlphabetically(sortKey, isSortedDescending)
+        ascendOrDescendSort(sortKey, isSortedDescending)
     );
 
     const staticColumns: IColumn[] = [
@@ -56,8 +56,9 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
             isPadded: true,
             onColumnClick: () => {
                 setSortDescending(
-                    isSortedDescending ? !isSortedDescending : true
+                    sortKey === '$dtId' ? !isSortedDescending : false
                 );
+                setSortKey('$dtId');
             }
         },
         {
@@ -76,7 +77,13 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
             fieldName: name,
             minWidth: 100,
             maxWidth: 150,
-            isResizable: true
+            isResizable: true,
+            onColumnClick: () => {
+                setSortDescending(
+                    sortKey === name ? !isSortedDescending : false
+                );
+                setSortKey(name);
+            }
         };
     });
 
@@ -87,7 +94,7 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
     // mark each column based on whether it's currently the one sorted
     columns.forEach((x) => {
         x.isSorted = sortKey === x.fieldName;
-        x.isSortedDescending = isSortedDescending;
+        x.isSortedDescending = x.isSorted ? isSortedDescending : false;
     });
 
     const renderItemColumn: IDetailsListProps['onRenderItemColumn'] = (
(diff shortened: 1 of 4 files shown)


src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -34,11 +35,16 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
 }) => {
     const { t } = useTranslation();
     const twinCount = twins.length;
-
     const classNames = getClassNames(styles, {
         theme: useTheme()
     });
-    const columns: IColumn[] = [
+    const sortKey = '$dtId';
```
msnyder-msft, line 41: the sort key needs to live in state so that it can be updated and triggers renders when changed.

msnyder-msft:
Please add a chromatic interaction test to click the column header so that when chromatic comes back you can see the columns getting sorted.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
abalogun-dev: Ignore this ^ @msnyder-msft . I understand now

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -68,6 +79,17 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
...
     });
+
+    const columns: IColumn[] =
+        additionalColumns.length > 0
+            ? staticColumns.concat(additionalColumns)
+            : staticColumns;
+    // mark each column based on whether it's currently the one sorted
+    columns.forEach((x) => {
+        x.isSorted = sortKey === x.fieldName;
+        x.isSortedDescending = isSortedDescending;
```
msnyder-msft, line 90: we should not set the value to isSortDescending. It should be set to false unless the columnname matches the sort key, then use the state variable. This way when you click on one column, and then another column, it always starts in the same sort order instead of being dependent on the last column's sort direction. That is how most lists work so we should follow that behavior so users have a consistent and expected experience.

src/Models/Services/Utils.ts
```diff
@@ -648,6 +648,35 @@ export function sortAlphabetically<T>(
     };
 }
 
+/**
+ * Sort function to order items from ascending or descending order, for boolean, numbers and strings. Case insensitive sort
+ * NOTE: only works when property is one layer down
+ * @param propertyName name of the property to sort on
+ * @returns Sort function to pass to `.sort()`
+ */
+export function ascendOrDescendSort<T>(
```
msnyder-msft, line 657: can we please not create yet another sort function? We already have a ton of them scattered in the app. This should functionally be the same as the sortAlphabetcially function. if you don't like that name, maybe we just rename it to this name?

abalogun-dev: yeah sortAlphabetically is not the best name anymore since it does more than strings. And i decided to make it a separate function because i was trying to take the less 'intrusive' route.

msnyder-msft: Yeah. I get that. Are you cool with just doing the rename as the last commit now that we are done reviewing and we can just delete the existing function? Otherwise it's just more cleanup later. :)

abalogun-dev: yup definitely. i'll do that now

msnyder-msft (approved):

darsney (approved):
Played with it, felt good.  I think the one thing we will want to be conscious of is the behavior on subsequent searches.  Making sure that sort order is either reset, or adhered to in subsequent searches (it appears like it is adhered to).  The thing is, the column that is used for sorting may also not be present in subsequent renders, let's make sure we handle that gracefully as well

src/Models/Services/Utils.ts
```diff
@@ -627,17 +627,32 @@ export const deleteModel = (id, data, models) => {
...
- * Sort function to order items alphabetically. Case insensitive sort
+ * Sort function to order items from ascending or descending order, for boolean, numbers and strings. Case insensitive sort
  * NOTE: only works when property is one layer down
  * @param propertyName name of the property to sort on
- * @example listItems.sort(sortAlphabetically('textPrimary'))
+ *  @example listItems.sort(sortAscendingOrDescending('textPrimary'))
  * @returns Sort function to pass to `.sort()`
  */
-export function sortAlphabetically<T>(propertyName: keyof T) {
+export function sortAscendingOrDescending<T>(
```
iremgokce, line 636: I think we should keep the old name (sortAlphabetically) but add the type of sort as second param.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -34,11 +35,16 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
 }) => {
     const { t } = useTranslation();
     const twinCount = twins.length;
-
     const classNames = getClassNames(styles, {
         theme: useTheme()
     });
-    const columns: IColumn[] = [
+    const [sortKey, setSortKey] = useState<keyof IADTTwin>('$dtId');
```
iremgokce, line 41: can we move this '$dtId' to a const like `primarySortingKey` or something? it seems it is specialized key it would be nicer to have it in a separate value to easily change it in the future if we need to.

src/Models/Services/Utils.ts
msnyder-msft: The thinking here was that it does more than just strings so the alphabetically name is misleading. I agree that the new name is very long. We could do with `defaultSort` or `sortDescending` and then it takes a prop whether it's ascending/descening.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
msnyder-msft: I'd almost think this deserves to be a const at the app level like we do for PrimaryTwinName.

PR #688, opened 2022-08-31
Querying adapter method (and some UI updates)

pasanchMSFT:
### Summary of changes 🔍 
Created an adapter for querying in ADT Advanced search component. Linked search results with Advanced search results details list as well.

Querying updates from previous PR comments:
- ~~Review if model name needs to be included with same name properties~~ (not in scope)
- Limit of 1000

UI updates to be included in this PR:
- ~~Value text box has a placeholder~~ (will be replaced next PR)
- Iconography for every option
- Divider spacing improvements

### Testing 🧪
- With Mock data:
  -  Open story for Advanced search
  -  Select Car name property
  -  Select Equals operator
  -  Input: 'Model Y' in value field
  -  See results in details list 
- With Real data:
  -  Go to this branch and run the project
  -  Open local story for Advanced search
  -  Query for your Twins

### Checklist ✔️
- [x] Added relevant labels


src/Adapters/ADTAdapter.ts
```diff
@@ -473,6 +475,30 @@ export default class ADTAdapter implements IADTAdapter {
         );
     }
 
+    advancedSearchADTTwins(
```
msnyder-msft, line 478: can we give it a name that agnostic of where it's used? Maybe `searchTwins` or `executeTwinSearch`?

src/Adapters/ADTAdapter.ts
```diff
@@ -473,6 +475,30 @@ export default class ADTAdapter implements IADTAdapter {
...
+        params: AdapterMethodParamsForAdvancedSearchADTwins
+    ) {
+        const adapterMethodSandbox = new AdapterMethodSandbox(this.authService);
+        console.log(params);
```
msnyder-msft, line 482: please remove the console log

src/Adapters/ADTAdapter.ts
```diff
@@ -473,6 +475,30 @@ export default class ADTAdapter implements IADTAdapter {
...
+                url: `${this.adtProxyServerPath}/query`,
+                headers: {
+                    'x-adt-host': this.adtHostUrl
+                },
+                params: {
+                    'api-version': ADT_ApiVersion
+                },
+                data: {
+                    query: params.query,
+                    continuationToken: params.continuationToken
```
msnyder-msft, line 496: are you going to do anything with the continuation token? Or does the caller have to deal with that?

src/Adapters/MockAdapter.ts
```diff
@@ -680,6 +681,27 @@ export default class MockAdapter
...
 
+    async advancedSearchADTTwins(
+        _params: AdapterMethodParamsForAdvancedSearchADTwins
+    ) {
+        try {
+            await this.mockNetwork();
+
+            return new AdapterResult({
+                result: new ADTAdapterTwinsData({
+                    // TODO: filter out based on properties from params
```
msnyder-msft, line 692: it'd be great if we could do the filtering here so we get actual results in the story.

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
```diff
@@ -43,11 +42,35 @@ const AdvancedSearchModal: React.FC<IAdvancedSearchProps> = (props) => {
...
+    useEffect(() => {
+        const selectedTwins: IADTTwin[] = [];
+        if (searchForTwinAdapterData.adapterResult?.result?.data) {
+            searchForTwinAdapterData.adapterResult.result.data.value.map(
+                (twinData: IADTTwin) => {
+                    selectedTwins.push(twinData);
+                }
+            );
+        }
+        filteredTwins.current = selectedTwins;
```
msnyder-msft, line 71: why store the results in a ref? Don't you want to trigger a render when it updates that list?

src/Models/Constants/Types.ts
```diff
@@ -125,6 +125,11 @@ export type AdapterMethodParamsForSearchADTTwins = {
     continuationToken?: string;
 };
 
+export type AdapterMethodParamsForAdvancedSearchADTwins = {
+    query: string;
+    continuationToken?: string;
```
msnyder-msft, line 130: I would rather the components not need to worry about this. How do you plan to handle triggering the next page?

msnyder-msft: I thought we talked about including `top 1000` in the query. Then we could have the adapter simply keep resolving the pages until it hits the end.

msnyder-msft (commented):
Looks fine from the code perspective, just a comment about naming and one about handling the paging.

src/Adapters/ADTAdapter.ts
pasanchMSFT: I'll change the name, but I prefer something around `customTwinSearch` or something that talks about a free-form query.

msnyder-msft: hmm, `twinSearchByQuery`?

src/Adapters/ADTAdapter.ts
pasanchMSFT: There is a TODO item for this, in case the response yields paginated results have the details list scroll trigger a search with continuation token to get the remaining items.

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
pasanchMSFT: It will trigger a render, since the results are passed in as a prop to the list.

msnyder-msft: Hmmm. but why woudl this parent rerender to do the passing? I think this should be stored in state instead. That seems like an unnessecary risk. We _want_ this component to render when this chagnes so I don't see why we would make it a ref.

src/Adapters/ADTAdapter.ts
darsney: This is actually complex, we don't want to stream 100K twins to the client.  I think for now it's perhaps best to stop at, say, 1000, and show content in the results that says "First 1000, of many" and an info bubble that tells them to filter more.  This is very very similar to what we did for this problem in TSI

abalogun-dev (approved):
Functionally looks good. Mitch left some good comments

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
```diff
@@ -89,9 +112,11 @@ const AdvancedSearchModal: React.FC<IAdvancedSearchProps> = (props) => {
                     </div>
                     <div className={classNames.resultsContainer}>
                         <AdvancedSearchResultDetailsList
```
iremgokce, line 114: can we show "No results" text instead of showing the header of the table when the result is 0 item?
image

Also is there a reason why we made the Results title bold? or maybe we can use h4 instead of h3 what do you think @azimamansuri ?

iremgokce: also I think the options in the dropdown are bolder now, but after selecting the input text is lighter weight. Is there a reason why we made the options bolder? What do you think @azimamansuri?:
image

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
```diff
@@ -89,9 +112,11 @@ const AdvancedSearchModal: React.FC<IAdvancedSearchProps> = (props) => {
...
-                            twins={filteredTwins}
-                            searchedProperties={cols}
-                            adapter={new MockAdapter()}
+                            twins={filteredTwins.current}
+                            searchedProperties={Array.from(
```
iremgokce, line 116: can we make this const var with useMemo insread of creating the array each time?

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -249,13 +249,13 @@ const QueryBuilderRow: React.FC<IQueryBuilderRowProps> = (props) => {
...
-                                    text: 'And',
+                                    key: 'AND',
+                                    text: 'AND',
                                     selected: true
                                 },
                                 {
-                                    key: 'Or',
-                                    text: 'Or'
+                                    key: 'OR',
+                                    text: 'OR'
```
iremgokce, line 258: I think it would be cleaner to have enums for and/or and operations

src/Components/AdvancedSearch/AdvancedSearchModal.stories.local.tsx
```diff
@@ -0,0 +1,52 @@
...
+                new ADTAdapter(
+                    authenticationParameters.adt.hostUrl,
+                    new MsalAuthService(
+                        authenticationParameters.adt.aadParameters
+                    ),
+                    authenticationParameters.storage.blobContainerUrl
+                )
+            }
+            allowedPropertyValueTypes={[
+                'string',
```
iremgokce, line 37: is this list going to be the only allowed properties for advanced search, can we define this as const in AdvancedSearch.types.ts file? We used to use such const for property types before like `numericPropertyValueTypes`

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
```diff
@@ -43,11 +42,35 @@ const AdvancedSearchModal: React.FC<IAdvancedSearchProps> = (props) => {
...
+        refetchDependencies: [],
+        isAdapterCalledOnMount: false
+    });
+
     const executeQuery = (query: string) => {
-        alert(query);
+        searchForTwinAdapterData.callAdapter({
+            query: query,
+            // TODO: Define onScroll method on list to change continuation token
+            continuationToken: null
```
iremgokce, line 58: we need to keep track of the continuationToken and pass it here when you do the pagination, right?

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
msnyder-msft: +1 on the no data message. 
Should probably have a different message before the query is fired too "Build a query to search the twins graph" or something.

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
msnyder-msft: I'm not sure the creation process is going to be very intensive. Based on [the guidance](https://overreacted.io/before-you-memo/) I've read it seems they recommend you avoid memo unless it's truly an expensive operation.

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
msnyder-msft: I have asked him to remove this from the component level and keep it contained inside the adapter. At this point we don't plan to do pagination on the control level so to get all the data we have to do taht recursion internally. Though it might be in the next PR.

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
pasanchMSFT: Why do you want the parent to re-render? There are no visual updates in the parent that require a re-render.

src/Components/AdvancedSearch/AdvancedSearchModal.stories.local.tsx
pasanchMSFT: Good suggestion, I'll make this change

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
pasanchMSFT: removed

src/Adapters/MockAdapter.ts
pasanchMSFT: filtered if user selects equals

src/Components/AdvancedSearch/AdvancedSearchModal.tsx
pasanchMSFT: Added no data message for both scenarios

src/Adapters/MockAdapter.ts
```diff
@@ -681,16 +681,47 @@ export default class MockAdapter
         }
     }
 
-    async advancedSearchADTTwins(
-        _params: AdapterMethodParamsForAdvancedSearchADTwins
-    ) {
+    getFirstPropertyFromQuery = (query: string) => {
+        // Initial position index is the index after WHERE in the query
+        // used here to search for first property after the WHERE clause
+        const initialPositionIndex = 43;
```
msnyder-msft, line 687: i know it's only for mock data, but can we at least make it resilient to a single character change? Can we just look up the position of `where`?

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -146,52 +149,59 @@ const QueryBuilder: React.FC<IQueryBuilderProps> = (props) => {
 
     return (
         <div className={classNames.root}>
```
msnyder-msft, line 151: hsouldn't we have a stack here to separate the query section from the buttons? THe buttons should ideally have space between them and it'd be great if that came from a stack vs styling on each button individually

src/Components/AdvancedSearch/AdvancedSearch.stories.tsx
```diff
@@ -32,14 +30,7 @@ Base.args = {
...
     adapter: new MockAdapter(),
-    allowedPropertyValueTypes: [
-        'string',
-        'boolean',
-        'float',
-        'integer',
-        'double',
-        'long'
-    ],
+    allowedPropertyValueTypes: queryAllowedPropertyValueTypes,
```
msnyder-msft, line 33: since it's a constant, please make the name in all caps. QUERY_ALLOWED_.....

src/Components/AdvancedSearch/AdvancedSearch.tsx
```diff
@@ -85,45 +86,44 @@ const AdvancedSearchModal: React.FC<IAdvancedSearchProps> = (props) => {
...
-                            styles={classNames.subComponentStyles.icon}
-                        />
-                        <h3 id={'titleId'} className={classNames.headerText}>
-                            {t('advancedSearch.modalTitle')}
-                        </h3>
-                    </div>
-                    <p className={classNames.subtitle}>
-                        {t('advancedSearch.modalSubtitle')}
-                    </p>
+            <div className={classNames.headerContainer}>
```
msnyder-msft, line 89: if we are removing the base component, are we killing off the theme prop too?

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.styles.ts
```diff
@@ -13,7 +13,12 @@ export const getStyles = (
     _props: IAdvancedSearchResultDetailsListStyleProps
 ): IAdvancedSearchResultDetailsListStyles => {
     return {
-        root: [classNames.root],
+        root: [
+            classNames.root,
+            {
+                marginTop: 0
```
msnyder-msft, line 19: why does the root have a margin on top by default? Isn't it just a div?

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -95,29 +98,72 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
         }
     });
 
+    const getNoDataHeaderText = useCallback(() => {
+        if (additionalColumns.length) {
+            return 'No twins found';
```
msnyder-msft: don't forget to loc.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -95,29 +98,72 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
...
-                ariaLabelForSelectionColumn={t(
-                    'advancedSearch.toggleSelection'
-                )}
-                checkButtonAriaLabel={t('advancedSearch.selectRow')}
-                onRenderItemColumn={renderItemColumn}
-                selectionMode={SelectionMode.single}
-                selection={selection}
-                styles={classNames.subComponentStyles.detailsList}
-            />
+            {twinCount === 1000 && (
```
msnyder-msft: please pull the 1000 out to a constant that you also use in generating the query so they stay in sync.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
```diff
@@ -95,29 +98,72 @@ const AdvancedSearchResultDetailsList: React.FC<IAdvancedSearchResultDetailsList
...
+                <span>
+                    There were more than 1000 results, add more filters to
```
msnyder-msft: and you'll want to pass in the constant to this string so it can be in sync there too.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
```diff
@@ -54,13 +55,19 @@ export const getStyles = (
...
+                    width: 100
                 }
             },
             addButton: {
                 root: {
                     display: 'block',
-                    color: props.theme.palette.themePrimary
+                    color: props.theme.palette.themePrimary,
+                    padding: 0,
+                    marginLeft: 0
```
msnyder-msft, line 67: why are we manipulating the spacing so much on a button? these are usually pretty standard.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
```diff
@@ -102,35 +109,57 @@ export const getRowStyles = (
...
+        ],
         subComponentStyles: {
             deleteButton: {
                 rootDisabled: {
                     background: 'transparent'
                 }
+            },
+            propertyIcon: {
+                root: {
+                    paddingRight: 4,
```
msnyder-msft, line 130: ideally we would just throw this in a Stack and let it take care of the spacing.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.types.ts
```diff
@@ -96,9 +118,20 @@ export interface PropertyOption {
     data: {
         name: string;
         type: PropertyValueType;
+        icon: string;
```
msnyder-msft, line 121: nit: `iconName` would be more clear since I'm sure you pass this into the Icon control.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderUtils.ts
```diff
@@ -154,14 +154,14 @@ export const getOperators = (
 };
 
 export const buildQuery = (querySnippets: QueryRowData[]) => {
-    let fullQuery = `SELECT *\nFROM DIGITALTWINS T\n`;
+    let fullQuery = `SELECT TOP(1000)\nFROM DIGITALTWINS T\nWHERE `;
```
msnyder-msft, line 157: please use the constant here that you use to check for 1000 rows in the response.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.types.ts
```diff
@@ -169,6 +169,15 @@ export const numericPropertyValueTypes: PropertyValueType[] = [
     'long'
 ];
 
+export const queryAllowedPropertyValueTypes: PropertyValueType[] = [
```
msnyder-msft, line 172: name should be all caps since it's a const.

src/Adapters/ADTAdapter.ts
```diff
@@ -472,6 +474,26 @@ export default class ADTAdapter implements IADTAdapter {
         );
     }
 
+    searchTwinsByQuery(params: AdapterMethodParamsForSearchTwinsByQuery) {
+        const adapterMethodSandbox = new AdapterMethodSandbox(this.authService);
+        return adapterMethodSandbox.safelyFetchDataCancellableAxiosPromise(
```
msnyder-msft, line 479: I'm still expecting to see the recursion to fetch all the pages of results in this PR since it's primarily for the querying logic.

msnyder-msft (commented):
Left mostly nits. Biggest thing I see missing right now is the recursion to fetch all the pages of results.

src/Components/AdvancedSearch/AdvancedSearch.stories.tsx
pasanchMSFT: I'm following convention of other consts in that file, ModelledPropertyBuilder.types.ts. I would prefer not to change all of them to all caps.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.tsx
pasanchMSFT: you are looking at an old version of this, should already be loc'ed

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.types.ts
pasanchMSFT: I'm following convention of other consts in that file, ModelledPropertyBuilder.types.ts. I would prefer not to change all of them to all caps.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
pasanchMSFT: Stack would make the buttons flex to 100% width, unless I wrap them in a div and style that. I feel this is a simpler solution.

src/Components/AdvancedSearch/AdvancedSearch.tsx
pasanchMSFT: It should already be removed, check if you are seeing the latest changes

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.styles.ts
pasanchMSFT: stack props, nice catch

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
msnyder-msft: I'm referring to this. 
image
Which results in it looking like this.
image

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
pasanchMSFT: Had to remove styling to align with the primary button. Since designs called for aligning text to the edge of the primary button I had to remove the padding + margin applied.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.styles.ts
pasanchMSFT: Had to do some shenanigans to allow text overflow ellipsis, I can try with Stack but don't think it will be much cleaner.

src/Components/ModelledPropertyBuilder/ModelledPropertyBuilder.types.ts
msnyder-msft: hmmm. I'm biased. I would rather know something is a constant and not a variable. I wouldn't chang ethe rest, but I would question whether this const is really for the ModeledPropertyBuilder since your component is separate from that. Shouldn't this be out in your own file for your component?

pasanchMSFT: There are shared consts for propertyValueTypes in this file, wanted to follow that pattern as well.

msnyder-msft: we should not be cross referencing between components. Please pull those shared consts out to a common place and use them there.

iremgokce (approved):
Looks good to me!

msnyder-msft (approved):

PR #689, opened 2022-08-31
[Telemetry] Wire up telemetry events for KPI metrics & initial user actions

msnyder-msft:
### Summary of changes 🔍 
Creating the initial infrastructure and events to emit KPI telemetry of the user's config file. The core of the changes are in the `BlobAdapterUtility` file. This is where we iterate through the entire config and emit events for every scene, behavior and element in the file.
The data will not actually get emitted to any telemetry service as part of Cardboard. That will be up to the hosting app to send it to a telemetry service. 
In our case, we have added that connection in the Flights app to emit the data to application insights and show it in a [PowerBI report](https://msit.powerbi.com/links/hBV4U-Y5QF?ctid=72f988bf-86f1-41af-91ab-2d7cd011db47&pbi_source=linkShare).

#### Event names and properties
You can see that all of the event names and the properties attached to each event are denoted in the `TelemetryConstants` file. This will help us to have a single place to look for the property names & event names. 
Note that there is a `TelemetryEvents` object as well as a `TelemetryMetrics` object. Right now we don't have any metrics but we likely will in the future. 
Each of those objects is broken down into the following structure to help group events together.

> - App region (Builder/Viewer/Adapter)
>    - High level component (SceneList, ElementForm, etc)
>       - User action/System action
>          - Event/action
>             - variant (ex: initiate, confirm, cancel)
>             - eventName: the name of the beacon to emit
>             - properties: names of the custom properties that the event will have attached to it

#### Metadata
There are a few properties that will be attached to every telemetry event:
- Current Environment id (hash)
- Current Scene id (hash)

These are updated anytime the deeplink context is has a value change so they will always reflect the current state of the app. We use the `md5` library for generating the hashes. The EnvironmentId is based on a combination of storageUrl and the ADT instance URL. The scene id is simply a hash of the GUID stored in the config file.

#### System action telemetry
Some events will be emitted by the system without user interaction. I'm defining these as "SystemAction" events. These will be most commonly sent from Adapters or services where we want to send a KPI type metric so we can identify patterns in the configs or usage of the app rather than specific user actions.
An example of these can be found in the `BlobAdapterUtility.ts` where we parse the config and emit events for every scene, behavior and element. These will be sent when the config is fetched and does not get triggered by a particular user action.

#### User action telemetry
Most events from the app will eventually by "UserAction" events which are prompted by a user taking some action in the app.
User actions will have a few extra properties attached to them which are not required for system actions coming from events in adapters. We want to capture a few extra points such as the ComponentName, AppRegion and the Gesture (hover, click) that triggered that event. Those will be required in addition to the event name when utilizing the `useTelemetry` hook and calling the `sendEventTelemetry` method. This will help us to slice and dice the data in the cases where the same event name is emitted by multiple different places in the app. Rather that proliferating the event names and using different names in multiple places, we can use the same event and simply slice by the "source"

An example of this kind of event can be seen in the `SceneList.tsx` file where we capture events when the user clicks on a list item or chooses to create/edit/delete a scene.

#### Debugging telemetry
If you run the following command in your browser console, you can enable telemetry logging (running locally or in production) to see all telemetry events that are being fired. NOTE: This does not mean it was sent to the server. This just means the hosting app was notified, it's up to that app to determine whether to send it to a telemetry service.

```ts
localStorage.setItem('cardboard.debug.telemetryLogging', true);
```

### Testing 🧪
first, set the local storage key for logging & then reload the page to see the events get fired

- [ ] click the edit button on a scene in the scene list --> see a telemetry event fire
- [ ] click 'create scene' on the scene list page --> see the telemetry event fire
- [ ] launch the app and see the KPI events fire for your scenes

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Adapters/BlobAdapterUtility.ts
+++ b/src/Adapters/BlobAdapterUtility.ts
@@ -129,29 +129,25 @@ function sendBehaviorTelemetry(behavior: IBehavior, sceneHash: string) {
     });
 
     // capture the Behavior level metrics
-    const sceneEvent =
-        TelemetryEvents.Adapter.SceneLoad.SystemAction.ParseBehavior;
+    const event = TelemetryEvents.Adapter.SceneLoad.SystemAction.ParseBehavior;
     TelemetryService.sendEvent(
         new TelemetryEvent({
-            name: sceneEvent.eventName,
+            name: event.eventName,
             customProperties: {
-                [sceneEvent.properties.countAliases]: aliasCount,
-                [sceneEvent.properties.countDataSources]: dataSourceCount,
-                [sceneEvent.properties.countElements]: elementCount,
-                [sceneEvent.properties.countVisualBadgeType]:
-                    visualStats.badgeCount,
-                [sceneEvent.properties.countVisualColorType]:
-                    visualStats.colorCount,
-                [sceneEvent.properties.countVisuals]: visualStats.totalCount,
-                [sceneEvent.properties.countWidgetGaugeType]:
+                [event.properties.countAliases]: aliasCount,
+                [event.properties.countDataSources]: dataSourceCount,
+                [event.properties.countElements]: elementCount,
+                [event.properties.countVisualBadgeType]: visualStats.badgeCount,
+                [event.properties.countVisualColorType]: visualStats.colorCount,
+                [event.properties.countVisuals]: visualStats.totalCount,
+                [event.properties.countWidgetGaugeType]:
                     visualStats.widgets.gaugeCount,
-                [sceneEvent.properties.countWidgetLinkType]:
+                [event.properties.countWidgetLinkType]:
                     visualStats.widgets.linkCount,
-                [sceneEvent.properties.countWidgetPropertyType]:
+                [event.properties.countWidgetPropertyType]:
                     visualStats.widgets.propertyCount,
-                [sceneEvent.properties.countWidgets]:
-                    visualStats.widgets.totalCount,
-                [sceneEvent.properties.sceneHash]: sceneHash
+                [event.properties.countWidgets]: visualStats.widgets.totalCount,
+                [event.properties.parentSceneHash]: sceneHash
             },
             triggerType: TelemetryTrigger.SystemAction
         })
@@ -161,8 +157,8 @@ function sendBehaviorTelemetry(behavior: IBehavior, sceneHash: string) {
 /** send the KPI telemetry related to individual Elements */
 function sendElementTelemetry(element: IElement, sceneHash: string) {
     const type = element.type;
-    let meshCount = -1;
-    let aliasCount = -1;
+    let meshCount = 0;
+    let aliasCount = 0;
     // only capture the count when it's actually an 'element'
     if (ViewerConfigUtility.isTwinToObjectMappingElement(element)) {
         meshCount = element.objectIDs?.length;
@@ -172,16 +168,15 @@ function sendElementTelemetry(element: IElement, sceneHash: string) {
     }
 
     // capture the Element level metrics
-    const sceneEvent =
-        TelemetryEvents.Adapter.SceneLoad.SystemAction.ParseElement;
+    const event = TelemetryEvents.Adapter.SceneLoad.SystemAction.ParseElement;
     TelemetryService.sendEvent(
         new TelemetryEvent({
-            name: sceneEvent.eventName,
+            name: event.eventName,
             customProperties: {
-                [sceneEvent.properties.countAliases]: aliasCount,
-                [sceneEvent.properties.countMeshes]: meshCount,
-                [sceneEvent.properties.sceneHash]: sceneHash,
-                [sceneEvent.properties.type]: type
+                [event.properties.countAliases]: aliasCount,
+                [event.properties.countMeshes]: meshCount,
+                [event.properties.parentSceneHash]: sceneHash,
+                [event.properties.elementType]: type
             },
             triggerType: TelemetryTrigger.SystemAction
         })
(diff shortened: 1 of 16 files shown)


darsney (approved):
The proof is in the PBI dashboards for me on this bad boy.  Let's hold off on enabling logging of this stuff for real for real until Riley validates that the telemetry shape allows him to answer the questions he wants to ask.  

Let's also make sure the team knows how to use this on their features (perhaps a little documentation in the wiki). 

Comments on code are deferred to dev team

abalogun-dev (approved):
Functionally works.

src/Adapters/BlobAdapterUtility.ts
```diff
@@ -0,0 +1,234 @@
...
+                [event.properties.pollingDelay]: scenePollingDelay,
+                [event.properties.sceneHash]: sceneHash
+            },
+            triggerType: TelemetryTrigger.SystemAction
+        })
+    );
+}
+
+/** send the KPI telemetry related to the Config file as a whole */
+function sendConfigTelemetry(data: I3DScenesConfig) {
```
iremgokce, line 217: after saving a scene, ParseConfiguration events are fired multiple times, is there a way to make sure they are called once?
image

msnyder-msft: In this case you're seeing an event per Element and per Behavior in that scene. That is actually by design. BUT if what you're saying is you don't want that event do go out for the same behavior on each save, then yes, we could throttle it to only send once a day per environment. There's not really any harm though in firing more except that it has more data we have to crawl through. It should be straightforward to eliminate the duplicates in the reports by doing a top 1 by environmentId.

src/Components/SceneList/SceneList.tsx
```diff
@@ -209,8 +246,18 @@ const SceneList: React.FC<SceneListProps> = ({
...
                                 setIsSceneDialogOpen(true);
+                                const telemetryEvent =
+                                    TelemetryEvents.Builder.SceneList.UserAction
+                                        .EditScene.Initiate;
                                 sendEventTelemetry({
-                                    name: 'Edit scene - open'
+                                    name: telemetryEvent.eventName,
+                                    customProperties: {
+                                        [telemetryEvent.properties
+                                            .itemIndex]: itemIndex
```
iremgokce, line 256: itemIndex seems to be a string type in the `TelemetryConstants` file, but it seems it is assigned as a number here. can we change the itemIndex name to itemKey in the telemetry constants?
image

src/Models/Constants/TelemetryConstants.ts
```diff
@@ -0,0 +1,191 @@
...
+                    Cancel: {
+                        eventName: 'SceneList.DeleteScene.Cancel'
+                    },
+                    Confirm: {
+                        eventName: 'SceneList.DeleteScene.Confirm'
+                    },
+                    Initiate: {
+                        eventName: 'SceneList.DeleteScene.Initiate',
+                        properties: {
+                            itemIndex: 'itemIndex'
```
iremgokce, line 151: why itemIndex is same for other type of actions below?

src/Models/Services/TelemetryService/TelemetryService.ts
```diff
@@ -1,21 +1,106 @@
...
+    // #endregion
+
+    // #region public setters
+
+    /**
+     * store the current adt instance
+     * @param adtInstanceUrl
+     */
+    static setAdtInstance(adtInstanceUrl: string): void {
+        this.adtInstanceHash = Md5.hashStr(adtInstanceUrl);
```
iremgokce, line 64: just wondering where do we keep the mapping for hash to url? I mean after keeping the urls as hash in the telemetry how to we backtrack to the urls for debugging purposes?

src/Models/Services/TelemetryService/TelemetryService.types.ts
```diff
@@ -22,8 +28,8 @@ export type SeverityLevel =
     | 'Critical';
 
 export interface IBaseTelemetryParams {
-    name: string;
     customProperties?: CustomProperties;
```
iremgokce, line 31: CustomProperties are defined as Record<string, any>, is it possible to define a more strict interface for it by adding what do we expect there?

src/Models/Constants/TelemetryConstants.ts
```diff
@@ -0,0 +1,191 @@
...
+ * The metrics the app can emit
+ * Structure is like this:
+ *  - App region (Builder/Viewer)
+ *    - High level component (SceneList, ElementForm, etc)
+ *      - User action/System metric
+ *        - Event/action
+ *           - eventName: the name of the beacon to emit
+ *           - properties: names of the custom properties that the event will have attached to it
+ */
+export const TelemetryMetrics = { Adapter: {}, Builder: {}, Viewer: {} };
```
iremgokce, line 191: do you use this const somewhere? I didn't see. did you intend to make it as interface and use for TelemetryEvents above?

iremgokce (approved):
just left couple of comments, otherwise looks good to me!

src/Components/SceneList/SceneList.tsx
msnyder-msft: the const is the property name that is used to log to Application Insights. So yes, the key is a string, but the value can be whatever you want. That's why the custom properties are a `string` to `any` mapping.

src/Models/Constants/TelemetryConstants.ts
msnyder-msft: because I want to define the properties that each event has, and I wanted the names of those properties to be consistent. They don't HAVE to match, but it's easier when you're querying to know. And it's flexible if we want to change them for some reason. I'm only pulling common names out to constants if they really need to match exactly everywhere for some reason (like the sceneHash)

src/Models/Services/TelemetryService/TelemetryService.ts
msnyder-msft: We don't that's very intentional. We don't want that private data in clear text anywhere in our telemetry. We won't be able to reverse the hash without the original value to hash again. The point is to provide a unique and consistent id without being able to tie it to a particular customer.

src/Models/Services/TelemetryService/TelemetryService.types.ts
msnyder-msft: the custom properties are inherently generic and unknown. We don't want a strict interface there because every event can then provide it's own collection of properties it wants to capture. That is how we get the values that are important to that event. Besides, the telemetry service doesn't really care what is in that collection, it simply passes it through to the logger. If we want to require certain metadata for an event, it should be one of the params provided to the look like we do for ComponentName.

src/Models/Constants/TelemetryConstants.ts
msnyder-msft: Not yet, I just made it so it's ready if we want to create metrics in addition to events. They are a specific type of event with different properties. It will be a different collection of beacons that are specifically for that type.

PR #693, opened 2022-09-06
Adding auto complete functionality to the Advanced Search control

msnyder-msft:
### Summary of changes 🔍 
[Product Backlog Item 13048797](https://msazure.visualstudio.com/One/_workitems/edit/13048797): Domain Expert can discover and confidently select digital twins when defining a Linked Twin

Today, when you are searching for a particular value in the Advanced Search, you have to know the right values to type to get a perfect match. We already have a component that can query the graph and return the values for a given property name that exist in the graph `TwinSearchDropdown`. Today that component is specific to $dtId but we can easily adapt it to work for any property name. 
This allows us to get a form of autocomplete in the advanced search form that gives users a dropdown menu that helps them get the right values right off the bat.

NOTE: this auto complete only applies to the string properties and not any other type of property.

image

Also fixed some styling on the property dropdown since we now share styles between the two React-Select components.
image

### Testing 🧪
- [ ] Open the advanced search story and select the 'CarName' property to see the auto complete component show up
- [ ] Try clearing the value
- [ ] Select an option from the list
- [ ] Type a value and see it auto suggest options
- [ ] Select a suggested value
- [ ] Type an incomplete value and click out of the searchbox to use the partial value

### Checklist ✔️
- [ ] Linked associated issue (if present)
- [ ] Added relevant labels
- [ ] Requested developer reviews
- [ ] Request UI review from design / PM
- [ ] Verify all code checks are passing

PR #694, opened 2022-09-09
Adding auto complete functionality to the Advanced Search control

msnyder-msft:
### Summary of changes 🔍 
NOTE: Accidentally merged original PR to a feature branch. See comments on #693 

[Product Backlog Item 13048797](https://msazure.visualstudio.com/One/_workitems/edit/13048797): Domain Expert can discover and confidently select digital twins when defining a Linked Twin

Today, when you are searching for a particular value in the Advanced Search, you have to know the right values to type to get a perfect match. We already have a component that can query the graph and return the values for a given property name that exist in the graph `TwinSearchDropdown`. Today that component is specific to $dtId but we can easily adapt it to work for any property name. 
This allows us to get a form of autocomplete in the advanced search form that gives users a dropdown menu that helps them get the right values right off the bat.

NOTE: this auto complete only applies to the string properties and not any other type of property.

image

Also fixed some styling on the property dropdown since we now share styles between the two React-Select components.
image

### Testing 🧪
- [ ] Open the advanced search story and select the 'CarName' property to see the auto complete component show up
- [ ] Try clearing the value
- [ ] Select an option from the list
- [ ] Type a value and see it auto suggest options
- [ ] Select a suggested value
- [ ] Type an incomplete value and click out of the searchbox to use the partial value

### Checklist ✔️
- [ ] Added relevant labels


pasanchMSFT (approved):

.eslintrc.json
```diff
@@ -1,18 +1,23 @@
...
+    ],
     "extends": [
         "eslint:recommended",
         "plugin:@typescript-eslint/recommended",
         "prettier",
         "prettier/@typescript-eslint"
     ],
     "rules": {
-        "no-debugger": "error",
+        "no-debugger": "warn",
```
pasanchMSFT, line 15: do you want to check this in?

PR #696, opened 2022-09-12
Integrating Advanced search to Element form

pasanchMSFT:
### Summary of changes 🔍 
This PR integrates Advanced Search functionality with the Element form. This new feature allows the users to select a Primary twin after building a query to search for it. Previously the Element form only allowed users to select a Twin by searching for its `$dtId`, now both `$dtId` and advanced search are enabled.

Also included in this PR a button was added to the Element form to show a callout containing the `PropertyInspector`. The `PropertyInspector` component here will allow users to know which properties are part of the selected Twin in case they want to search for another Twin of the same model. 

### Testing 🧪
PropertyInspector:
- Navigate to Element form
- If a Twin is selected: (else select Twin from list)
  - Click the first icon besides the Primary Twin dropdown.
  - Make sure properties are correct for the selected Twin model

Advanced search:
- Navigate to Element form
- Click the second icon besides the Primary Twin dropdown
- Create a query to search for Twins, in Mock data any Operator other than Equals will return all available data
  - An example of an Equals operation that will yield some filtered values is: 
image

- Select a twin from the list
- Click on the `Select` button in the modal
- Review:
  - If dropdown changed to selected twin
  - If Property inspector shows correct twin data
  - If saving the form persists data

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/AdvancedSearch/AdvancedSearch.tsx
+++ b/src/Components/AdvancedSearch/AdvancedSearch.tsx
@@ -31,9 +31,12 @@ const getClassNames = classNamesFunction<
     IAdvancedSearchStyles
 >();
 
-const stackTokens: IStackTokens = {
-    maxHeight: 550
+const CONTENT_MAX_HEIGHT = 515;
+const containerStackTokens: IStackTokens = { childrenGap: 8 };
+const contentStackTokens: IStackTokens = {
+    maxHeight: CONTENT_MAX_HEIGHT
 };
+const footerStackTokens: IStackTokens = { childrenGap: 8 };
 
 const AdvancedSearch: React.FC<IAdvancedSearchProps> = (props) => {
     const {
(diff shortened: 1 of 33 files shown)


src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.types.ts
```diff
@@ -16,7 +16,7 @@ export interface IAdvancedSearchResultDetailsListProps {
     containsError: boolean;
     isLoading: boolean;
     /* Callback function from parent on what to do once the user selects or deselects a twin. */
-    onTwinSelection: (IADTTwin) => void;
+    onTwinIdSelect: (twinId: string) => void;
```
msnyder-msft, line 19: nit: `onRowSelect`? or `onSelectRow`

src/Components/TwinPropertySearchDropdown/TwinPropertySearchDropdown.types.ts
```diff
@@ -10,6 +10,10 @@ import ADTAdapter from '../../Adapters/ADTAdapter';
 import MockAdapter from '../../Adapters/MockAdapter';
 import { ITooltipCalloutContent } from '../TooltipCallout/TooltipCallout.types';
 
+export type PropertyValueHandle = {
+    updateValue: (newValue: string) => void;
```
msnyder-msft, line 14: nit: `setSearchText` or `setValue`.

src/Components/TwinPropertySearchDropdown/TwinPropertySearchDropdown.tsx
```diff
@@ -197,6 +209,13 @@ const TwinPropertySearchDropdown: React.FC<ITwinPropertySearchDropdownProps> = (
         );
     };
 
+    // imperative handle to expose modifying the state
+    useImperativeHandle(ref, () => ({
+        updateValue: (newValue: string) => {
```
msnyder-msft, line 214: I'm thinking we probably need to create an option for the value as well right? what happens if you set it to a value not found in the list? Ideally wouldn't happen but it should be resilient to that and add the item if not already there. I think you can just set the `selectedOption` since we add the selected option as an item if it's not in the list.

src/Components/TwinPropertySearchDropdown/TwinPropertySearchDropdown.tsx
```diff
@@ -47,22 +56,25 @@ const SuggestionListScrollThresholdFactor = 40;
...
-    labelIconName,
-    labelTooltip,
-    onChange,
-    noOptionsText,
-    placeholderText,
-    resetInputOnBlur = true,
-    searchPropertyName,
-    styles
-}) => {
+const TwinPropertySearchDropdown = (
```
msnyder-msft, line 59: where do we define which interface we use for the props?

src/Components/AdvancedSearch/AdvancedSearch.tsx
```diff
@@ -123,6 +132,23 @@ const AdvancedSearch: React.FC<IAdvancedSearchProps> = (props) => {
                             }
                         }}
                     />
+                    <div>
```
msnyder-msft, line 135: we need the wrapper div?

src/Components/AdvancedSearch/AdvancedSearch.tsx
```diff
@@ -123,6 +132,23 @@ const AdvancedSearch: React.FC<IAdvancedSearchProps> = (props) => {
...
+                        <Stack
```
msnyder-msft, line 136: the buttons should not float up. They should get pegged to the bottom. 
Look into whether we can use a ModalFooter (I think that's what it's called) and then Modal just takes care of the scrolling the content and all that jazz for us.
image

src/Components/AdvancedSearch/AdvancedSearch.tsx
msnyder-msft, line 135: Where's the spacing above the buttons? Seems like the root of the modal needs a stack to space the buttons off from teh list control.
image

msnyder-msft (approved):
Looking really good. Needs a pass on the styling though.

src/Components/AdvancedSearch/Internal/AdvancedSearchResultDetailsList/AdvancedSearchResultDetailsList.types.ts
pasanchMSFT: If it where generic I would change it to any of those, but I prefer to specify what is being selected.

src/Components/TwinPropertySearchDropdown/TwinPropertySearchDropdown.tsx
pasanchMSFT: Should be fixed, please take a look at it to confirm

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -332,7 +332,7 @@ const QueryBuilderRow: React.FC<IQueryBuilderRowProps> = (props) => {
             )}
             <div className={classNames.inputColumn}>
                 {/* Property */}
-                <Select<PropertyOption>
+                <Select
```
msnyder-msft, line 335: why did we have to drop this typing?

msnyder-msft (approved):
looks good.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
pasanchMSFT: This is the type I'm struggling with in tests, it breaks the styles for some reason

msnyder-msft (approved):

src/Models/Services/StoryUtilities.tsx
```diff
@@ -132,7 +132,14 @@ export const selectReactSelectOption = (
     fireEvent.mouseDown(
         document.getElementsByClassName(`${className}__control`)[0]
     );
-    fireEvent.click(
-        document.getElementsByClassName(`${className}__option`)[optionIndex]
+    // Wait for callout to pop-up
+    setTimeout(
```
msnyder-msft, line 136: this might work locally, but it will break when we get back to chromatic tests because the test will end and take the picture before this timeout fires.

msnyder-msft: I typically do an `await sleep(1)` and then it will give the render loop a chance to complete and then comes back here to execute. It just pushes your function execution to the bottom of the javascript execution stack.

abalogun-dev (approved):
This looks really good, codewise and functionally. Only nit comment i have is, I'm not really a fan of the ellipses in the placeholder text here. Make it feel like there is important info that you can't see. 
image

pasanchMSFT:
> This looks really good, codewise and functionally. Only nit comment i have is, I'm not really a fan of the ellipses in the placeholder text here. Make it feel like there is important info that you can't see. image

Nice catch. Made the change

PR #704, opened 2022-09-22
Bug Fix: Overflow tabs cause react error

msnyder-msft:
### Summary of changes 🔍 
The tabs overflow menu was throwing a react error anytime it was opened. This only showed up in other languages so it wasn't caught internally. 
The issue was that the custom rendering function on each tab item was calling a hook and since it got pushed inside the overflow set it changed the hook execution and broke the React rules for hooks always executing in the same order.

Original bug
react error

Fixed
fixed

### Testing 🧪
- [x] Switch your language to something that causes the behavior form tabs to overflow. Then go to the behavior form. Click the overflow set and use the app as normal. It used to crash

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -21,7 +21,7 @@
         "@typescript-eslint/no-var-requires": 0,
         "@typescript-eslint/no-explicit-any": 0,
         "@typescript-eslint/no-unused-vars": [
-            "warn",
+            "error",
             {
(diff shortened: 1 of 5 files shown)


darsney (approved):
LGTM, SEND IT!

iremgokce (approved):
good finding!

PR #705, opened 2022-09-23
Bug Fix: Advanced Search - Fix query builder property search to search the right property

msnyder-msft:
### Summary of changes 🔍 
The query builder was always searching twin ids instead of the selected property.
Before
image

After
image

**Fixing duplicate values in dropdown**
Before

After
image

**Fixing query failing due to new line characters**
Before

After
image

### Testing 🧪
- [x]  select a string type property in advanced search and see the values in the auto complete are showing the values for that property

### Checklist ✔️
- [ ] Added relevant labels


pasanchMSFT (approved):
The fix looks good to me, just curious about the behavior of the component with an empty string.

darsney (approved):

darsney (changes requested):
Yeahh this just doesnt work, it always 400s

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilderRow.tsx
```diff
@@ -271,7 +268,7 @@ const QueryBuilderRow: React.FC<IQueryBuilderRowProps> = (props) => {
                     noOptionsText={t('advancedSearch.noOptionsFoundMessage')}
                     placeholderText={t('advancedSearch.valueFieldPlaceholder')}
                     resetInputOnBlur={false}
-                    searchPropertyName={DTID_PROPERTY_NAME}
+                    searchPropertyName={selectedProperty?.data.name || ''}
```
iremgokce, line 271: is '' a valid search property name? should we keep DTID_PROPERTY_NAME in the right side of the ||?

iremgokce (approved):

darsney (approved):
LGTM

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
```diff
@@ -18,9 +18,12 @@ import {
 import QueryBuilderRow from './QueryBuilderRow';
 import { buildQuery } from './QueryBuilderUtils';
 import { useTranslation } from 'react-i18next';
-import { createGUID } from '../../../../Models/Services/Utils';
+import { createGUID, getDebugLogger } from '../../../../Models/Services/Utils';
 import { PropertyValueType } from '../../../../Models/Constants';
 
+const debugLogging = true;
```
pasanchMSFT, line 24: remove logging

src/Components/TwinPropertySearchDropdown/TwinPropertySearchDropdown.tsx
```diff
@@ -108,7 +109,7 @@ const TwinPropertySearchDropdown = (
     const searchTwinAdapterData = useAdapter({
         adapterMethod: (params: AdapterMethodParamsForSearchADTTwins) =>
             adapter.searchADTTwins(params),
-        refetchDependencies: [],
+        refetchDependencies: [searchPropertyName],
```
pasanchMSFT, line 112: Why would you want to re-fetch here? This method is done on click.

src/Models/Services/Utils.ts
```diff
@@ -483,6 +483,19 @@ export function sortAscendingOrDescending<T>(
...
 
+/**
+ * remove duplicate objects from an array
+ * @param array array of objects to operate on
+ * @param key key of the property to use as the discriminator
+ * @returns a new copy of the array with only unique values
+ */
+export function removeDuplicatesFromArray<T>(array: T[], key: keyof T) {
+    const check = new Set<string>();
+    return array.filter(
```
pasanchMSFT, line 494: Why not `Array.from(check)`?

src/Components/TwinPropertySearchDropdown/TwinPropertySearchDropdown.tsx
msnyder-msft: this is in my twin property search. :)

src/Models/Services/Utils.ts
msnyder-msft: Short answer: I just copied a stack overflow post.
Long answer: I'm not sure you can have a set of objects. Or at least it doesn't help you with getting a distinct list. So instead, we just keep a list of distinct keys and only include items where the key is not found in the set already.

src/Components/AdvancedSearch/Internal/QueryBuilder/QueryBuilder.tsx
msnyder-msft: good catch. Fixed

src/Models/Services/Utils.ts
pasanchMSFT: Fair enough, LGTM then

PR #707, opened 2022-09-30
Creating a generic modal component

msnyder-msft:
### Summary of changes 🔍 
We have a lot of modals and it seemed logical that we should standardize some of that styling and components when possible so this is a generic control that conforms to most of our default styles and patterns.

This is heavily lifted from the work @pasanchMSFT did on the AdvancedSearch modal so props there! It was super easy to extract and convert to a generic implementation.

Generic modal that has title, subtitle
image

Generic with destructive action
image

Customizable title and sub title
image

I migrated AdvancedSearch to the generic control as an example
image

### Testing 🧪
- Check out the `CardboardModal` stories.
image

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/CardboardModal/CardboardModal.styles.ts
+++ b/src/Components/CardboardModal/CardboardModal.styles.ts
@@ -1,0 +1,67 @@
+import { FontSizes, FontWeights } from '@fluentui/react';
+import {
+    ICardboardModalStyleProps,
+    ICardboardModalStyles
+} from './CardboardModal.types';
+
+export const classPrefix = 'cb-oat-modal';
+const classNames = {
+    content: `${classPrefix}-content`,
+    footer: `${classPrefix}-footer`,
+    headerContainer: `${classPrefix}-headerContainer`,
+    subtitle: `${classPrefix}-subtitle`,
+    title: `${classPrefix}-title`,
+    titleContainer: `${classPrefix}-titleContainer`
+};
+export const getStyles = (
+    _props: ICardboardModalStyleProps
+): ICardboardModalStyles => {
+    return {
+        content: [
+            classNames.content,
+            {
+                height: '100%',
+                overflow: 'auto'
+            }
+        ],
+        footer: [classNames.footer],
+        headerContainer: [classNames.headerContainer],
+        title: [
+            classNames.title,
+            {
+                margin: 0,
+                fontSize: FontSizes.size24,
+                fontWeight: FontWeights.semibold
+            }
+        ],
+        titleContainer: [
+            classNames.titleContainer,
+            {
+                display: 'flex',
+                paddingBottom: 8
+            }
+        ],
+        subtitle: [classNames.subtitle],
+        subComponentStyles: {
+            modal: {
+                main: {
+                    height: 690,
+                    width: 940,
+                    padding: 20
+                },
+                scrollableContent: {
+                    height: '100%'
+                }
+            },
+            icon: {
+                root: {
+                    textAlign: 'center',
+                    alignSelf: 'center',
+                    paddingRight: 12,
+                    paddingTop: 8,
+                    fontSize: FontSizes.size20
+                }
+            }
+        }
+    };
+};
--- a/src/Components/CardboardModal/CardboardModal.styles.ts
+++ b/src/Components/CardboardModal/CardboardModal.styles.ts
@@ -16,7 +16,7 @@ const classNames = {
 export const getStyles = (
     props: ICardboardModalStyleProps
 ): ICardboardModalStyles => {
-    const { isDestructiveFooterActionVisible, theme } = props;
+    const { splitFooter, theme } = props;
     return {
         content: [
             classNames.content,
@@ -68,7 +68,7 @@ export const getStyles = (
             },
             footerStack: {
                 root: {
-                    justifyContent: isDestructiveFooterActionVisible
+                    justifyContent: splitFooter
                         ? 'space-between !important'
                         : 'end'
                 }
--- a/src/Components/CardboardModal/CardboardModal.styles.ts
+++ b/src/Components/CardboardModal/CardboardModal.styles.ts
@@ -14,8 +14,10 @@ const classNames = {
     titleContainer: `${classPrefix}-titleContainer`
 };
 export const getStyles = (
-    _props: ICardboardModalStyleProps
+    props: ICardboardModalStyleProps
 ): ICardboardModalStyles => {
+    const { isDestructiveFooterActionVisible, theme } = props;
+    console.log('***Is destructive', isDestructiveFooterActionVisible);
     return {
         content: [
             classNames.content,
@@ -43,6 +45,35 @@ export const getStyles = (
         ],
         subtitle: [classNames.subtitle],
         subComponentStyles: {
+            destructiveButton: {
+                root: {
+                    backgroundColor: 'transparent',
+                    borderColor: 'var(--cb-color-text-danger)',
+                    color: 'var(--cb-color-text-danger)'
+                },
+                rootHovered: {
+                    backgroundColor: 'var(--cb-color-text-danger)',
+                    borderColor: theme.palette.white,
+                    color: theme.palette.white
+                },
+                rootFocused: {
+                    backgroundColor: 'var(--cb-color-text-danger)',
+                    borderColor: theme.palette.white,
+                    color: theme.palette.white
+                },
+                rootPressed: {
+                    backgroundColor: 'var(--cb-color-text-danger)',
+                    borderColor: theme.palette.white,
+                    color: theme.palette.white
+                }
+            },
+            footerStack: {
+                root: {
+                    justifyContent: isDestructiveFooterActionVisible
+                        ? 'space-between !important'
+                        : 'end'
+                }
+            },
             modal: {
                 main: {
                     height: 690,
--- a/src/Components/CardboardModal/CardboardModal.styles.ts
+++ b/src/Components/CardboardModal/CardboardModal.styles.ts
@@ -17,7 +17,6 @@ export const getStyles = (
     props: ICardboardModalStyleProps
 ): ICardboardModalStyles => {
     const { isDestructiveFooterActionVisible, theme } = props;
-    console.log('***Is destructive', isDestructiveFooterActionVisible);
     return {
         content: [
             classNames.content,
--- a/src/Components/CardboardModal/CardboardModal.tsx
+++ b/src/Components/CardboardModal/CardboardModal.tsx
@@ -30,7 +30,7 @@ const CardboardModal: React.FC<ICardboardModalProps> = (props) => {
     const {
         contentStackProps,
         children,
-        destructiveButtonProps,
+        dangerButtonProps,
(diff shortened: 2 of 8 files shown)


msnyder-msft:
@microsoft-github-policy-service agree company="Microsoft"

src/Components/CardboardModal/CardboardModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+    // side effects
+
+    // styles
+    const classNames = getClassNames(styles, {
+        theme: useTheme(),
+        isDestructiveFooterActionVisible: !!destructiveButtonProps
+    });
+
+    return (
+        <Modal
```
pasanchMSFT, line 61: Won't this need to be wrapped in a BaseComponent to resolve all `var` values?

src/Components/CardboardModal/CardboardModal.styles.ts
```diff
@@ -0,0 +1,98 @@
...
+    headerContainer: `${classPrefix}-headerContainer`,
+    subtitle: `${classPrefix}-subtitle`,
+    title: `${classPrefix}-title`,
+    titleContainer: `${classPrefix}-titleContainer`
+};
+export const getStyles = (
+    props: ICardboardModalStyleProps
+): ICardboardModalStyles => {
+    const { isDestructiveFooterActionVisible, theme } = props;
+    console.log('***Is destructive', isDestructiveFooterActionVisible);
```
pasanchMSFT, line 20: remove console log

pasanchMSFT (approved):
Code LGTM, I'll let you know if I notice anything when I test it out.

src/Components/CardboardModal/CardboardModal.styles.ts
msnyder-msft: Done.

src/Components/CardboardModal/CardboardModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+                                {...destructiveButtonProps}
+                                styles={classNames.subComponentStyles.destructiveButton?.()}
+                            />
+                        )}
+                        <Stack
+                            tokens={stackTokens}
+                            horizontal
+                            horizontalAlign={'end'}
+                        >
+                            <PrimaryButton
```
iremgokce, line 108: I think the preference we have is placing the primary button to the right and secondary/default button to the left

src/Components/CardboardModal/CardboardModal.types.ts
```diff
@@ -0,0 +1,65 @@
...
+    IStackStyles,
+    IStyle,
+    IStyleFunctionOrObject,
+    ITheme
+} from '@fluentui/react';
+import { CardboardIconNames } from '../../Models/Constants';
+
+export interface ICardboardModalProps {
+    /** on click of the destructive action. If not provided, button is hidden */
+    destructiveButtonProps?: IButtonProps;
```
iremgokce, line 15: what is this "destructive" means in a modal?

src/Components/CardboardModal/CardboardModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+    // contexts
+
+    // state
+
+    // hooks
+    const { t } = useTranslation();
+    const titleId = useId('modal-title');
+
+    // callbacks
```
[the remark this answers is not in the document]
msnyder-msft: Done.

src/Components/CardboardModal/CardboardModal.types.ts
iremgokce: Hmm, so this is for actions to operations like "Delete" right? I just checked it on other websites and looks like "danger" is used more to refer that button is not necessarily destroying something but a dangerous action.

src/Components/CardboardModal/CardboardModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+                        <Stack
+                            tokens={stackTokens}
+                            horizontal
+                            horizontalAlign={'end'}
+                        >
+                            <PrimaryButton
+                                {...primaryButtonProps}
+                                styles={classNames.subComponentStyles.primaryButton?.()}
+                            />
+                            <DefaultButton
```
iremgokce, line 100: should we assume the secondary/default button is always going to be "Cancel"? I think we can expose secondaryButtonProps and use it here to make it more generic modal, maybe?

src/Components/CardboardModal/CardboardModal.types.ts
msnyder-msft: Are you referring to the style for teh colors? Generally destroying things is considered dangerous when it comes to UX patterns because you can't undo it. 
Here is an example from ADO.
image.png

msnyder-msft: image

src/Components/CardboardModal/CardboardModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+                <div className={classNames.footer}>
+                    <Stack
+                        horizontal
+                        tokens={stackTokens}
+                        styles={classNames.subComponentStyles.footerStack}
+                    >
+                        {destructiveButtonProps && (
+                            <PrimaryButton
+                                {...destructiveButtonProps}
+                                styles={classNames.subComponentStyles.destructiveButton?.()}
```
iremgokce, line 92: sometimes we use "informational" link on the left corner of the footer in a modal. So should we place it in the footer? 
image

src/Components/CardboardModal/CardboardModal.tsx
msnyder-msft: Hmm. If we find we need an override we could do that, but I feel like we don't have any cases today where we do something different. 🤔

src/Components/CardboardModal/CardboardModal.tsx
msnyder-msft: I thought about that. Sure, I can add it as an alternative to the destructive action.

src/Components/CardboardModal/CardboardModal.tsx
```diff
@@ -0,0 +1,129 @@
...
+                            horizontalAlign={'end'}
+                        >
+                            <PrimaryButton
+                                {...primaryButtonProps}
+                                styles={classNames.subComponentStyles.primaryButton?.()}
+                            />
+                            <DefaultButton
+                                text={t('cancel')}
+                                onClick={onDismiss}
+                                styles={classNames.subComponentStyles.cancelButton?.()}
```
iremgokce, line 103: do consumer have to provide styling for it (seems it is optional as subcomponent style), shouldn't we have default styling for secondary actions (like cancel)?

src/Components/CardboardModal/CardboardModal.tsx
```diff
@@ -0,0 +1,121 @@
...
+import { useTranslation } from 'react-i18next';
+
+const stackTokens: IStackTokens = { childrenGap: 8 };
+
+const getClassNames = classNamesFunction<
+    ICardboardModalStyleProps,
+    ICardboardModalStyles
+>();
+
+const CardboardModal: React.FC<ICardboardModalProps> = (props) => {
```
iremgokce, line 29: Don't you think it is more like Dialog than Modal?

src/Components/CardboardModal/CardboardModal.types.ts
```diff
@@ -0,0 +1,65 @@
...
+    IIconStyles,
+    IModalStyles,
+    IStackStyles,
+    IStyle,
+    IStyleFunctionOrObject,
+    ITheme
+} from '@fluentui/react';
+import { CardboardIconNames } from '../../Models/Constants';
+
+export interface ICardboardModalProps {
```
iremgokce, line 13: do you think we should add targetRef to keep track of the element which opened that dialog? I kind of remember cases that we needed to trace back the element to focus back or something on dismiss the dialog (probably for accessibility)?

src/Components/CardboardModal/CardboardModal.tsx
msnyder-msft: hmm no. this is a modal in my book. Dialog is more like the "Are you sure" type windows we see. Modals are the bigger version that show lots of rich content.
image

image

src/Components/CardboardModal/CardboardModal.types.ts
msnyder-msft: hmmm. I could add that if you can point me to where we do it today. I haven't noticed that before.

src/Components/CardboardModal/CardboardModal.tsx
msnyder-msft: we do have default styling. it comes from the styles file. If they want to override, they have the option to, but the defautl is to have these styles here.

src/Components/CardboardModal/CardboardModal.types.ts
iremgokce: hmm nvm, looks like fluent does that by itself

src/Components/CardboardModal/CardboardModal.types.ts
msnyder-msft: oh! I didn't know this existed. Lovely!! Great find.

src/Components/CardboardModal/CardboardModal.types.ts
iremgokce: not sure why we needed it but worth to check: 
image

iremgokce (approved):
left couple of comments, very good refactoring and standardization! Code looks good to me!

src/Components/CardboardModal/CardboardModal.types.ts
msnyder-msft: oh gotcha. You can still do that though right? Cause this is handled in the onDismiss function in the caller? It's a ref to something outside the modal.

src/Components/CardboardModal/CardboardModal.tsx
msnyder-msft: Made the danger button non-optional since we always supply that.

src/Components/CardboardModal/CardboardModal.tsx
msnyder-msft: Added.

src/Components/CardboardModal/CardboardModal.types.ts
msnyder-msft: Renamed.

src/Components/CardboardModal/CardboardModal.tsx
msnyder-msft: Updated.

msnyder-msft:
@microsoft-github-policy-service agree company="Microsoft"

src/Components/CardboardModal/CardboardModal.types.ts
iremgokce: yeah, that is true! :)

msnyder-msft:
> @microsoft-github-policy-service agree company="Microsoft"

@microsoft-github-policy-service agree company="Microsoft"

PR #709, opened 2022-10-04
Fetching and Storing ADX Connection information of the Selected ADT Instance

iremgokce:
### Summary of changes 🔍 
- Added logic to retrieve the ADX connection information of the selected ADT instance in the app on load of ADT3DScenePage using the resource id -if exist- or url of the selected ADT instance from deeplink. We store this information in the ADT3DScenePage state which exposes variables through `ADT3DScenePageContext` for other components to access.
   - If resource id exists; **(1) fetch the resource** using its id to get the "location" information of the resource to construct the adx table name, **(2) fetch the available timeseriesDatabaseConnections for this resource**. So we make only two calls.
image

   - If resource id is not present, pass the url of the ADT instance and; **(1) fetch all the digital twin instance resources and find the one that matches that url**, **(2) fetch the available timeseriesDatabaseConnections for this resource**. So we might end up making thousands of calls. _The worst case is when user comes to ADT3DScene page and if there is no resource id, we start this series of calls and user also go to the EnvironmentPicker which also makes the same series of calls to list the ADT instances in the dropdown._
- Revised some state management for `ADT3DScenePage` with strong typings

This way, children components like `WidgetLibraryDialog` can show the loading state while creating a "Data History Widget" and Data History Widget authoring form can access the connection information to be used default connectionString:
image

### Testing 🧪
- Go to ADT3DScenePage and in the Network tab see the calls made to the AzureManagement Rest API to (1) fetch the resource using either resource id if present or url of the selected ADT instance and (2) fetching timeseriesDatabaseConnections of that instance

### Checklist ✔️
- [x] Added relevant labels
- [ x] Requested developer reviews


src/Models/Classes/AdapterDataClasses/ADTInstanceTimeSeriesConnectionData.ts
```diff
@@ -0,0 +1,15 @@
+import { IAdapterData, IADXConnection } from '../../Constants/Interfaces';
+
+class ADTInstanceTimeSeriesConnectionData implements IAdapterData {
```
msnyder-msft, line 3: nit: does it need to be TimeSeries specific? Could we call it `DataHistory` instead?

src/Models/Constants/Interfaces.ts
```diff
@@ -312,6 +320,14 @@ export interface IAzureStorageAccountPropertyData {
     primaryEndpoints: { blob: string; [additionalProperty: string]: any };
     [additionalProperty: string]: any;
 }
+
+export interface IAzureTimeSeriesDatabaseConnectionPropertyData {
+    connectionType: 'AzureDataExplorer' | string;
+    adxEndpointUri: string; // cluster url
```
msnyder-msft, line 326: nit: can we move the comment to the /** syntax on the line above so it shows on the intellisense?

src/Pages/ADT3DScenePage/ADT3DScenePage.types.ts
```diff
@@ -56,21 +57,84 @@ export interface ISceneContentsProps {
...
+    SET_ADT_SCENE_CONFIG = 'SET_ADT_SCENE_CONFIG',
+    SET_CURRENT_STEP = 'SET_CURRENT_STEP',
+    SET_SELECTED_SCENE = 'SET_SELECTED_SCENE',
+    SET_ERRORS = 'SET_ERRORS',
+    SET_ERROR_CALLBACK = 'SET_ERROR_CALLBACK',
+    SET_ADX_CONNECTION_INFORMATION_LOADING_STATE = 'SET_ADX_CONNECTION_INFORMATION_LOADING_STATE',
+    SET_ADX_CONNECTION_INFORMATION = 'SET_ADX_CONNECTION_INFORMATION'
+}
+
+export type ADT3DScenePageAction =
```
msnyder-msft, line 98: yessssssssss. I have wanted to do this forever.

src/Pages/ADT3DScenePage/ADT3DScenePage.types.ts
```diff
@@ -56,21 +57,84 @@ export interface ISceneContentsProps {
...
     handleOnSceneClick: (scene: IScene) => void;
     handleOnSceneSwap: (sceneId: string) => void;
     isTwinPropertyInspectorPatchModeEnabled: boolean;
 }
+
+export enum ADXConnectionInformationLoadingState {
+    IDLE,
+    LOADING,
+    EXISTED,
+    NOT_EXISTED
```
msnyder-msft, line 85: nit: NOT_EXIST

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
```diff
@@ -231,6 +238,34 @@ const ADT3DScenePageBase: React.FC<IADT3DScenePageProps> = ({
...
     );
 
+    // if the adx connection information is not in the environment context. fetch it and update the context
+    useEffect(() => {
+        if (!adapter.getADXConnectionInformation()) {
+            {
+                connectionState.callAdapter({
+                    adtInstanceIdentifier: deeplinkState.adtResourceId
+                        ? { id: deeplinkState.adtResourceId }
+                        : { hostName: deeplinkState.adtUrl?.slice(7) }
```
msnyder-msft, line 248: how can we avoid copying this "slice(7)" logic all around? can we pull it off to a util function at least with a name? I saw it before in your last PR and this is seems very similar how we got into trouble with the adding and removing https in the environment picker before. 🤔

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
```diff
@@ -231,6 +238,34 @@ const ADT3DScenePageBase: React.FC<IADT3DScenePageProps> = ({
...
+                });
+                dispatch({
```
msnyder-msft, line 250: it's usually a code smell when you need 2 dispatch operations in a row. Is there a way we can combine them into a single action so that it's more of an "action" rather than just a buch of "setState" calls?

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
```diff
@@ -477,6 +531,43 @@ const ADT3DScenePageBase: React.FC<IADT3DScenePageProps> = ({
         }
     }, [setCorsPropertiesAdapterData?.adapterResult]);
 
+    // update the adx information of environment context with the fetched data
+    useEffect(() => {
+        if (!connectionState?.adapterResult.hasNoData()) {
+            const connectionData = connectionState.adapterResult.getData();
+
+            dispatch({
```
msnyder-msft, line 539: same comment here, can we combine these actions into real "actions" that update multiple state variables in one shot? It seems reasonable that if you are setting the connection information, you would always want to set the status at the same time as well so maybe they can just be a single "StoreConnectionInformation" action.

src/Adapters/MockAdapter.ts
```diff
@@ -1005,4 +1018,27 @@ export default class MockAdapter
...
+    async getTimeSeriesConnectionInformation(_adtInstanceIdentifier: {
+        id?: string;
+        hostName?: string;
+    }) {
+        try {
+            await this.mockNetwork();
+
+            return new AdapterResult({
+                result: new ADTInstanceTimeSeriesConnectionData({
+                    kustoClusterUrl: '',
```
msnyder-msft, line 1031: can we return some actual strings just so we can see them flowing through the system when we show them in the UI or write them to the config?

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -567,4 +571,94 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
             return new AzureResourcesData([newRoleAssignment]);
         }, 'azureManagement');
     }
+
+    async getTimeSeriesConnectionInformation(adtInstanceIdentifier: {
+        id?: string;
```
msnyder-msft, line 576: is it possible to make this type a bit more strict? At least one of these properties is required but it looks like right now both are optional.

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -567,4 +571,94 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
             return new AzureResourcesData([newRoleAssignment]);
         }, 'azureManagement');
     }
+
+    async getTimeSeriesConnectionInformation(adtInstanceIdentifier: {
```
msnyder-msft, line 575: can we add a comment about the priority of the args? It seems like they are mutually exclusive, if both are provided, which is used?

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -567,4 +571,94 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
...
+                        d.properties.hostName === adtInstanceIdentifier.hostName
+                );
+            } else if (adtInstanceIdentifier.id) {
+                const digitalTwinInstance = await this.getResourceById(
+                    adtInstanceIdentifier.id
+                );
+                adtInstance = digitalTwinInstance.result.data;
+            }
+
+            try {
```
msnyder-msft, line 605: do we need to confirm that we got back an ADT Instance before proceeding?

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -567,4 +571,94 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
...
+            } catch (error) {
+                adapterMethodSandbox.pushError({
+                    isCatastrophic: false,
+                    rawError: error
+                });
+            }
+        }, 'azureManagement');
+    }
+
+    async getResourceById(resourceId: string) {
```
msnyder-msft, line 642: don't we already have this API? Do we need antoher copy? Or can we not access other adapters from here?

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -567,4 +571,94 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
...
+                    url,
+                    {
+                        apiVersion:
+                            AzureResourcesAPIVersions[
+                                'Microsoft.DigitalTwins/digitalTwinsInstances/timeSeriesDatabaseConnections'
+                            ]
+                    }
+                );
+
+                const connectionData = timeSeriesDatabaseConnections[0]; // TODO: for now get the first connection information
```
msnyder-msft, line 620: do we need to assert that we got at least 1 connection back?

msnyder-msft (approved):
Looking really good! I didn't run the code but love the refactor to add strong typing to the reducer.

src/Models/Classes/AdapterDataClasses/ADTInstanceTimeSeriesConnectionData.ts
iremgokce: Since the name of the method in Azure Management Rest api is /timeSeriesDatabaseConnections, I wanted to keep it aligned with it.

src/Models/Constants/Interfaces.ts
iremgokce: ahh I will learn this one day... :) fixing it

src/Adapters/AzureManagementAdapter.ts
iremgokce: added strict type, yeah they are mutually exclusive, only one of them should be provided

src/Adapters/AzureManagementAdapter.ts
iremgokce: added strict type, yeah they are mutually exclusive, only one of them should be provided

src/Adapters/AzureManagementAdapter.ts
iremgokce: we didn't have `getResourceById` method in the AzureManagementAdapter before? We have been using getResources type of methods before. There is no other copy?

src/Adapters/AzureManagementAdapter.ts
iremgokce: nope, since it is in try catch block, if instance not exist it will fall into catch and adapter will resolve null (just added that).

src/Adapters/AzureManagementAdapter.ts
iremgokce: good point, no since it is wrapped in try catch, if it fails it will fall into catch block which returns null

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
iremgokce: added utils method for those

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
```diff
@@ -231,6 +238,34 @@ const ADT3DScenePageBase: React.FC<IADT3DScenePageProps> = ({
         [deeplinkDispatch, environmentPickerOptions?.adt]
     );
 
+    // if the adx connection information is not in the environment context. fetch it and update the context
+    useEffect(() => {
+        if (!adapter.getADXConnectionInformation()) {
+            {
+                connectionState.callAdapter({
```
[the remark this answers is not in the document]
iremgokce: nope

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
```diff
@@ -231,6 +238,34 @@ const ADT3DScenePageBase: React.FC<IADT3DScenePageProps> = ({
         [deeplinkDispatch, environmentPickerOptions?.adt]
     );
 
+    // if the adx connection information is not in the environment context. fetch it and update the context
+    useEffect(() => {
+        if (!adapter.getADXConnectionInformation()) {
```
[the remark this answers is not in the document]
iremgokce: added cache logic for connection information

darsney (approved):

PR #711, opened 2022-10-05
Fixed wrong cached adx connection value returning for the selected adt instance

iremgokce:
### Summary of changes 🔍 
- Changed the cached value key in the adapter to map to the current ADT instance passed in parameter
- Increased the refresh rate for pulling adx connection information to 1 hour (thinking it won't be changed as frequent)

### Testing 🧪
Go to ADT3DScene page, after the first adx connection calls, change your environment in the EnvironmentPicker and see next adx connection calls are triggered (it used to return the cached value due to the same key)

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Adapters/ADTAdapter.ts
+++ b/src/Adapters/ADTAdapter.ts
@@ -51,6 +51,7 @@ import ADTTwinLookupData from '../Models/Classes/AdapterDataClasses/ADTTwinLooku
 import axios, { AxiosError, AxiosInstance } from 'axios';
 import { DtdlInterface } from '../Models/Constants/dtdlInterfaces';
 import {
+    buildCacheKey,
     getDebugLogger,
     getModelContentType,
     parseDTDLModelsAsync
@@ -199,7 +200,7 @@ export default class ADTAdapter implements IADTAdapter {
             );
         if (useCache) {
             return this.adtTwinCache.getEntity(
-                twinId,
+                buildCacheKey('adt_twin', twinId),
                 getDataMethod,
                 forceRefresh
             );
@@ -234,7 +235,10 @@ export default class ADTAdapter implements IADTAdapter {
                 }
             });
 
-        return this.adtTwinToModelMappingCache.getEntity(twinId, getDataMethod);
+        return this.adtTwinToModelMappingCache.getEntity(
+            buildCacheKey('adt_model_id', twinId),
+            getDataMethod
+        );
     }
 
     async getAllAdtModels() {
@@ -292,7 +296,10 @@ export default class ADTAdapter implements IADTAdapter {
                     });
                 }
             });
-        return this.adtModelsCache.getEntity('adt_models', getDataMethod);
+        return this.adtModelsCache.getEntity(
+            buildCacheKey('adt_models', this.adtHostUrl),
+            getDataMethod
+        );
     }
 
     getADTModel(modelId: string) {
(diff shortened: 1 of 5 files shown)


src/Adapters/AzureManagementAdapter.ts
```diff
@@ -685,7 +685,10 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
 
         if (useCache) {
             return this.timeSeriesConnectionCache.getEntity(
-                'adt_adx_connection_information',
+                `${
```
msnyder-msft, line 688: should we have a "BuildCacheKey" function so we use the same logci in the set as we do in the get?

msnyder-msft (approved):

src/Adapters/AzureManagementAdapter.ts
iremgokce: is this something you would think of @msnyder-msft ? https://github.com/microsoft/iot-cardboard-js/pull/711/commits/30c57457f1884c3f98989b71400cdec3354f8c27

src/Adapters/ADTAdapter.ts
```diff
@@ -199,7 +200,7 @@ export default class ADTAdapter implements IADTAdapter {
             );
         if (useCache) {
             return this.adtTwinCache.getEntity(
-                twinId,
+                buildCacheKey('adt_twin', twinId),
```
msnyder-msft, line 203: why are we changing the cache keys for twins and other entities?

src/Models/Classes/AdapterEntityCache.ts
```diff
@@ -47,12 +47,8 @@ class AdapterEntityCache<T extends IAdapterData> {
         }
     }
 
-    setEntity(
-        key: string,
-        getEntityData: () => Promise<AdapterResult<T>>
-    ): void {
-        const newCachedEntity = new CachedEntity(getEntityData);
```
msnyder-msft, line 54: this seems significant. Why do we change this from the ability to delay populating the data to forcing a resolved value? I thought this was being used already.

src/Adapters/ADTAdapter.ts
iremgokce: to make the keys more explicit. do you have an idea in your mind for buildCacheKey method? I dont see much contribution of adding that buildCacheKey method otherwise

msnyder-msft: I see. I was expecting the "buildCacheKey" to build this key.
 `adtInstanceIdentifier['hostName'] || adtInstanceIdentifier['id']`

That way we can use it when we add an item to the cache as well as reading it. BUT now I notice that we never actually call Set explicitly. We call get and pass a way to fetch the value if it's a cache miss. In that case I think your original approach is better.

src/Models/Classes/AdapterEntityCache.ts
iremgokce: `setEntity` was not being used anywhere, I revised it that way to use it in the `getEntity`. I can revert this change, but again  setEntity is not called anywhere

src/Adapters/ADTAdapter.ts
msnyder-msft: We don't really need the discriminator like `adt_twin` since every cache is separate so I think your original plan makes the most sense. Sorry for the confusion.

iremgokce: no worries, reverting changes about key then

pasanchMSFT (approved):
LGTM

PR #716, opened 2022-10-10
Data history widget builder (only authoring experience, not including the view model)

iremgokce:
### Summary of changes 🔍 
- Schema changes for Data History Widget
- Components necessary for authoring a Data History Widget
- Some revisions in existing code ( e.g. to get resource id if not exist in "selected" value but in options value in local storage to improve performance, fixed suffix checks in proxy for security)

**Notes**
- Requires https://github.com/microsoft/iot-cardboard-js/pull/711 to be completed first
- **[!] For now, enabled Data History Widget by default for Design/PM review using Chromatic link with Mock data, don't merge until that flag is disabled back again**.

**ToDos with the following PR:**
1. DataHistoryWidget.tsx component as view model, currently it is a placeholder showing the values in the form to reflect the changes (consuming HighCharts library with a wrapper component)
2. Adding query builder by transforming the expression and units into an ADX query to fetch timeseries data from ADX rest api
3. HighCharts in view mode
4. Update the doc link for Learn more when Data Widget is not enabled with go.microsoft link (@msnyder-msft?)

_ToDos as a separate PR:_
Cancelling all the requests when EnvironmentPicker is unmounted to improve performance in browser. Because, on page load we try to fetch the ADX connection information using resource id of the selected ADT instance which is the same flow to get the resources in EnvironmentPicker. So, worst scenario is, when user land on 3DScene page, if resource id is not present, the flow is kicked and if the user also click on EnvironmentPicker the same series of requests are also kicked off, which may slow down the app a lot.

### Testing 🧪
1. First, enable data history widget in `availableWidgets` in Constants.ts by setting the `disabled: false`. This is already done for design/PM review using Chromatic link. Need to disable again before this PR is merged! Or you can set a local storage flag item to enable it using `localStorage.setItem('cardboard.feature.dataHistoryWidget', true)`
5. You need a data history enabled ADT instance, I will share one that Matt created a while ago in the dev chat, or you can enable it for your ADT instance following the steps [here](https://learn.microsoft.com/en-us/azure/digital-twins/concepts-data-history#creating-a-data-history-connection).
6. Run the app and open the EnvironmentPicker first time and let all the azure resources load, select an environment with data history enabled and then hit save to preserve the resource id of the selected ADT instance, if you have not already. If you don't do that, each time on page load the flow of requests are being made to get the resource id of the selected ADT instance based on adtUrl - which you can also test.
7. Then go to either in Mock 3D Page or ADT3DScenePage story, go to a behavior, Widgets tab and "Add widget", observe Data history widget there and test by adding one. You should have a proper elements to match with proper primary twin ids in the selected ADT instance. So, you may need to create a new element and update your behavior accordingly. 

image
image
image
image
image
image
image

### Checklist ✔️
- [x] Added relevant labels


src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesFormCallout.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+    const handleLabelChange = useCallback((_event, label: string) => {
+        setSeriesToEdit(
+            produce((draft) => {
+                draft.label = label;
+            })
+        );
+    }, []);
+
+    const isFormValid = useMemo(() => {
```
darsney, line 110: wondering if we should actually enforce that unit is always present, or just make unit and label both optional.  my preference is that unit and label are both optional personally

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesFormCallout.tsx
```diff
@@ -0,0 +1,173 @@
+import React, {
```
darsney, line 1: When there are 3 series, editing one causes it to be anchored at the top of the form, not on the thing you clicked on.  a bit confusing IMO

darsney (commented):
Looks pretty good to me experience wise!  Couple small nits, i defer to the rest of the team for comments on code right now

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesFormCallout.tsx
iremgokce: yeah when thinking about more, unit may not be enforced. I will make it optional

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesFormCallout.tsx
iremgokce: couldn't reproduce it, can you send me screen recording? When there are 3 series, editing one didn't cause it to be anchored at the top of the form?
3TimeSeries_onedit

msnyder-msft: that anchor doesn't seem to be right even in your recording here. It should be next tot he item you clicked and not at the top of the screen right?

iremgokce: edit on overflow menu and onclick of list item tied to the same callback. not sure which anchor are you both referring to?

msnyder-msft: The callout should open to the right of the list item.

iremgokce: oh okay you mention the callout target, yeah it is set to the "Add time series" action button (callout appears at the bottom of that action button target). but since when there are 3 series I am hiding that action button that is why it was anchoring to the top then. let me see if I can improve it.
image

iremgokce: edit callout is anchored to the bottom of the list item

src/Resources/Locales/en.json
```diff
@@ -583,6 +583,61 @@
...
+                    "yAxisType": {
+                        "description": "Chart option to set if each series has separate (Independent) y-axes or sharing the same y-axis (Shared)",
+                        "label": "Y axis type",
+                        "sharedLabel": "Shared",
+                        "independentLabel": "Independent"
+                    },
+                    "quickTimeSpan": {
+                        "label": "Default time span",
+                        "options": {
+                            "Last 15 mins": "Last 15 mins",
```
msnyder-msft, line 620: where do we use these loc versions? I thoguht I saw they were hardcoded in an enum

msnyder-msft (changes requested):

msnyder-msft (changes requested):
Couple thoughts on the schema.
Let's add some stories.
Couple nits on styling 
Biggest one is probably the time horizon options need to be localized instead of having an enum. Left a reference to how I did this in the past.

src/Resources/Locales/en.json
iremgokce: Because we need to localize the quick span options. The quick span options itself is an enum type and this enum is used to generate a mapping for millisecond number value for each enum key.

msnyder-msft:
Spacing feels off between the list and "add" button
image

src/Resources/Locales/en.json
```diff
@@ -583,6 +583,61 @@
...
+                    "unit": "Unit",
+                    "label": "Label",
+                    "labelPlaceholder": "Enter label for this series to show in chart",
+                    "add": "Add time series",
+                    "edit": "Edit time series",
+                    "remove": "Remove time series"
+                },
+                "chartOptions": {
+                    "yAxisType": {
+                        "description": "Chart option to set if each series has separate (Independent) y-axes or sharing the same y-axis (Shared)",
```
msnyder-msft, line 612: spelling: y-axis

src/Resources/Locales/en.json
msnyder-msft: Doesn't look like they are getting localized.
image

msnyder-msft: Same in aggregation type.
image

src/Resources/Locales/en.json
```diff
@@ -583,6 +583,61 @@
...
+                "noConnectionInformationText": "No connection information found",
+                "displayNamePlaceholder": "Enter display name",
+                "timeSeries": {
+                    "timeSeriesDescription": "Add up to {{maxTimeSeriesCount}} timeseries to display",
+                    "title": "Time series",
+                    "description": "A series to display in the line chart",
+                    "property": "Property",
+                    "unit": "Unit",
+                    "label": "Label",
+                    "labelPlaceholder": "Enter label for this series to show in chart",
```
msnyder-msft, line 605: This value is too long to show. We should shorten it so it is visible in the field.
Maybe "Enter label to show in chart legend"
image

src/Models/Services/LocalStorageManager/LocalStorageManager.ts
```diff
@@ -53,6 +53,33 @@ export const getEnvironmentItemFromResource = (
...
+ * @param item environment item in local storage
+ * @param type intended type for the Azure resource
+ */
+export const getResourceFromEnvironmentItem = (
+    item: EnvironmentItemInLocalStorage,
+    type: AzureResourceTypes
+): IAzureResource => {
+    if (!item) return null;
+    return {
+        id: item.id ?? null,
```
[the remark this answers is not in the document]
msnyder-msft: gotchaaa. Ok. Sure. We usually dont' differentiate between null and undefined in our code base. Do we have a specific dependency here?

src/Models/Classes/3DVConfig.ts
```diff
@@ -34,12 +37,15 @@ export enum WidgetType {
     Link = 'Link',
     Value = 'Value',
     Trend = 'Trend',
-    Panel = 'Panel'
+    Panel = 'Panel',
+    DataHistory = 'Data history'
```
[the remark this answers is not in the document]
iremgokce: this enum is used in builder components which is mapping to the "type" string of widgets in schema. There has been no enum for type of widget generated from schema. We need this enum since the value of the keys are used as labels on UI (e.g. the title of the widgets in WidgetDialog are the values of this WidgetType enum)

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
```diff
@@ -236,6 +236,62 @@
...
+            "required": ["expression"]
+        },
+        "IDataHistoryTimeSeries": {
+            "type": "array",
+            "description": "A list of timeseries to render in the chart",
+            "items": {
+                "$ref": "#/$defs/IDataHistoryBasicTimeSeries"
+            }
+        },
+        "IDataHistoryConnectionString": {
```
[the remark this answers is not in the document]
iremgokce: the intention behind was if we need to change the the syntax of the connection string, I wanted to keep it as a separate type. But since it is just string, we don't need it yes.

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
```diff
@@ -236,6 +236,62 @@
...
+            "type": "string",
+            "description": "Timeseries database connection string in key1=value1;key2=value2;key3=value3 format that is used for the connection of a data history widget"
+        },
+        "IDataHistoryChartYAxisType": {
+            "type": "string",
+            "enum": ["Shared", "Independent"]
```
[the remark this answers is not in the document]
iremgokce: since I use the generated types directly in the UI as dropdown options I wanted to keep them like this: 
image
image

I can do the casing manipulation on the UI side though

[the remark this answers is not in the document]
iremgokce: nevermind, looks like I was already using translation for it. Okay making the type all lowercase

src/Resources/Locales/en.json
iremgokce: it was intentional referring plural axes. is it wrong?

iremgokce:
> Spacing feels off between the list and "add" button. SHouldn't the spacing on top and bottom be the same? image

I think it was intentional which is aligning with the design since it is a separate action button which should be differentiated the "Time series" list block. @azimamansuri any feedback?

src/Resources/Locales/en.json
iremgokce: since it is not just legend, I just changed it to "Enter label to show in chart"

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
```diff
@@ -465,6 +544,29 @@
             },
             "required": ["type", "id", "widgetConfiguration"]
         },
+        "IDataHistoryWidget": {
+            "type": "object",
+            "description": "A data history widget which uses twin properties to show timeseries data",
+            "additionalProperties": false,
+            "properties": {
+                "type": {
+                    "const": "Data history"
```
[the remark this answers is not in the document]
iremgokce: clearly yes :) the generated value for type prop is a const string, which works
image

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
```diff
@@ -0,0 +1,360 @@
...
+    }, [formData]);
+
+    useEffect(() => {
+        if (
+            adxConnectionInformation.loadingState ===
+            ADXConnectionInformationLoadingState.EXIST
+        ) {
+            const connection = adxConnectionInformation.connection;
+            updateWidgetData(
+                produce(formData, (draft) => {
```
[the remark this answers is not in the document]
iremgokce: can you see the change I made here to pull this out to a method and checking individual fields in the connection object? Only thing I can think of is adding additional check for kustoClusterUrl which needs to ends with "**.kusto.windows.net**". Do you think I should add this to this new method when creating the connection string @msnyder-msft ? (But of course we will let arbitrary connections in the future)
https://github.com/microsoft/iot-cardboard-js/pull/716/commits/5f70e73a7f444b531b8b79ff173b5d313b72c2f9#diff-0472d5c3363025938a9935310b39167e83a00ed260c18efd58d965a621b73f87R352

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.types.ts
```diff
@@ -0,0 +1,86 @@
...
+    IDataHistoryChartOptions
+} from '../../../../../../../Models/Types/Generated/3DScenesConfiguration-v1.0.0';
+
+export const MAX_NUMBER_OF_TIME_SERIES = 3;
+export const SERIES_LIST_ITEM_ID_PREFIX = 'cb-data-history-widget-series';
+
+/** Existing chart options for data history widget in the 3D config schema */
+export type ChartOptionKeys = keyof IDataHistoryChartOptions;
+
+export enum QuickTimeSpanKey {
```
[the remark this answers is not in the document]
iremgokce: They are already being localized? Please see: https://github.com/microsoft/iot-cardboard-js/pull/716/files#diff-a59240bd3c23fdf5d744ce2d4ae1cfbb0d497b66292f649c6b1f78616cf85d29R68

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
```diff
@@ -0,0 +1,360 @@
...
+            return i18next.t(
+                'widgets.dataHistory.form.noConnectionInformationText'
+            );
+        }
+    }, [
+        formData.widgetConfiguration.connectionString,
+        adxConnectionInformation.loadingState
+    ]);
+
+    const quickTimeSpanKeyByValue = useMemo((): QuickTimeSpanKey => {
```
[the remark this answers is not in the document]
iremgokce: They are already being localized? Please see: https://github.com/microsoft/iot-cardboard-js/pull/716/files#diff-a59240bd3c23fdf5d744ce2d4ae1cfbb0d497b66292f649c6b1f78616cf85d29R68

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
```diff
@@ -0,0 +1,360 @@
...
+        [
+            updateWidgetData,
+            handleTimeSeriesFormDismiss,
+            formData,
+            selectedSeries
+        ]
+    );
+
+    const handleTimeSeriesEditClick = useCallback((idx: number) => {
+        setSelectedTimeSeriesIdx(idx);
```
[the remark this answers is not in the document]
iremgokce: there is no id I defined for time series in data history widget. I can change the schema

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.types.ts
```diff
@@ -310,3 +311,9 @@ export interface IValueWidgetBuilderProps extends IWidgetBuilderFormDataProps {
     formData: IValueWidget;
     updateWidgetData: (widgetData: IValueWidget) => void;
 }
+
+export interface IDataHistoryWidgetBuilderProps
```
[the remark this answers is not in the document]
iremgokce: this was the place we used to put for all widget types, okay will update:
image

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetLibraryDialog/WidgetLibraryDialog.tsx
```diff
@@ -51,6 +60,107 @@ const WidgetLibraryDialog: React.FC<{
...
+    );
+
+    const handleOnRenderCell = useCallback(
+        (widget: IWidgetLibraryItem, index: number) => (
+            <DefaultButton
+                disabled={
+                    widget.data.type === WidgetType.DataHistory &&
+                    scenePageContext?.state.adxConnectionInformation
+                        ?.loadingState !==
+                        ADXConnectionInformationLoadingState.EXIST
```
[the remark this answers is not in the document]
iremgokce: isWidgetNotAvailable - now isWidgetAvailable determines if the widget creation option is all together not possible. In this case, the button will be disabled along with loading spinner until we fetch the connection data. I can improve the logic though

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetLibraryDialog/WidgetLibraryDialog.tsx
```diff
@@ -51,6 +60,107 @@ const WidgetLibraryDialog: React.FC<{
...
+                    flexContainer: { justifyContent: 'start' }
+                }}
+                selected={index === selectedWidget}
+            >
+                <Stack horizontal>
+                    <Stack
+                        className="cb-widget-dialog-icon-container"
+                        aria-hidden={true}
+                    >
+                        {isSpinnerVisible(widget.data.type as WidgetType) ? (
```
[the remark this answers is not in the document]
iremgokce: it happens because in schema file there is no separate type defined for widget type, but it is list of strings. But on the components side we use a strict typing for it which is defined in 3DVConfig.ts (which we use for additional typings on top of schema generated typings). That is why I need to do this casting. I can update the schema to generate a separate type and remove the one we use in 3DVConfig.ts:
image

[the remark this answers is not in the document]
iremgokce: in fact I don't know how to define a string const using an enum in json schema. I think that is why we used a const string for "type" of widgets and define an enum based on it in 3DVConfig.ts. So, I will leave this as is for now.

src/Resources/Locales/en.json
msnyder-msft: we don't usually say where it's used, I'd almost think it's better to just go with "Enter series label". I don't feel strongly so feel free to resolve if you disagree.

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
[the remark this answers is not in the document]
msnyder-msft: ok. I wasn't sure if you'd tried to use it yet. I'd personally rather not have a space in a type name. That feels weird. 🤷

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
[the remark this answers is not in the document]
msnyder-msft: yeah. but taht doesn't really validate anything from a security perspective. I remember we talked in the design review about the Hullabaloo vulnerability that was exploited recently through connection strings. I guess in this case we are only creating the connection string and not actually using it. So perhaps this comment needs to actually be on the data execution side of the code rather than the construction.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
```diff
@@ -0,0 +1,360 @@
...
+                {formData.widgetConfiguration.timeSeries.length <
+                    MAX_NUMBER_OF_TIME_SERIES && (
+                    <ActionButton
+                        id={addTimeSeriesCalloutId}
+                        styles={actionButtonStyles}
+                        text={t('widgets.dataHistory.form.timeSeries.add')}
+                        onClick={toggleIsAddTimeSeriesCalloutVisible}
+                    />
+                )}
+                <Label className={customStyles.label} id={yAxisLabelId}>
```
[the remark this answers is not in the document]
iremgokce: I need to render a tooltip for this label but since there is no "onRenderLabel" prop for `ChoiceGroup `component in fluent this is the suggested customization method.

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.tsx
```diff
@@ -0,0 +1,47 @@
+import React, { useContext } from 'react';
+import { IDataHistoryWidget } from '../../../../../Models/Types/Generated/3DScenesConfiguration-v1.0.0';
+import { BehaviorsModalContext } from '../../../BehaviorsModal';
+
+interface IProp {
+    widget: IDataHistoryWidget;
+}
+
+export const DataHistoryWidget: React.FC<IProp> = ({ widget }) => {
```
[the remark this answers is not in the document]
iremgokce: :) this is the view model component for Data History Widget and it is currently a placeholder as I explained in the description of my PR. I will render the HighCharts here in the following PR.

src/Components/CardboardListCallout/CardboardListCallout.styles.ts
```diff
@@ -64,6 +64,13 @@ export const getCardboardListCalloutComponentStyles = memoizeFunction(
                     marginTop: 8,
                     opacity: 0.6
                 }
+            ],
+            primaryButton: [
+                classNames.primaryButton,
+                {
+                    marginTop: '16px !important',
```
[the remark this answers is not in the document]
iremgokce: I just moved the content here and made sure this margin top style is in effect (it was not), it used to be a separate method to get the styles of primaryButton, please see the other change in the same file

src/Models/Services/LocalStorageManager/LocalStorageManager.ts
```diff
@@ -53,6 +53,33 @@ export const getEnvironmentItemFromResource = (
     };
 };
 
+/** This is used for converting an environment configuration item to Azure resource
+ * @param item environment item in local storage
+ * @param type intended type for the Azure resource
+ */
+export const getResourceFromEnvironmentItem = (
```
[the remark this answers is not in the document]
iremgokce: in fact it is local storage related. the "environmentItem" is the item that we keep in local storage. Please see the method  `getEnvironmentItemFromResource` above, this method is the other way around by getting a hardcoded resource from environment item in local storage. Will keep it here.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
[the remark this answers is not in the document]
iremgokce: okay added this check ^

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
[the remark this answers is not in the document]
iremgokce: by the way you will need to reset your config file in your container if you already created a data history widget using this scehema. otherwise it will throw error saying that parsing failed.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.types.ts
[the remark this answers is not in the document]
iremgokce: off I blame i18next, fixed the translation: 
image

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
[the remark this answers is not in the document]
iremgokce: off I blame i18next, fixed the translation: 
image

src/Resources/Locales/en.json
iremgokce: off I blame i18next, fixed the translation: 
image

iremgokce: by the way aggregation type is **NOT** going to be localized since it is a numeric operations on data which is universal. This is what we also did in TSI app.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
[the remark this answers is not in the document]
msnyder-msft: how odd. That must be one of the few that doesn't let you do that. :(

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
```diff
@@ -183,6 +176,8 @@ const DataHistoryWidgetBuilder: React.FC<IDataHistoryWidgetBuilderProps> = ({
         [formData]
     );
 
+    const theme = useTheme();
+    const customStyles = getWidgetFormStyles(theme);
```
msnyder-msft, line 180: can we use our styled pattern instead of this older way of doing styles?

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
```diff
@@ -79,12 +79,14 @@ export const DeeplinkContextReducer: (
                                 AzureResourceDisplayFields.url
                             )
                     );
-                    draft.adtResourceId = itemInLocalStorage.id;
-                    const resourceFromItem = getResourceFromEnvironmentItem(
-                        itemInLocalStorage,
-                        AzureResourceTypes.DigitalTwinInstance
-                    ) as IADTInstance;
-                    setSelectedAdtInstanceInLocalStorage(resourceFromItem);
+                    if (itemInLocalStorage) {
```
msnyder-msft, line 82: This makes us null safe, but will the app respond effectively if we don't find the value in storage?

src/Resources/Locales/en.json
msnyder-msft: i guess you're right. I have never seen it in plural form before. lol I guess it doesn't really come up much.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesFormCallout.tsx
msnyder-msft: looks great now

msnyder-msft (approved):
Looks really good now. Let's add some stories for your form and then send it.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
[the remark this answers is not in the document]
iremgokce: added id for time series items and changed logic in authoring experience to use id as identifier instead of index

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
```diff
@@ -0,0 +1,360 @@
...
+    QuickTimeSpanKey,
+    QuickTimeSpanOptions,
+    QuickTimeSpans,
+    SERIES_LIST_ITEM_ID_PREFIX,
+    YAxisTypeOptions
+} from './DataHistoryWidgetBuilder.types';
+import TimeSeriesFormCallout from './Internal/TimeSeriesFormCallout';
+import TimeSeriesList from './Internal/TimeSeriesList';
+
+const DataHistoryWidgetBuilder: React.FC<IDataHistoryWidgetBuilderProps> = ({
```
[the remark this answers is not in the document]
iremgokce: added stories for data history widget (create/edit)

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesList.tsx
```diff
@@ -0,0 +1,117 @@
...
+    SERIES_LIST_ITEM_ID_PREFIX
+} from '../DataHistoryWidgetBuilder.types';
+
+interface IProp {
+    series: IDataHistoryTimeSeries;
+    onSeriesEditClick: (idx: number) => void;
+    onSeriesRemoveClick: (idx: number) => void;
+}
+
+const TimeSeriesList: React.FC<IProp> = ({
```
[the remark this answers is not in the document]
iremgokce: added a story for data history widget to edit a time series item from list

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
iremgokce: yep, it will just try to fetch the id (the long flow)

darsney (approved):
LGTM, keep it rollin!

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
iremgokce: changing it now

iremgokce: hey @msnyder-msft after checking this, all these styles are coming from "shared" stylings of WidgetForm which is shared by all the forms in widget builder components (e.g. widgetFormContents from WidgetForm.styles). I tried to make it separate for only one classname which seems to me a component specific but I don't see much contribution of it: https://github.com/microsoft/iot-cardboard-js/pull/716/commits/edb3e45e1656bd5af5dba4b6809075ffd171aee1 Is this what you want to see here? If not, do we have an example to consume shared stylings in a component which using the new styling somewhere?

msnyder-msft: nah. If it's shared they are kinda all bolted together. I hadn't noticed that it was shared, I saw `CustomStyles` and assumed it was specific to this component. It's fine to just share the old way if that's waht we have going in these forms. Eventually we can transition all of them at the same time.
I like what you did in that commit, teh only change I would make is to use the `styles` prop instead of classNames. To do that though, you'd need to use `subComponentStyles` and have a property that is of type `IStackStyles`.

src/Models/Context/DeeplinkContext/DeeplinkContext.tsx
msnyder-msft: great. Glad we caught that.

pasanchMSFT (approved):
LGTM

PR #720, opened 2022-10-13
Visual Rule W/ Mocked Data

abalogun-dev:
To achieve this, the Status and Alert pivots will be removed. In their place a Visual rules pivot will be shown to display visual rules.

This list will contain all the visual rules that this behavior contains, overflow of the list will scroll as other lists on the right-hand side panel do. It is recommended to use Cardboard list component to render the list.

PLEASE EXCUSE THE COMMENTS, they are there for me to remember my logic, once the component is done, i'll remove them.

### Testing 🧪
 >
First, In browser console run this command, localStorage.setItem('cardboard.feature.visualRulesFeature',true);
You should be able to see visual list tab, just make sure you can see the list of rules and the add, delete and edit button work on click.

### Checklist ✔️
- [ ] Added relevant labels
https://msazure.visualstudio.com/One/_workitems/edit/15354443/


src/Components/ADT3DSceneBuilder/Internal/Behaviors/BehaviorsForm.tsx
```diff
@@ -592,6 +592,16 @@ const SceneBehaviorsForm: React.FC<IADT3DSceneBuilderBehaviorFormProps> = ({
                                 className={commonPanelStyles.formTabContents}
                                 headerText={t('3dSceneBuilder.visualRulesTab')}
                                 itemKey={BehaviorPivot.visualRules}
+                                onRenderItemLink={(props, defaultRenderer) =>
```
msnyder-msft, line 595: @pasanchMSFT  is this tab required? Why do we need a validity map for this tab?

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
```diff
@@ -1,7 +1,75 @@
...
+
+    return rules.map((item) => {
+        const viewModel: ICardboardListItem<IVisualRule> = {
+            ariaLabel: '',
+            iconStart: { name: '' },
+            item: item,
+            openMenuOnClick: true,
+            overflowMenuItems: getMenuItems(item),
+            textPrimary: item.displayName,
+            textSecondary: item.conditions.join(', ')
```
msnyder-msft, line 71: does this align to the latest designs? I think we decided to list out the number of conditions or something? @pasanchMSFT do you recall the final decision?

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRules.types.ts
```diff
@@ -1,5 +1,12 @@
...
-
 export interface IVisualRulesListProps {
-    ruleItems: IExpressionRangeVisual[];
+    ruleItems: IVisualRule[];
+    onRemoveRule: (ruleId: string) => void;
+    onEditRule: (ruleId: string) => void;
+}
+export type IConditionsType = 'Badge' | 'Mesh coloring';
+
+export interface IVisualRule {
```
msnyder-msft, line 8: don't we have a visual rule type @pasanchMSFT ? Can we share that here?

msnyder-msft (approved):
LGTM

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRules.types.ts
pasanchMSFT: These types we'll probably have to handle in a merge conflict, because it is something I overlooked when I did the schema change PR.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
pasanchMSFT: Seems like it matches the mock data, but when real data comes in this will have to change @abalogun-dev. This should actually show the number of badges and mesh colorings. i.e. `x badges, y mesh colorings`. Can be done when wired to real data.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/BehaviorsForm.tsx
pasanchMSFT: I agree, don't think we need validity map for this. Since this will spawn a form to edit the contents that will handle the validity of data.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/VisualRulesTab.tsx
```diff
@@ -1,6 +1,96 @@
...
+
+import { useTranslation } from 'react-i18next';
+import {
+    getActionButtonStyles,
+    getLeftPanelStyles
+} from '../../Shared/LeftPanel.styles';
 import { VisualRulesList } from '../VisualRules/VisualRuleList';
+import { IVisualRule } from '../VisualRules/VisualRules.types';
 
+const ROOT_LOC = '3dSceneBuilder.behaviorVisualRulesForm';
```
pasanchMSFT, line 13: nit: I would change this to '3dSceneBuilder.visualRulesTab', this is technically not a form. If there is already a '3dSceneBuilder.visualRulesTab' value, I would include it in this object ('3dSceneBuilder.visualRulesTab.tabName').

pasanchMSFT (approved):
LGTM, pretty much same comments as Mitch, and the nit comment I have for loc key names.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
pasanchMSFT: imo, this can change in next PR.

abalogun-dev: yup, it can. That'll be in the next PR

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRules.types.ts
pasanchMSFT: imo, this can change in next PR.

msnyder-msft: what do you mean? Do we not have an IVisualRule type in the app right now?

PR #724, opened 2022-10-18
[Visual Rules] Adding visual rule list tab

abalogun-dev:
### Testing 🧪

1.  > First enable visual rule flag in the console of your browser, 
2.  >You can use, mock scene data or real scene data > Go to Behaviors > Click a behavior > pivot to Visual rules tab
3.  >Delete a rule > save the behavior > return to the Behaviors list > Click on the previous behavior and make sure the deleted visual rule is no longer there
4.  >Go to the Visual rules tab > click on a Visual rule and/or click 'Edit Rule' in overflow section > make sure the "Edit Visual Rule form" shows
5.  >Go to the Visual rules tab > click Add Visual rule > Make sure 'Add Visual Rule form' shows 

### Checklist ✔️
- [ ] Added relevant labels
https://msazure.visualstudio.com/One/_workitems/edit/15354443/


src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/VisualRulesTab.tsx
```diff
@@ -17,50 +36,94 @@ const LOC_KEYS = {
...
-    const [
-        isVisualRuleFormOpen,
-        { toggle: toggleIsVisualRuleFormOpen }
-    ] = useBoolean(false);
+        return rules;
+    };
+
+    // set the list IVisualRules that are going to be shown
+    useEffect(() => {
+        setRuleItems(getVisualList());
```
msnyder-msft, line 112: since this has the behaviorFormState in the dependency list, can we pass it into the function as an argument so that the dependency is actually used? It is a bit suspect when we have a depencency on something that is not actually used in the `useEffect` directly. That means that we have to have some knowledge of how that function works internally which smells bad to me.

msnyder-msft (commented):
This is a good first cut. We need to handle the different scenarios for the secondary text as mentioned in the comments. It'd also be good if we can simplify the logic in the tab for building out the list items there.

src/Adapters/__mockData__/TruckAndMachinesConfig.json
```diff
@@ -25,7 +25,9 @@
                         "id": "4cb0990d646a4bbea3e1102676e200fe",
                         "displayName": "tank",
                         "primaryTwinID": "SaltMachine_C1",
-                        "objectIDs": ["tankMesh"],
```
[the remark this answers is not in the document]
abalogun-dev: i just added more expressions visuals so there are 3 visual rules under the Wheels too low behavior.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
```diff
@@ -35,6 +35,23 @@ export const VisualRulesList: React.FC<IVisualRulesListProps> = ({
     );
 };
 
+function getBadgesAndMeshesCount(item: IVisualRule, t: TFunction<string>) {
+    let badgeCount = 0;
+    let meshColoringCount = 0;
+    item.conditions.forEach((condition) => {
+        if (condition.visual.iconName) {
```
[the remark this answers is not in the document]
pasanchMSFT: Nope, it would require a new schema change to add action type. I believe it is not necessary since we can determine the type if it contains or not an iconName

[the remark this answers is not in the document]
msnyder-msft: maybe we can just pull this check out to the ViewerConfigUtility like we do for all the other types so this check is centralized at least. In the future, it might be worth putting a type on any new object we create where not all props are used for all variants of that object.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/BehaviorsForm.tsx
```diff
@@ -105,6 +105,7 @@ const SceneBehaviorsForm: React.FC<IADT3DSceneBuilderBehaviorFormProps> = ({
     onRemoveElement
 }) => {
     const { t } = useTranslation();
+    const [visualRuleId, setVisualRuleId] = useState<string | null>('');
```
pasanchMSFT, line 108: nit: remove the null type now that we have made id empty string

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Internal/VisualRulesTab.tsx
```diff
@@ -9,58 +8,88 @@ import {
...
+import {
+    IBehavior,
+    IExpressionRangeVisual
+} from '../../../../../Models/Types/Generated/3DScenesConfiguration-v1.0.0';
+import ViewerConfigUtility from '../../../../../Models/Classes/ViewerConfigUtility';
+import { BehaviorFormContextActionType } from '../../../../../Models/Context/BehaviorFormContext/BehaviorFormContext.types';
+
+//get an array of all the visual rules in the behavior
+const getVisualRulesFromBehavior = (behavior: IBehavior) =>
+    behavior.visuals.filter(ViewerConfigUtility.isVisualRule) || null;
```
pasanchMSFT, line 21: would it be better if this returned an empty array instead of null?

pasanchMSFT (approved):
Looks good, left some nit comments but overall functionality looks good.

src/Resources/Locales/en.json
```diff
@@ -461,7 +461,11 @@
             "noData": "You haven't configured any Visual Rules yet",
             "addRuleButtonText": "Add Rule",
             "editRule": "Edit rule",
-            "removeRule": "Remove rule"
+            "removeRule": "Remove rule",
```
msnyder-msft, line 464: we usually call it Delete instead of Remove in our app. Can we update that?

abalogun-dev: really? i saw remove 
image

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
```diff
@@ -45,10 +45,36 @@ function getBadgesAndMeshesCount(item: IVisualRule, t: TFunction<string>) {
...
-    });
+
+    return [badgeCount, meshColoringCount];
+}
+
+function getSecondaryText(item: IVisualRule, t: TFunction<string>) {
+    const [badgeCount, meshColoringCount] = getBadgesAndMeshesCount(item);
+    const badgeCondition =
+        badgeCount > 1
+            ? t('3dSceneBuilder.behaviorVisualRulesTab.multipleBadges', {
```
msnyder-msft, line 56: we should really use multiple locale strings like I mentioned earlier. In languages other than english they might not concatenate these values together the same way. So we should have 3 locale strings, 1. Only badges, 2 only Color, 3. both

msnyder-msft (approved):
Looking really good! The interaction is nice now that it's all wired up.
We just need to update the locale string logic and then it should be good.

src/Resources/Locales/en.json
msnyder-msft: gahh. ok. You got me. Leave it.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
abalogun-dev: i guess i'm not this comment, i have the locale string setup like 
image

msnyder-msft: yeah. but you concatenate them with a , which might not be valid in spanish or something.

abalogun-dev: I know for sure in spanish it's valid but i get your point...so like should i include the comma in the locale strings??

msnyder-msft: yeah. but thinking of the plural cases it gets kind of verbose. :( We should need 
1. single badge
2. multi badge
3. single color
4. multi color
5. single badge + multi color
6. multi badge + single color
7. multi badge + multi color

msnyder-msft: So what you have is a good start, but then add the extra conditions for the combined strings

PR #727, opened 2022-10-19
Conditions flyout form

pasanchMSFT:
### Summary of changes 🔍 
These changes should enable the Conditions flyout to be enabled so users can create/edit conditions in the Visual rule form. This should wrap up the overall functionality of Status and Alerts merge on the Builder side, however flighting is still in place until the viewer experience is updated.

TODO (in following PRs): 
- Remove flighting, will be done after viewer experience is updated.
- Add icon in Conditions list items
- Viewer experience

#### Flight local storage key, to be able to see the functionality:
- Open developer tools of the browser and select application tab.
- Open localstorage variables
- Add this key: cardboard.feature.visualRulesFeature and set its value to true
- Refresh the page
OR -
- Open browser dev tools and run this in console: localStorage.setItem('cardboard.feature.visualRulesFeature',true);

#### To test:
- Open a scene and navigate to behaviors form
- Click on visual rules button
- Test out the UI for visual rules form. Suggested things to check out:
  - Edit a condition
  - Change the property type or expression and write new conditions based on those
  - Save changes and come back to review them
  - Click different conditions on the list to be able to see their values

### Checklist ✔️
- [x] Added relevant labels


src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { ComponentStory } from '@storybook/react';
+import { getDefaultStoryDecorator } from '../../../../../../Models/Services/StoryUtilities';
+import ConditionsCallout from './ConditionsCallout';
+import { IConditionsCalloutProps } from './ConditionsCallout.types';
+
+const wrapperStyle = { width: '100%', height: '600px', padding: 8 };
+
+export default {
+    title: 'Components/ConditionsCallout',
```
msnyder-msft, line 10: can we make a folder for all the visual rule components?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { ComponentStory } from '@storybook/react';
+import { getDefaultStoryDecorator } from '../../../../../../Models/Services/StoryUtilities';
+import ConditionsCallout from './ConditionsCallout';
+import { IConditionsCalloutProps } from './ConditionsCallout.types';
+
+const wrapperStyle = { width: '100%', height: '600px', padding: 8 };
+
+export default {
```
msnyder-msft, line 9: this story is busted
image

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.stories.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { ComponentStory } from '@storybook/react';
+import { getDefaultStoryDecorator } from '../../../../../../Models/Services/StoryUtilities';
+import ConditionsCallout from './ConditionsCallout';
+import { IConditionsCalloutProps } from './ConditionsCallout.types';
+
+const wrapperStyle = { width: '100%', height: '600px', padding: 8 };
+
```
msnyder-msft, line 8: is there really only one scenario here to validate? I can't tell since the story won't load, but just asking to prompt the thinking.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -41,63 +65,109 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
...
+                        textSecondary: condition.secondaryText,
+                        overflowMenuItems: getOverflowMenuItems(condition.id),
+                        onClick: () => {
+                            setCalloutInfo({
+                                calloutType: CalloutInfoType.edit,
+                                isOpen: true,
+                                selectedCondition: valueRanges.find(
+                                    (vr) => vr.id === condition.id
+                                ),
+                                selectedTarget: `#${LIST_KEY}`
```
msnyder-msft, line 159: the flyout should open next to the item you are editing. Right now it only seems to open in a static location.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.styles.ts
```diff
@@ -0,0 +1,130 @@
...
+                root: {
+                    width: 300,
+                    padding: 20
+                }
+            }
+        }
+    };
+};
+
+export const getSummaryStyles = (theme: ITheme) =>
```
msnyder-msft, line 48: can we please put the styles for each component in their own file so they are easy to find? We follow that pattern across teh repo and it's good to stay consistent.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.styles.ts
```diff
@@ -0,0 +1,130 @@
...
+    mergeStyleSets({
+        title: {
+            margin: 0
+        },
+        invalidText: {
+            color: theme.semanticColors.errorText,
```
msnyder-msft, line 54: I think we actually usually use one of our scss variable for error text. This might match fine but might be worth checking other places we show errors. I didn't actually know they had this semantic color so this is good to know about. Let's just make sure it's consistent.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.styles.ts
```diff
@@ -0,0 +1,130 @@
...
+            border: `1px solid ${theme.semanticColors.inputBorder}`,
+            borderRadius: 2,
+            color: theme.semanticColors.inputText,
+            fontSize: FontSizes.size14,
+            height: 32,
+            padding: '0 8px 0 8px',
+            userSelect: 'text',
+            width: 100
+        } as IStyle,
+        negativeInfinityButton: {
```
msnyder-msft, line 83: these look like they are probably copy paste from the last time we implemented this control. Can we pull these styles out to a standalone control so we dont' have to keep duplicating them in the future?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
```diff
@@ -0,0 +1,255 @@
...
+    ] = useReducer<ConditionCalloutReducerType>(ConditionCalloutReducer, {
+        conditionToEdit: valueRange,
+        originalCondition: valueRange,
+        isDirty: false
+    });
+
+    // side-effects
+    useEffect(() => {
+        // RESET CONDITION IN STATE AND VALIDITY MAP
+        conditionCalloutDispatch({
```
msnyder-msft, line 78: I'm confused by this useEffect. We take in a valueRange via props and then any time that changes we update the dispatch with that value? Why do we need to do that?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
```diff
@@ -0,0 +1,255 @@
...
+        };
+    }, [valueRange, valueRangeType]);
+
+    // callbacks
+    const onLabelChange = useCallback(
+        (
+            _event: React.FormEvent<HTMLInputElement | HTMLTextAreaElement>,
+            newValue?: string
+        ) => {
+            if (newValue) {
```
msnyder-msft, line 94: nit: might want to trim the value here just to avoid setting to valid when it's empty string.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
```diff
@@ -0,0 +1,255 @@
...
+                payload: { label: newValue }
+            });
+        },
+        [conditionCalloutDispatch]
+    );
+
+    const onValuesChange = useCallback(
+        (
+            valueType: IDTDLPropertyType,
+            newValues: unknown[],
```
msnyder-msft, line 110: why don't we know the type of the values? surely it's at least string[] or number[]?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
```diff
@@ -0,0 +1,255 @@
...
+            {isOpen && (
+                <Callout
+                    target={target}
+                    styles={classNames.subComponentStyles.callout}
+                    onDismiss={onDismiss}
+                    directionalHint={DirectionalHint.rightCenter}
+                >
+                    <Stack tokens={{ childrenGap: 12 }}>
+                        {calloutType === CalloutInfoType.create ? (
+                            <h4 className={classNames.title}>
```
msnyder-msft, line 184: nit: instead of duplicating the whole h4 tag here, can we just switch the string we include?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -8,22 +8,39 @@ import {
...
 import {
+    CalloutInfo,
+    CalloutInfoType,
     Condition,
     IConditionsListProps,
     IConditionsListStyles,
     IConditionsListStylesProps
 } from './ConditionsList.types';
 
 const LIST_KEY = 'cb-visual-rule-conditions-list';
```
msnyder-msft, line 34: we usually call `useId` for ids of elements right?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -8,22 +8,39 @@ import {
...
+const getDefaultCondition = (type: IDTDLPropertyType): IValueRange => ({
```
msnyder-msft, line 35: we actually have a bunch of these "default" initializers somewhere in the app. It might be in ViewerConfigUtility? Let's add this one there as well to have them all together.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -41,63 +65,109 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
...
+        calloutType: CalloutInfoType.inactive,
+        isOpen: false,
+        selectedCondition: null,
+        selectedTarget: null
+    });
+
+    // Callbacks
+    const handleOpenNewConditionFlyout = useCallback(() => {
+        setCalloutInfo({
+            calloutType: CalloutInfoType.create,
```
msnyder-msft, line 79: can't you tell whether to open it by when the calloutType === inactive? do we need the `isOpen` flag too?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -41,63 +65,109 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
...
+            isOpen: true,
+            selectedCondition: getDefaultCondition(valueRangeType),
+            selectedTarget: `#${LIST_KEY}`
+        });
+    }, [valueRangeType]);
+
+    const handleDismissFlyout = useCallback(() => {
+        setCalloutInfo({
+            calloutType: CalloutInfoType.inactive,
```
msnyder-msft, line 88: can we create a "defaultInfo" object? It seems like that could be used in both the initial setState as well as the dismiss function. I'm guessing we want those to be the same values so we want dismiss to go back to the same as the initial state.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -106,30 +176,41 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
...
+    }, [valueRanges, expressionType, getConditionItems]);
 
     return (
-        <div className={classNames.container}>
-            <Stack>
-                <CardboardList<Condition>
-                    listKey={LIST_KEY}
-                    items={conditions}
+        <>
+            <div className={classNames.container}>
```
msnyder-msft, line 183: naming nit; we generally call the outer most container "root", following the pattern from Fluent.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -106,30 +176,41 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
...
+                        id={'visual-rule-add-condition'}
+                        data-testid={'visual-rule-add-condition'}
+                        styles={classNames.subComponentStyles.addButton?.()}
+                        onClick={handleOpenNewConditionFlyout}
+                    >
+                        {t('3dSceneBuilder.visualRuleForm.newCondition')}
+                    </ActionButton>
+                </Stack>
+            </div>
+            {calloutInfo.isOpen && (
```
msnyder-msft, line 202: do we need to add/remove from teh DOM here? doesn't the callout internally have show/hide logic based on isOpen? (also can probably move that to the calloutType)

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleForm.state.ts
```diff
@@ -33,11 +33,12 @@ export const VisualRuleFormReducer: IVisualRuleFormReducerType = produce(
         );
         switch (action.type) {
             case VisualRuleFormActionType.FORM_VISUAL_RULE_DISPLAY_NAME_SET: {
-                draft.visualRuleToEdit.displayName = action.payload.name.trim();
```
msnyder-msft, line 36: how come we remove the trim here?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleForm.tsx
```diff
@@ -222,31 +225,53 @@ const VisualRuleForm: React.FC<IVisualRuleFormProps> = (props) => {
...
-    }, []);
+    const onInternalModeChanged = useCallback(
+        (internalMode) => {
+            if (internalMode === 'INTELLISENSE') {
+                visualRuleFormDispatch({
+                    type:
+                        VisualRuleFormActionType.FORM_VISUAL_RULE_EXPRESSION_TYPE_SET,
+                    payload: { type: 'CategoricalValues' }
+                });
+                handleExpressionTextFieldEnabled(true);
```
msnyder-msft, line 236: what was the reasoning behind keeping a local state variable while also updating dispatch? Does it make sense to move this to the reducer since you're alwayss setting this value when you call that particular dispatch action?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleForm.tsx
```diff
@@ -258,6 +283,24 @@ const VisualRuleForm: React.FC<IVisualRuleFormProps> = (props) => {
...
+    const handleSaveCondition = useCallback(
+        (condition: IValueRange) => {
+            visualRuleFormDispatch({
+                type: VisualRuleFormActionType.FORM_CONDITION_ADD_OR_UPDATE,
+                payload: {
+                    condition: {
+                        ...condition,
+                        visual: {
+                            ...condition.visual,
+                            labelExpression: condition.visual.labelExpression.trim()
```
msnyder-msft, line 295: I see we have to do this trimming again on save, why do we not just store the valid value so we don't have to do these manipulations and all the object expansions?

src/Components/BehaviorsModal/Internal/BehaviorSection/BehaviorSection.tsx
```diff
@@ -77,30 +77,33 @@ const BehaviorSection: React.FC<IBehaviorsSectionProps> = ({ behavior }) => {
 const AlertBlock: React.FC<{ alertVisual: IExpressionRangeVisual }> = ({
     alertVisual
 }) => {
-    const styles = getStyles();
-
-    const {
-        visual: { color, iconName, labelExpression }
-    } = alertVisual.valueRanges[0];
+    if (alertVisual.valueRanges[0]) {
```
msnyder-msft, line 80: nit: shorter way to do this is to check the inverse, then you don't have to wrap the rest of the render block. 
```jsx
if(!myCheck){
  return null
}
```

src/Components/CardboardMultiSelect/CardboardMultiSelect.styles.ts
```diff
@@ -0,0 +1,101 @@
...
+    _props: ICardboardMultiSelectStyleProps
+): ICardboardMultiSelectStyles => {
+    return {
+        root: [classNames.root],
+        input: [classNames.input],
+        subComponentStyles: {}
+    };
+};
+
+export const getMultiSelectStyles = (theme: Theme): StylesConfig => {
```
msnyder-msft, line 23: can we not inherit some of our styling from the other react-select control? This feels very duplicative.

src/Components/CardboardMultiSelect/CardboardMultiSelect.tsx
```diff
@@ -0,0 +1,117 @@
...
+    // hooks
+    const { t } = useTranslation();
+
+    // callbacks
+    const handleKeyDown = useCallback(
+        (event: React.KeyboardEvent<HTMLDivElement>) => {
+            if (!inputValue) return;
+            switch (event.key) {
+                case 'Enter':
+                case 'Tab': {
```
msnyder-msft, line 55: tab too? That doesn't feel like standard behavior to me. Can you still tab out of the control?

src/Components/CardboardMultiSelect/CardboardMultiSelect.stories.tsx
```diff
@@ -0,0 +1,24 @@
...
+};
+
+type CardboardMultiSelectStory = ComponentStory<typeof CardboardMultiSelect>;
+
+const Template: CardboardMultiSelectStory = (args) => {
+    return <CardboardMultiSelect {...args} />;
+};
+
+export const Base = Template.bind({}) as CardboardMultiSelectStory;
+Base.args = {} as ICardboardMultiSelectProps;
```
msnyder-msft, line 24: can we have a story for having values entered in as well?

src/Resources/Locales/en.json
```diff
@@ -568,14 +568,27 @@
...
+            "unlabeledCondition": "Unlabeled",
+            "actionsTitle": "Actions",
+            "actionLabel": "Action",
+            "colorLabel": "Color",
+            "iconLabel": "Icon",
+            "invalidRanges": "Invalid range, min should be less than max, and values should be numeric.",
+            "labelField": "Label",
+            "choiceGroupTrue": "True",
+            "choiceGroupFalse": "False",
+            "placeholderMultiselect": "Type values, press enter to add"
```
msnyder-msft, line 591: is this used only in your visual rule form or also in the multi select control?

src/Components/ADT3DSceneBuilder/Internal/Behaviors/BehaviorsForm.tsx
```diff
@@ -399,27 +399,40 @@ const SceneBehaviorsForm: React.FC<IADT3DSceneBuilderBehaviorFormProps> = ({
...
     const onVisualRuleSaveClick = useCallback(
         (visualRule: IExpressionRangeVisual) => {
             behaviorFormDispatch({
                 type:
                     BehaviorFormContextActionType.FORM_BEHAVIOR_VISUAL_RULE_ADD_OR_UPDATE,
-                payload: { visualRule: visualRule }
+                payload: {
+                    visualRule: {
+                        ...visualRule,
+                        displayName: visualRule.displayName.trim(),
```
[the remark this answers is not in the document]
pasanchMSFT: The visual rule controls a text field, `.trim()` removes blank spaces when the user inputs a space on the text field.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
```diff
@@ -64,7 +64,7 @@ function getSecondaryText(item: IVisualRule, t: TFunction<string>) {
             ? t('3dSceneBuilder.behaviorVisualRulesTab.multipleMeshColorings', {
                   meshColoringCount: meshColoringCount
               })
-            : t('3dSceneBuilder.behaviorVisualRulesTab.multipleMeshColoring', {
+            : t('3dSceneBuilder.behaviorVisualRulesTab.singleMeshColoring', {
                   meshColoringCount: meshColoringCount
```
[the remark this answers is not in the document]
pasanchMSFT: Fair, but designs called for 'n mesh coloring(s)', I'll leave the more explicit version of including a number. I also believe this `m badges, 1 mesh coloring` looks better than `m badges, mesh coloring`.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ActionItem.tsx
```diff
@@ -0,0 +1,117 @@
...
+    // hooks
+    const { t } = useTranslation();
+
+    // state
+    const [selectedOption, setSelectedOption] = useState(
+        iconName ? DROPDOWN_OPTIONS[1].key : DROPDOWN_OPTIONS[0].key
+    );
+
+    // side-effects
+    useEffect(() => {
```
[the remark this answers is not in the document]
pasanchMSFT: The callout does not unmount when another item is clicked, I have to sync props to match the new values.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ActionItem.tsx
```diff
@@ -0,0 +1,117 @@
...
+                        onChangeItem={onColorChange}
+                        styles={{
+                            // match the icon picker
+                            button: {
+                                height: 32,
+                                width: 32
+                            }
+                        }}
+                    />
+                    {selectedOption === 'Badge-action-item' && (
```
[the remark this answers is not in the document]
pasanchMSFT: just tried to make it unique, nothing special

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/BoundaryInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    useEffect(() => {
+        setInputValue(value);
+    }, [value]);
+
+    const isMin = boundary === Boundary.min;
+    const infinityIconMessage = isMin
+        ? t('valueRangeBuilder.negativeInfinityIconMessage')
+        : t('valueRangeBuilder.positiveInfinityIconMessage');
+    const styles = getBoundaryInputStyles(useTheme());
```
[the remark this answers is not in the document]
pasanchMSFT: Question on this, are we going to make every single component styled?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/BoundaryInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    return (
+        <>
+            <div className={styles.container}>
+                <label className={styles.label} htmlFor={guid}>
+                    {isMin ? t('min') : t('max')}
+                </label>
+                <div className={styles.inputContainer}>
+                    <input
```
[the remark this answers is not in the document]
pasanchMSFT: These won't be used elsewhere, the previous component where they are used will be removed. The previous component using these is using SCSS for styling and is dependent on a context, re-writing this to a re-usable component seems like a waste of time.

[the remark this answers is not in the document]
pasanchMSFT: I believe the intention was to allow the case were there are number inputs + the infinity value. Re-writing this into a re-usable control seems out of scope for this PR but a good suggestion. I'll add it to the cleanup tasks on our board.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionCalloutUtility.ts
```diff
@@ -0,0 +1,78 @@
...
+    return hasConditionChanged;
+}
+
+export function checkValidity(validityMap: ConditionValidityMap): boolean {
+    return validityMap.label && validityMap.ranges;
+}
+
+/** Looks at a min-max value and determines if it is valid */
+export const getRangeValidation = (
+    values: unknown[]
```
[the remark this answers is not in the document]
pasanchMSFT: I can cast it at some point, but since this comes from the ValueRange type it will be unknown.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
pasanchMSFT: This requires an edit to cardboard list, currently the list items don't take in an id as props and I can't target them. I can add the prop and change the target in a following PR.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
pasanchMSFT: The callout does not unmount when another item is clicked on the list, the reducer therefore is not re-initialized. This use effect resets the reducer state when the item selected is changed.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
pasanchMSFT: String, number or boolean. I could narrow it down to those 3

pasanchMSFT: I'd have to cast it though, since the type is unknown[ ] in our types file.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleForm.state.ts
pasanchMSFT: Trim here doesn't allow blank space inputs. If I want my label to be `New label` the space would be trimmed every time it is the last character.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleForm.tsx
pasanchMSFT: See answers on trimming above.

src/Components/CardboardMultiSelect/CardboardMultiSelect.tsx
pasanchMSFT: good catch!

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleForm.tsx
pasanchMSFT: This is something the parent component needs to manage, since it is the component using this value to set styles correctly. This function is a prop.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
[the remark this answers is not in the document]
msnyder-msft: I agree that it should have 1 in the value. I'm just saying it could be in the localized string instead of a variable.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
msnyder-msft: I'd rather do a caste at one point than have unknown spread throughout the codebase. I'm finding all sorts of crazy bugs and issues in OAT because of their use of any and unknowns.

msnyder-msft (commented):
Getting close. There's still quite a few open comments taht have been acknowledged but not fixed yet. Once those are completed we can get it in.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/BoundaryInput.tsx
```diff
@@ -0,0 +1,84 @@
...
+                        id={guid}
+                        value={String(inputValue)}
+                        type="string"
+                        onChange={(event) => setInputValue(event.target.value)}
+                        className={styles.input}
+                        onBlur={() => {
+                            setNewValues(inputValue);
+                        }}
+                    />
+                    <button
```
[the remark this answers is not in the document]
pasanchMSFT: Can fluent handle svgs as the content of their buttons?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.state.ts
```diff
@@ -0,0 +1,68 @@
...
+                draft.conditionToEdit.visual.labelExpression =
+                    action.payload.label;
+                break;
+            }
+            case ConditionCalloutActionType.FORM_CONDITION_VALUES_SET: {
+                draft.conditionToEdit.values = action.payload.values;
+                break;
+            }
+            case ConditionCalloutActionType.RESET_CONDITION: {
+                const resetValueRange = deepCopy(action.payload.valueRange);
```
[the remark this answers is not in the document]
pasanchMSFT: Will call initialize

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.styles.ts
msnyder-msft: That should be a good red then. let's just make sure it hits contrast ratios.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.state.ts
```diff
@@ -0,0 +1,68 @@
...
+                draft.conditionToEdit.visual.iconName = action.payload.iconName;
+                break;
+            }
+            case ConditionCalloutActionType.FORM_CONDITION_LABEL_SET: {
+                draft.conditionToEdit.visual.labelExpression =
+                    action.payload.label;
+                break;
+            }
+            case ConditionCalloutActionType.FORM_CONDITION_VALUES_SET: {
+                draft.conditionToEdit.values = action.payload.values;
```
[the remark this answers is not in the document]
pasanchMSFT: Values can either be `[true/false]`, `[number, number]` or `[strings...]`. I think for the string case it could make sense but don't see harm in keeping it as a single dispatch and using the set pattern.

msnyder-msft (approved):

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
```diff
@@ -0,0 +1,241 @@
...
+    editCondition: `${ROOT_LOC}.editCondition`,
+    labelField: `${ROOT_LOC}.labelField`
+};
+
+const ConditionsCallout: React.FC<IConditionsCalloutProps> = (props) => {
+    const {
+        calloutType,
+        onDismiss,
+        onSave,
+        target,
```
iremgokce, line 49: When I click on Add new rule, the flyout appears all the way to the right, was it intended? Cant we make it appear close to target which is Add rule action button?

Also see the values prefilled with "undefined", was it also intended?

Last thing, is the order of Save, cancel buttons. As far as I remember the pattern we use is secondary, primary
image

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleForm.tsx
```diff
@@ -258,6 +283,24 @@ const VisualRuleForm: React.FC<IVisualRuleFormProps> = (props) => {
         [visualRuleFormDispatch]
     );
 
+    const handleSaveCondition = useCallback(
```
iremgokce, line 286: is it indented that we can create a rule without condition? The save button is getting enabled without me adding any condition.
image

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -106,30 +158,40 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
...
     return (
-        <div className={classNames.container}>
-            <Stack>
-                <CardboardList<Condition>
-                    listKey={LIST_KEY}
-                    items={conditions}
+        <>
+            <div className={classNames.root}>
+                <Stack>
+                    <CardboardList<Condition>
```
iremgokce, line 167: What happens when I add overlapping ranges for conditions? Also the visual representation of between using "-" might be confusing since we allow negative numbers: For example I added between -100 and 10 and -10 and 100 as two conditions but it seems a little confusing in the list since there is another dash is used to represent min and max values between. What do you think about this @azimamansuri ?
image

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
iremgokce: ah also with the edit condition callout, the beak shows a little upward of the selected list item, can we center it with the selected list items overflow menu?:
image

iremgokce:
it is interesting that I got schema validation error as I update the behavior with two conditions that I just created but the request to update the schema went through, but now when I refresh page I get schema error:
image
image
image

iremgokce (commented):
added some comments after testing it

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
```diff
@@ -0,0 +1,241 @@
...
+                    onChangeValues={onValuesChange}
+                    currentValues={
+                        conditionCalloutState.conditionToEdit?.values
+                    }
+                    conditionType={valueRangeType}
+                    areValuesValid={validityMap.current.ranges}
+                />
+                <ActionItem
+                    setActionSelectedValue={onActionValueChange}
+                    color={conditionCalloutState.conditionToEdit?.visual.color}
```
iremgokce, line 208: The default color value seems null:
image

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+    calloutType: CalloutInfoType.create,
+    onDismiss: () => {
+        return;
+    },
+    onSave: (_condition) => {
+        return;
+    },
+    target: '',
+    valueRange: booleanCondition,
+    valueRangeType: 'boolean'
```
iremgokce, line 52: nit: is there any enum you use for valueRangeTypes? maybe it would be nice to use it instead of direct strings here.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
pasanchMSFT: I will fix this in a future PR, there needs to be an update to cardboard list to allow id so I can target the item as the callout target. Good catch on the 'new condition' button, I'll fix that as well in the PR.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
pasanchMSFT: Overlapping ranges are allowed, it was part of the update. The dash is something I'll take up with Azima, good catch.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.stories.tsx
pasanchMSFT: valueRangeTypes are tied to DTDL property types, I don't think we have an enum for that or do we?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
pasanchMSFT: First thing got fixed

iremgokce: sounds good!

iremgokce (approved):
this was a huge refactor, great job pulling everything together!

PR #728, opened 2022-10-19
CardboardList: Add custom icon support to cardboard list

msnyder-msft:
### Summary of changes 🔍 
Adding support for rendering arbitrary content at the start and end of the list item in place of the icons.
Consumers will be responsible for dealing with the right amount of spacing on the start icon since that would require a lot of backwards validation on existing stuff to add a stack.

Both start & end customized
image

Start only
image

End only
image

### Testing 🧪
Check the following stories
- [x] Lists --> Items --> WithStartIconCustom
- [x] Lists --> Items --> WithEndIconCustom
- [x] Lists --> Items --> WithStartAndEndIconCustom

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/CardboardList/CardboardList.types.ts
+++ b/src/Components/CardboardList/CardboardList.types.ts
@@ -27,14 +27,18 @@ type IListItemBaseProps<T> = {
         customStyles?: IButtonStyles;
     };
     /** icon to render on the right side of the list item */
-    iconEnd?: {
-        name: IIconNames;
-        onClick?: (item: T) => void;
-    };
+    iconEnd?:
+        | {
+              name: IIconNames;
+              onClick?: (item: T) => void;
+          }
+        | ((item: T) => React.ReactElement);
     /** icon or JSX element to render at the left side of the list item */
-    iconStart?: {
-        name: IIconNames | JSX.Element;
-    };
+    iconStart?:
+        | {
+              name: IIconNames | JSX.Element;
+          }
+        | ((item: T) => React.ReactElement);
     /** if provided false will result in rendering the red dot at the very left of the element. If not provided, will assume it is valid and not render any dot */
     isValid?: boolean;
     /** if provided will result in rendering the checkbox in either checked or unchecked state. If not provided, will not render a checkbox */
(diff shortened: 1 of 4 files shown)


src/Components/CardboardList/CardboardList.types.ts
```diff
@@ -27,14 +27,18 @@ type IListItemBaseProps<T> = {
...
-    iconEnd?: {
-        name: IIconNames;
-        onClick?: (item: T) => void;
-    };
+    iconEnd?:
+        | {
+              name: IIconNames;
+              onClick?: (item: T) => void;
+          }
+        | ((item: T) => React.ReactElement);
```
pasanchMSFT, line 35: This is merely a question about the feasibility of something, feel free to resolve. Do you know if it is possible to type this with a specific component signature? i.e. `((item: T) => IconComponentType)`. If possible, does it make sense to have that constraint here?

pasanchMSFT (approved):

src/Components/CardboardList/CardboardList.types.ts
```diff
@@ -27,14 +27,18 @@ type IListItemBaseProps<T> = {
...
     /** icon or JSX element to render at the left side of the list item */
-    iconStart?: {
-        name: IIconNames | JSX.Element;
-    };
+    iconStart?:
+        | {
+              name: IIconNames | JSX.Element;
```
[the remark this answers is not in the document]
msnyder-msft: we don't have any use cases for having a click action there though. And IMO that is not really an intuitive place to click. I'd rather hold off till we decide if that's a pattern we really want to start using. It's something that's pretty easy to extend if we do want it.

src/Components/CardboardList/CardboardList.types.ts
msnyder-msft: I'm not sure I follow the question. Right now this is passing back the list item itself. The return type is any React element you want. I believe `ReactElement` is actually a union type of JSXElement as well as some other stuff maybe? I know it's more broad than just JSXElement but I don't recall exactly why that's better, just that we used to use it on MSX portal. :)

msnyder-msft: ohhh. idk if you can do that. But I'm also not sure we want to do that here since the whole idea is to be custom. Ideally you woulnd't try to render text here but I guess that's on you if you use it that way! 😜

abalogun-dev (approved):

PR #732, opened 2022-10-21
[Data History]: Title changes in widget and quick time dropdown

iremgokce:
### Summary of changes 🔍 
- Added `QuickTimesDropdown` standalone component to list quick time spans
- Consume it in DataHistoryWidgetBuilder and DataHistoryWidget's overflow menu
- Changed props for HighchartWrapper component to make title optional and put the title and ellipsis overflow menu (including deeplink share option and quick time span) as a separate dom element in the DataHistoryWidget.
Ellipsis1

### Testing 🧪
Check out QuickTimesDropdown stories to see the standalone component and Mock3DScenePage or ADT3DScenePage in builder and view mode for DataHistoryWidget to see the title and ellipsis menu (ellipsis menu is only visible in view mode)

### Checklist ✔️
- [x] Added relevant labels


src/Models/Constants/Constants.ts
```diff
@@ -542,3 +543,20 @@ export const RequiredAccessRoleGroupForADTInstance: AzureAccessPermissionRoleGro
 };
 
 export const CONNECTION_STRING_SUFFIX = '.kusto.windows.net';
+
+/** Quick time span key to value in millisecond mapping */
+export const QuickTimeSpans = {
```
msnyder-msft, line 548: nit: can we name this more generically since it's really just a lookup for time durations? Or does this object have to only contain these entries?

msnyder-msft: If it is specific to that control, can we move it to a contants file or that component instead of the global contants? (same for the enum)

src/Components/QuickTimesDropdown/QuickTimesDropdown.tsx
```diff
@@ -0,0 +1,71 @@
...
+import { TFunction } from 'i18next';
+import { QuickTimeSpans } from '../../Models/Constants/Constants';
+import { useTranslation } from 'react-i18next';
+
+const getClassNames = classNamesFunction<
+    IQuickTimesDropdownStyleProps,
+    IQuickTimesDropdownStyles
+>();
+
+const QuickTimesDropdown: React.FC<IQuickTimesDropdownProps> = (props) => {
```
msnyder-msft, line 24: It's not very standard to have a dropdown inside a context menu. The recommended/standard pattern for context menus is to have a nested menu like this. That should come natively in context menu if simply pass in the `subMenuProps` item to the context menu item.

image


image

msnyder-msft (changes requested):
The code looks great. I do think we need to swap from dropdown to sub menu though.

src/Components/QuickTimesDropdown/QuickTimesDropdown.tsx
iremgokce: yep I am aware of that prop and my initial commit was with using it, but since Matt talked about it to be a dropdown that is why I made a standalone component that we can consume both in the builder and viewer. @darsney do you have a preference here?

src/Models/Constants/Constants.ts
iremgokce: yeah it is a specific control, when I was moving the things around I forgot to move those to that standalone component, will move it there now.

iremgokce: oh in fact, I just noticed that why I kept those here as globals was because of the getQuickTimeSpanKeyByValue method which is used by multiple components but it also uses QuickTimeSpans inside, that is why QuickTimeSpans was in Constants. But I will move it to that component, let me know if it is okay (I know that we talked about not having methods in Utils which do imports from specific components type files but in this case I think it make sense). 2795f0678dc26486dfb58a6995891ae8184f9924

src/Components/QuickTimesDropdown/QuickTimesDropdown.types.ts
```diff
@@ -0,0 +1,50 @@
...
+    IStyle,
+    IStyleFunctionOrObject,
+    ITheme
+} from '@fluentui/react';
+import { QuickTimeSpanKey } from '../../Models/Constants/Enums';
+
+export interface IQuickTimesDropdownProps {
+    defaultSelectedKey?: QuickTimeSpanKey;
+    hasLabel?: boolean;
+    label?: string;
```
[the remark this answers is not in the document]
iremgokce: since there is a default label "Time span" within the component, this is additional check to see the intention if we want to show it or not:
image

src/Components/HighChartsWrapper/HighChartsWrapper.types.ts
```diff
@@ -27,11 +27,10 @@ export interface IHighChartSeriesData {
 }
 
 export interface IHighChartsWrapperProps {
-    title: string;
+    title?: string;
```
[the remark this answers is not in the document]
iremgokce: yes, dont forget this is the wrapper standalone component. yes, we dont use the title in data history widget, but if in the future if someone uses this highchart wrapper component and want to use the built in title, they should be able to

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.tsx
```diff
@@ -110,26 +137,106 @@ const DataHistoryWidget: React.FC<IDataHistoryWidgetProps> = ({
...
+        );
+    };
+    const menuProps: IOverflowMenuProps = {
+        ariaLabel: t('widgets.dataHistory.headerMenu'),
+        index: 0,
+        menuKey: `${widget.id}-overflow-menu`,
+        menuProps: {
+            directionalHint: DirectionalHint.bottomRightEdge,
+            items: [
+                {
```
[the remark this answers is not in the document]
iremgokce: as per our discussion moved it outside but kept the overflow menu

iremgokce:
> The code looks great. I do think we need to swap from dropdown to sub menu though.

since I am using the QuickTimeDropdown component in DataHistoryWIdgetBuilder as well, I prefer to keep it as dropdown as it is more flexible component to integrate as in my two use cases.

src/Models/Constants/Constants.ts
msnyder-msft: can we move the util to that component as well? 🤔

src/Components/QuickTimesDropdown/QuickTimesDropdown.tsx
msnyder-msft: since we opted to have it not in the menu we can ignore this now.

src/Models/Services/Utils.ts
```diff
@@ -38,6 +23,23 @@ import { createParser, ModelParsingOption } from 'azure-iot-dtdl-parser';
...
+    ComponentErrorType
+} from '../Constants/Enums';
+import { DTwin, IADTTwin, IAzureResource } from '../Constants/Interfaces';
+import {
+    AzureAccessPermissionRoleGroups,
+    DurationUnits,
+    IConsoleLogFunction,
+    TimeSeriesData
+} from '../Constants/Types';
+import {
```
msnyder-msft, line 39: can't we just pull out this util to the same place if it's for this component?

msnyder-msft:
can you please update the screeshots in the description before completing?

msnyder-msft (approved):

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.styles.ts
```diff
@@ -76,17 +84,24 @@ export const getStyles = (): IDataHistoryWidgetStyles => {
             quickTimePicker: {
                 dropdown: {
                     '.ms-Dropdown-title': {
-                        borderWidth: 0
+                        borderWidth: 0,
+                        padding: '0px 4px',
+                        height: 24,
+                        background: 'transparent',
+                        ':hover': {
```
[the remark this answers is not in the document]
iremgokce: added styling for keyboard navigation and tested

pasanchMSFT (approved):
LGTM

PR #733, opened 2022-10-21
[Data history]: Changed schema config to store adx connection info separately

iremgokce:
### Summary of changes 🔍 
- Changed schema to store the adx connection information in separate fields instead of single connection string with key and values (defined the `ITimeSeriesConnection` as connection field in the widget configuration which can be extendable easily in the future in "anyof" field if there is going to be another type of storage than `IADXTimeSeriesConnection`)
- Changed the code accordingly in data history widget related components

image
image

image

### Testing 🧪
If you already created data history widgets the config syntax validation would fail so you can reset your config file or you can delete the data history widget parts in your config file and upload back again. You can test it by creating a new data history widget in ADT3DScenePage

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
+++ b/schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
@@ -236,6 +236,30 @@
                 "time"
             ]
         },
+        "IADXTimeSeriesConnection": {
+            "type": "object",
+            "description": "Azure Data Explorer connection information for time series data",
+            "additionalProperties": false,
+            "properties": {
+                "adxClusterUrl": {
+                    "type": "string"
+                },
+                "adxDatabaseName": {
+                    "type": "string"
+                },
+                "adxTableName": {
+                    "type": "string"
+                }
+            },
+            "required": ["adxClusterUrl", "adxDatabaseName", "adxTableName"]
+        },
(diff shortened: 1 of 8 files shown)


schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
```diff
@@ -236,6 +236,30 @@
...
+                    "type": "string"
+                },
+                "adxTableName": {
+                    "type": "string"
+                }
+            },
+            "required": ["adxClusterUrl", "adxDatabaseName", "adxTableName"]
+        },
+        "ITimeSeriesConnection": {
+            "anyOf": [
```
msnyder-msft, line 257: why do we need to use `anyof` here? can't we just directly tell it the type is of `IADXTimeSeriesConnection`? Presumably this is for extensibility?

msnyder-msft (approved):
No issues from me. This looks solid.

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
iremgokce: yes I mentioned in the description, it is for extensibility in the future for any other type of connection that we might support

pasanchMSFT (approved):
LGTM

PR #737, opened 2022-10-26
Visual Rules list supports iconography for showing just badges, just mesh colorings, and both

abalogun-dev:
Then go to Behaviors> Visual Rule> you should see a ringer appear for the rule that has only alerts in it, and a cube for the rule that only has mesh coloring conditions in it. 
Viewing the icon for when both badge and mesh coloring conditions both exist in a rule will currently require you to go to one of the mock data .json files and add a value range that represents a badge. The conditions flyout is not yet built.

https://msazure.visualstudio.com/One/_workitems/edit/15913880/

### Checklist ✔️
- [ ] Added relevant labels


msnyder-msft (approved):
LGTM

msnyder-msft:
Can we add some stories for the list that include mock data for only badge, only color and then one with both? They can be a single test but we should have list items that have each of those scenarios.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
```diff
@@ -103,13 +115,37 @@ function getListItems(
         ];
     };
 
+    function getIconStart(item: IVisualRule) {
+        const [meshCount, badgeCount] = getBadgesAndMeshesCount(item);
+        let icon;
+        if (meshCount && badgeCount) {
+            icon = () => (
+                <Svg
```
pasanchMSFT, line 123: This could be a Fluent Icon or FontIcon, why are we using this new external component?

pasanchMSFT: Played around with this code a bit, and tried to do this with Fluent. This seems like a better solution and simpler to maintain.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
```diff
@@ -1,9 +1,12 @@
-import { IContextualMenuItem } from '@fluentui/react';
+import { IContextualMenuItem, Theme, useTheme } from '@fluentui/react';
+import Svg from 'react-inlinesvg';
 import React, { useEffect, useState } from 'react';
 import { TFunction, useTranslation } from 'react-i18next';
```
pasanchMSFT, line 4: Not sure if you added this `TFunction` import, but it showed me an error when I pulled this branch and looked at the code. The correct import seems to be `import { TFunction } from 'i18next';` and it is not a generic so the <string> besides it is not required.

pasanchMSFT (approved):
Code LGTM, I just noticed a faulty import when I pulled and played with it locally.

msnyder-msft (approved):

PR #740, opened 2022-10-26
CardboardList: Add id for list item

msnyder-msft:
### Summary of changes 🔍 
Adding an id prop explicitly for the list item 

### Testing 🧪

### DIFF
--- a/src/Components/CardboardList/CardboardListItem.tsx
+++ b/src/Components/CardboardList/CardboardListItem.tsx
@@ -19,6 +19,7 @@ export const CardboardListItem = <T extends unknown>(
         buttonProps,
         iconEnd,
         iconStart,
+        id,
         index,
         item,
         itemType,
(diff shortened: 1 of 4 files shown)


src/Components/CardboardList/CardboardListItem.tsx
```diff
@@ -102,6 +103,7 @@ export const CardboardListItem = <T extends unknown>(
                     {...buttonProps}
                     key={`cardboard-list-item-${listKey}-${index}`}
                     data-testid={`cardboard-list-item-${listKey}-${index}`}
+                    id={id}
```
pasanchMSFT, line 106: I think this should follow the same pattern as key and testid. with listKey and index.

pasanchMSFT (approved):

abalogun-dev (approved):

PR #742, opened 2022-10-27
[Data history]: Fixed data grouping for large data set

iremgokce:
### Summary of changes 🔍 
It seems data grouping was not working due to wrong import (should have been Highcharts from `highcharts/highstock` package). Fixed that and added a story for large data set to test it.

As a note, we can set the [pixel with for the data points of each data group](https://api.highcharts.com/highstock/plotOptions.series.dataGrouping.groupPixelWidth) if we want to have more granular control in plot in the future based on user feedback using `groupPixelWidth`.

### Testing 🧪
Use `LargeDataSet` story under HighChartsWrapper component which uses 10,000 data points. You can test the data grouping by setting the enabled prop in the highchart options in HighChartsWrapper component to `true` or `false` (`plotOptions.series.dataGrouping.enabled`). You can also change the number of data points in the mock data of `HighChartsMockLargeData` to see that [data grouping](https://www.highcharts.com/docs/stock/data-grouping) is getting only activated when necessary by the library component to increase the performance.

image
image
image

### Checklist ✔️
- [x] Added relevant labels


src/Components/HighChartsWrapper/HighChartsWrapper.types.ts
```diff
@@ -109,3 +110,11 @@ export const HighChartsMockData: Array<IHighChartSeriesData> = [
         tooltipSuffix: 'm'
     }
 ];
+
+export const HighChartsMockLargeData: Array<IHighChartSeriesData> = [
```
msnyder-msft, line 114: can we please pull out mock data to a .mock file so it's not in our production code base?

src/Components/HighChartsWrapper/HighChartsWrapper.tsx
```diff
@@ -7,13 +7,14 @@ import {
...
 import { classNamesFunction, useTheme, styled } from '@fluentui/react';
-import Highcharts, {
+import {
     AlignValue,
     ColorString,
     DataGroupingApproximationValue,
     OptionsLayoutValue,
     SeriesOptionsType
 } from 'highcharts';
+import Highcharts from 'highcharts/highstock';
```
msnyder-msft, line 17: it seems kind of odd that we wouldn't get the same thing here as we do from the root. What is the difference with the highstock folder? I saw your description in the PR but why do they have 2 versions of their library in the same package?

msnyder-msft (approved):
Please move the test data out of the production files and into a .mock file

src/Components/HighChartsWrapper/HighChartsWrapper.tsx
iremgokce: it is not two versions, they are different libraries under the same package: https://api.highcharts.com/ 
image

PR #746, opened 2022-10-28
Styling: Add extended palette to the theme

msnyder-msft:
### Summary of changes 🔍 
Adding a custom palette extension to the base theme. To use it, simply call `useExtendedTheme` instead of the fluent `useTheme` and pass that to your styles file as an `IExtendedTheme` instead of `ITheme` and you should be rocking and rolling. 

### Testing 🧪
None.

### DIFF
--- a/src/Models/Hooks/useExtendedTheme.ts
+++ b/src/Models/Hooks/useExtendedTheme.ts
@@ -1,0 +1,6 @@
+import { useTheme } from '@fluentui/react';
+import { IExtendedTheme } from '../../Theming/Theme.types';
+
+export const useExtendedTheme = () => {
+    return useTheme() as IExtendedTheme;
+};
--- a/src/Models/Hooks/useExtendedTheme.ts
+++ b/src/Models/Hooks/useExtendedTheme.ts
@@ -3,7 +3,7 @@ import { IExtendedTheme } from '../../Theming/Theme.types';
 
 /**
  * Gives back the current theme object for the app
- * @returns the curren theme object with the right colors for the selected theme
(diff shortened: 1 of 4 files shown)


src/Models/Hooks/useExtendedTheme.ts
```diff
@@ -0,0 +1,10 @@
+import { useTheme } from '@fluentui/react';
+import { IExtendedTheme } from '../../Theming/Theme.types';
+
+/**
+ * Gives back the current theme object for the app
+ * @returns the curren theme object with the right colors for the selected theme
```
pasanchMSFT, line 6: nit: the current*

pasanchMSFT (approved):

src/Models/Hooks/useExtendedTheme.ts
msnyder-msft: Fixed.

src/Theming/Theme.types.ts
```diff
@@ -0,0 +1,24 @@
+import { IPalette, IPartialTheme, Theme } from '@fluentui/react';
+
+/** custom colors our app needs */
+export interface CustomPalette {
```
iremgokce, line 4: should this interface extend from a base pallette or something? otherwise, how can we make sure that all the <key, value> pairs are valid palette variables? or should we add [key:string]: string to the interface when we need to extend the varibles like glassyBorder? I think my main question is about how we decide the names like glassyBackground75, are they coming from a base interface?

iremgokce (approved):
Left one question/comment, otherwise LGTM

PR #748, opened 2022-10-28
Design review fixes - visual rules

pasanchMSFT:
### Summary of changes 🔍 
Made changes to the design of several visual rule components based on a design review with @azimamansuri and @rinisbet.

### Testing 🧪
- Open browser dev tools and run this in console: localStorage.setItem('cardboard.feature.visualRulesFeature',true);
- Review visual rules, which can be found on the behavior form.

### Checklist ✔️
- [ ] Added relevant labels


src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionSummary.tsx
```diff
@@ -159,11 +160,13 @@ const ConditionSummary: React.FC<IConditionSummaryProps> = (props) => {
 
     return (
         <Stack tokens={STACK_TOKENS}>
-            <h4 className={classNames.title}>
-                {conditionType === 'boolean'
-                    ? t(LOC_KEYS.flyoutValueTitle)
-                    : t(LOC_KEYS.flyoutValuesTitle)}
-            </h4>
+            {!isNumericType(conditionType) && (
```
msnyder-msft, line 163: from an accessibility perspective should this be a header? I feel like a callout should have a "header" to denote the content inside. If we are changing this for styling reasons, we should just override the defaults on the headers.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -122,6 +126,38 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
         [handleDismissFlyout, onDeleteCondition, t, valueRanges]
     );
 
+    const renderBadge = useCallback((iconName: string, color: string) => {
+        return (
+            <Icon
+                iconName={iconName}
+                styles={{
+                    root: {
```
msnyder-msft, line 134: can we pull these out to the styles file?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleFormUtility.ts
```diff
@@ -36,9 +36,9 @@ export function getConditionSecondaryText(
     values: unknown[]
 ): string {
     if (type === 'NumericRange') {
-        return `${values[0]} ${i18n.t('min')} ${i18n.t('to')} ${
+        return `${values[0]} (${i18n.t('min').toLowerCase()}) ${i18n.t('to')} ${
             values[1]
-        } ${i18n.t('max')}`;
+        } (${i18n.t('max').toLowerCase()})`;
```
msnyder-msft, line 41: I just had a thought, does toLowerCase do the right thing across all languages? I guess it must, but I hadn't ever thought about that.

src/Resources/Locales/en.json
```diff
@@ -461,16 +461,16 @@
             "tabDescription": "Define widgets that appear in the element's context"
         },
         "behaviorVisualRulesTab": {
-            "tabDescription": "Set rules for applying color to certain conditions—for instance, red to indicate high temperatures, or green for optimal speed.",
+            "tabSubtitle": "Create rules for applying visuals to elements based on twin data.",
```
msnyder-msft, line 464: did you make sure to change or delete these keys once before changing the values? THat way it will pick them up and generate for all the languages. By default it will not update translations for keys that already exist.

msnyder-msft (approved):

src/Resources/Locales/en.json
pasanchMSFT: I renamed (removed) all the old keys so the new translations would be generated.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleFormUtility.ts
pasanchMSFT: Good question, I don't think JS knows which language you are inputting. I can take a look into it, but my guess is if there are no lower/upper case letters in a specific language the string won't change.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionSummary.tsx
pasanchMSFT: Actually I changed it because I believe it should not be a header. It is a label for either string or boolean input, don't think it needs to be an h4.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/VisualRules/VisualRuleList.tsx
```diff
@@ -110,7 +115,16 @@ function getListItems(
             onClick: () => onEditRule(item.id),
             overflowMenuItems: getMenuItems(item),
             textPrimary: item.displayName,
-            textSecondary: getSecondaryText(item, t)
+            textSecondary: getSecondaryText(item, t),
+            buttonProps: {
+                customStyles: {
```
iremgokce, line 120: is it possible to move this styling to style class and assign it as style or classname here instead of setting the style explicitly here?

iremgokce: if it requires to be a style object you can get advantage from subcomponents and instead of classname you can use it as style object and set here. I did it in IHighChartsWrapperStyles

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -147,6 +176,17 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
                                 ),
                                 selectedTarget: `#${LIST_KEY}`
                             });
+                        },
+                        buttonProps: {
+                            customStyles: {
+                                root: {
+                                    [primaryTextClassName]: {
+                                        fontStyle: condition.isUnlabeled
```
iremgokce, line 184: similar to my previous comment about styling, can we move this out of the styles file and use a styling from there here?

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleFormUtility.ts
```diff
@@ -36,9 +36,9 @@ export function getConditionSecondaryText(
     values: unknown[]
 ): string {
     if (type === 'NumericRange') {
-        return `${values[0]} ${i18n.t('min')} ${i18n.t('to')} ${
+        return `${values[0]} (${i18n.t('min').toLowerCase()}) ${i18n.t('to')} ${
```
iremgokce, line 39: can't we just use t from useTranslation? is the i18n.t will get the current selected language from context? It happened to me once that is why I just wanted to make sure you tried it :)

iremgokce (approved):
I left couple of comments, otherwise code LGTM

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/VisualRuleFormUtility.ts
pasanchMSFT: Looks like it is not safe (https://stackoverflow.com/questions/30762886/is-javascripts-touppercase-language-safe#:~:text=JavaScript%27s%20toUpperCase%20%28%29%20method%20is%20supposed%20to%20use,is%20supposed%20to%20be%20I%20in%20all%20cases.). I'll add locales for this.

PR #755, opened 2022-11-01
Fix target for conditions callout

pasanchMSFT:
### Summary of changes 🔍 
Fixed the target HTML element for Conditions callout. Now it targets the Cardboard list items instead of the list itself.

### Testing 🧪
Open visual rules list form, click on several Condition list items and notice how the callout changes where it is rendered.

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
+++ b/src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
@@ -45,6 +45,9 @@ const defaultCalloutInfo: CalloutInfo = {
     selectedTarget: ''
 };
 
+const CONDITIONS_ITEM_ID_PREFIX = 'cb-visual-rule-conditions-list-item-';
+const CONDITIONS_ADD_BUTTON_ID = 'cb-visual-rule-add-condition-button';
+
 const ConditionsList: React.FC<IConditionsListProps> = (props) => {
     // Props
     const {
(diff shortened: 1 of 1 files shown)


src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -77,7 +80,7 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
                 valueRangeType,
                 getNextColor(valueRanges, defaultSwatchColors)
             ),
-            selectedTarget: `#${LIST_KEY}`
+            selectedTarget: `#${CONDITIONS_ADD_BUTTON_ID}`
```
msnyder-msft, line 83: any reason we don't do the `useId` thing? I don't know why that is beneficial but we do it most places.

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
```diff
@@ -99,7 +102,9 @@ const ConditionsList: React.FC<IConditionsListProps> = (props) => {
                         selectedCondition: valueRanges.find(
                             (vr) => vr.id === conditionId
                         ),
-                        selectedTarget: `#${LIST_KEY}`
+                        selectedTarget: `#${
+                            CONDITIONS_ITEM_ID_PREFIX + conditionId
```
msnyder-msft, line 106: maybe just have a util that takes in the condition id and gives back a string? then you don't need the constant at module scope cause all the places can just call that util. And it means the generation logic is all in one spot.

`getListItemId`

msnyder-msft (approved):

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsList.tsx
iremgokce: yeah we use useId for target stuff which ensures uniqueness of the generated string

montgomp (approved):

PR #763, opened 2022-11-03
[Data history]: Error handling

iremgokce:
### Summary of changes 🔍 
- Added internal error handling component of `DataHistoryWidgetErrorHandling` for the widget. Currently surfacing the following errors:
  - ADX connection information (cluster url, database)
  - query (including table since it is part of the query in the request)
  - permission (was not able to test it with any of the clusters since I have inherited contributor/owner or reader access always by subscription or group assignments)
- Changed useTimeSeriesData hook to surface erros
- Added go.microsoft link for data history document 
- Added some granular changes in the HighchartWrapper component
 -  Added [groupPixelWidth ](https://api.highcharts.com/highstock/plotOptions.series.dataGrouping.groupPixelWidth)to control the size of the buckets when dataGrouping is active, the default it was 2px, but increased it since we have small space
 - Added tooltip decimal back again since it is used when data grouping is active. But note that, it is adding .00 even it is zero, that was the reason I had removed this option before but forgot about data grouping is active, since data is changing - as I was already rounding data point values by 2 decimals.
 - Added rangeSelector:false just in case even if it was not rendered in our widget. Since it is enabled by default, I wanted to disable it
 - Added format for tooltip timestamp to control milliseconds with 2 digits and added AM/PM. Feel free to give suggestion for the datetime format in the tooltip.

_References:_
- https://learn.microsoft.com/en-us/azure/data-explorer/kusto/api/netfx/kusto-data-client-errors
- https://learn.microsoft.com/en-us/azure/data-explorer/error-codes

**TODO:**
Content is subject to change.

### Testing 🧪
Test it by changing your 3d config file connection information one by one and go to viewer:
image
image
image
image
image
image

or you can use mock stories under Components/Behaviors Modal/Data History

### Checklist ✔️
- [x] Added relevant labels


msnyder-msft:
Can we make the flyout fit the content? or at least scale a little bigger than this? It seems like the "View more" flyout should be able to show teh full contents and not then require additional scrolling. The idea here is to be able to screenshot the page and send it to someone to help fix the configuration or file a bug, so the more we can fit on the screen the better.

image

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.tsx
```diff
@@ -212,32 +212,46 @@ const DataHistoryWidget: React.FC<IDataHistoryWidgetProps> = ({
...
-                )}
-            </div>
+            {errors.length > 0 ? (
+                <>
+                    <div className={classNames.header}>
+                        <span className={classNames.title}>{displayName}</span>
+                    </div>
+                    <DataHistoryWidgetErrorHandling
+                        errors={errors}
+                        styles={styles}
```
msnyder-msft, line 222: why do we pass the entire styles object? SHouldn't this be a subcomponent style so that it only contains teh styles for the child?

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.types.ts
```diff
@@ -31,6 +36,17 @@ export interface IDataHistoryWidgetSubComponentStyles {
...
     };
+    errorContainer: {
+        root: CSSProperties;
+        image: Partial<IImageStyles>;
+        textContainer: IStyle;
+        descriptionContainer: IStyle;
+    };
+}
+
+export enum DataHistoryErrors {
```
msnyder-msft, line 48: can we call these "error codes" `DataHistoryServiceErrorCodes` maybe?

msnyder-msft: they probably belong in the constants file too rather than a types file.

msnyder-msft: That assumes these are used to look up the locale string for the corresponding error from the service.

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/Internal/DataHistoryWidgetErrorHandling.tsx
```diff
@@ -0,0 +1,188 @@
...
+                        styles={{
+                            ...sharedCalloutStyles,
+                            root: {
+                                ...(sharedCalloutStyles.root as any),
+                                fontSize: 12
+                            }
+                        }}
+                        target={`#${errorDetailsCalloutId}`}
+                        onDismiss={toggleIsDetailsCalloutVisible}
+                        role="alert"
```
msnyder-msft, line 123: what does setting this role do? I haven't seen us doing this much in our repo, what made you decide to do it here?

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/Internal/DataHistoryWidgetErrorHandling.tsx
```diff
@@ -0,0 +1,188 @@
...
+    const errorDetailsCalloutId = useId('show-error-details-button');
+    const errorDetailsCalloutDescriptionId = useId(
+        'error-details-callout-description'
+    );
+    const errorDetailsCallout = useMemo(() => {
+        if (response?.data?.error.innererror?.message) {
+            return (
+                isDetailsCalloutVisible && (
+                    <Callout
+                        ariaDescribedBy={errorDetailsCalloutDescriptionId}
```
msnyder-msft, line 113: nice attention to detail here.

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/Internal/DataHistoryWidgetErrorHandling.tsx
```diff
@@ -0,0 +1,188 @@
...
+} from '@fluentui/react';
+import { useBoolean, useId } from '@fluentui/react-hooks';
+import React, { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+import ConnectionErrorImg from '../../../../../../Resources/Static/connectionError.svg';
+import PermissionErrorImg from '../../../../../../Resources/Static/priviledgedAccess.svg';
+import GenericErrorImg from '../../../../../../Resources/Static/noResults.svg';
+import { DOCUMENTATION_LINKS } from '../../../../../../Models/Constants/Constants';
+import { IComponentError } from '../../../../../../Models/Constants/Interfaces';
+import { getCardboardListCalloutStyles } from '../../../../../CardboardListCallout/CardboardListCallout.styles';
```
msnyder-msft, line 16: why do we bring these styles from some other component? If they need to be generic shouldn't we pull them out to a shared file instead of referencing another component directly? If we make chagnes there I would never think to come look here to see if I broke something.

src/Resources/Locales/en.json
```diff
@@ -671,6 +673,13 @@
                         "information": "When there are too many datapoints to display, values will be grouped in the chart. Choose the aggregation method for these groupings."
                     }
                 }
+            },
+            "errors": {
+                "genericNetwork": "Cluster url is invalid or network error occured.",
```
msnyder-msft, line 678: Did we do a review on these with @rinisbet? They mostly look fine but I notice we don't ever give them direction on how to resolve the issue.

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.tsx
```diff
@@ -47,8 +47,9 @@ import {
     IDataHistoryWidgetStyleProps,
     IDataHistoryWidgetStyles
 } from './DataHistoryWidget.types';
+import { DataHistoryWidgetErrorHandling } from './Internal/DataHistoryWidgetErrorHandling';
 
-const getClassNames = classNamesFunction<
```
msnyder-msft, line 51: can we add a story or two to cover the different error cases?

msnyder-msft (approved):
I didn't actually try out the experience but this looks pretty good. Let's just add a couple stories for the error cases since those are tricky to test manually and then left a few comments about imports and such. Great work!

iremgokce:
> Can we make the flyout fit the content? or at least scale a little bigger than this? It seems like the "View more" flyout should be able to show teh full contents and not then require additional scrolling. The idea here is to be able to screenshot the page and send it to someone to help fix the configuration or file a bug, so the more we can fit on the screen the better.
> 
> image

This flyout fit the content in English language for all the error cases that we want to surface. This screenshot was an example of how it works with overflow content in case other languages are used.

msnyder-msft:
> > Can we make the flyout fit the content? or at least scale a little bigger than this? It seems like the "View more" flyout should be able to show teh full contents and not then require additional scrolling. The idea here is to be able to screenshot the page and send it to someone to help fix the configuration or file a bug, so the more we can fit on the screen the better.
> > image
> 
> This flyout fit the content in English language for all the error cases that we want to surface. This screenshot was an example of how it works with overflow content in case other languages are used.

Is there a reason it can't work across languages? Why not just let it expand?

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.types.ts
iremgokce: I changed the name, but will keep it here instead of root constants file, since this enum is only used by this component.

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/Internal/DataHistoryWidgetErrorHandling.tsx
iremgokce: it was in fluent example for the Callout, it sends the alert to the screen reader and the text content is being read automatically when the callout is opened.

msnyder-msft: very cool. Great find.

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/Internal/DataHistoryWidgetErrorHandling.tsx
iremgokce: copied the used styling to the shared file and importing from there instead.

src/Resources/Locales/en.json
iremgokce: They will do content pass on these errors. We put learn more link which points to the data history documentation. Not sure how much granular documentation we need here, I am sure @rinisbet will share his feedback on that.

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.tsx
iremgokce: ahh yeah, removed this and reading styles from parents subcomponent

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.tsx
iremgokce: added stories for each surfaced error under Components/Behaviors Modal/Data History

src/Components/BehaviorsModal/Internal/Widgets/DataHistoryWidget/DataHistoryWidget.stories.tsx
```diff
@@ -55,3 +60,103 @@ DataHistoryWidgetInPreviewMode.args = {
     mode: BehaviorModalMode.preview,
     activeWidgetId: 'b2cf0aa2415f4e029f27f957fb559db2'
 };
+
```
msnyder-msft, line 63: great stuff. Thanks for adding these failure cases.

msnyder-msft (approved):

src/Resources/Locales/en.json
```diff
@@ -671,6 +673,13 @@
...
+                "genericBadRequest": "Bad request exception occured.",
+                "generalBadRequestMessage": "Request is invalid and cannot be executed.",
+                "entityNotFound": "Database not found, failed to execute query.",
```
msnyder-msft, line 681: suggestion: _failed to execute the query_

pasanchMSFT (approved):
Great stuff, very clean!

PR #771, opened 2022-11-09
Added string to list of allowed property schemas in timeseries builder 

WaywardHayward:
### Summary of changes 🔍 

In some environments numeric values are streamed into string fields, for example they stream "N/A" sometimes - or if they use something like the isa-95 ontology equipment property values are all strings

In order to make it easier for them to use the timeseries widget - this change allows them to select "string" properties as well as numeric ones,

The ADX query has then been updated to filter out non-numeric values.

Also added a warning note to a timeseries field when the type is not numeric something like 

``` Note: Non-numeric properties will be cast to numeric at query time, and may result in some values being omitted```

To do this I had to add a nullable field 'PropertyType' to the IDataHistoryBasicTimeSeries type.

Without warning 

image

With warning
image

### Testing 🧪

1. Open the /pages-adt3dscenepage--adt-3-d-scene-page-card storybook 
2. Edit a behavior 
3. Add a timeseries widget 
4. The Properties dropdown should allow you to select "string" schema'd properties also

#### ADX Queries

I tested the query with a twin in the ISS adt Instance which has a string property "Value" in the ADT Instance.

1. Write a numeric value 1 into the Value Field
2. Write a string value "Fred" into the value field
3. Write a series of numeric values

The graph in the viewer mode should show the timeseries points without the "Fred" entry.

I also tested the query in adx:

``` 
adt_dh_ah_iss_adt_02_weu
| where TimeStamp > ago(3d)
| where Id == 'NODE2000005' and Key == 'Value'
| order by TimeStamp asc 
| project TimeStamp, Id, Key, Value

```
This brings back ALL the values in the timeseries regardless of type.

```
adt_dh_ah_iss_adt_02_weu
| where TimeStamp > ago(3d)
| where Id == 'NODE2000005' and Key == 'Value'
| extend Value=todouble(Value)
| where isnotnull(Value)
| order by TimeStamp asc 
| project TimeStamp, Id, Key, Value
```

This brings back the timeseries values **Excluding** the entry for "Fred"

### Checklist ✔️
- [x] Linked associated issue (if present) #770 
- [x] Added relevant labels

### CODE REVIEW

File: schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json [hunk 1]
```diff
@@ -271,6 +271,9 @@
                 "expression": {
                     "type": "string"
                 },
+                "propertyType": {
```
iremgokce, line 274: I think with this change `propertyType` should be listed in the required field in L#281 below and we should generate the types from config again. Because it seems we set the propertyType always in the `TimeSeriesFormCallout` with the `expression.property.propertyType`.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesFormCallout.tsx
```diff
@@ -151,8 +160,20 @@ const TimeSeriesFormCallout: React.FC<IProp> = ({
                     propertyExpression={{
                         expression: seriesToEdit?.expression
                     }}
+                    description={
+                        seriesToEdit?.propertyType &&
```
iremgokce, line 164: if we make the propertyType required as I suggested in my previous comment, lets remove this  `seriesToEdit?.propertyType` check from here.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesFormCallout.tsx
```diff
@@ -151,8 +160,20 @@ const TimeSeriesFormCallout: React.FC<IProp> = ({
...
+                        !numericPropertyValueTypes.includes(
+                            seriesToEdit.propertyType
+                        )
+                            ? t(
+                                  'widgets.dataHistory.form.timeSeries.nonNumericWarning'
+                              )
+                            : ''
```
iremgokce, line 171: else condition can be undefined instead of empty string, maybe?

src/Models/Hooks/useTimeSeriesData.ts
```diff
@@ -150,6 +150,8 @@ const getBulkADXQueryFromTimeSeriesTwins = (
         twins?.forEach((twin, idx) => {
             query += `${connection.kustoTableName} | where TimeStamp > ago(${agoTimeInMillis}ms)`;
             query += ` | where Id == '${twin.twinId}' and Key == '${twin.twinPropertyName}'`;
+            query += ` | extend Value = todouble(${ADXTableColumns.Value})`;
+            query += ` | where isnotnull(${ADXTableColumns.Value})`;
```
iremgokce, line 154: I have couple of questions here:
1. why do we filter out the null values? Because it is important for chart rendering like to connect line between null values or how to show it in chart in general. So I am worrying that even if it is null it is still valid data for us? @darsney ?
2. should we change the order here if we use this null check like `| where isnotnull(${ADXTableColumns.Value})` is first then `| extend Value = todouble(${ADXTableColumns.Value})`. But if you try to catch values which are null after the todouble operation I think we should use the `Value` column instead?
3. Why do we extend table with a column named `Value`? I didn't see we use it in the project below? Was it intended?

iremgokce:
Hey, thanks @WaywardHayward for picking this feature up! Code looks good to me but I added couple of feedback&questions.

src/Models/Hooks/useTimeSeriesData.ts
WaywardHayward: I’ll check in the am but I think todouble on a non numeric returns null so depending on if you want to filter out nulls - you have to do it after - not sure what todouble on a null does

msnyder-msft: For #2 I think that makes sense to exclude the null values because if they send N/A and expect taht to count as "null" do we want to convert it to 0 or just no data? I feel like the chart would look very different for those two cases. If they send 0 and we parse it, then it would be 0 and we would get that value, so I feel like this _should_ only impact the string scenarios and this logic makes sense to me.

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
WaywardHayward: done

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesFormCallout.tsx
WaywardHayward: agreed

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
msnyder-msft: @iremgokce This is only used in your history feature right? So it won't break existing configs that already exist. If we have any configs out there that don't have this value then it will fail to validate and crash on mount of the environment so we just have to be aware of that if we want to make it required.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/Internal/TimeSeriesFormCallout.tsx
msnyder-msft: true, why not just alwyas have a value and then check if it's string since that's the only time we want to show this?

src/Resources/Locales/en.json
```diff
@@ -654,6 +654,7 @@
                     "labelPlaceholder": "Enter label to show in chart",
                     "add": "Add time series",
                     "edit": "Edit time series",
+                    "nonNumericWarning": "Note: Non-numeric properties will be cast to numeric at query time, and may result in some values being omitted",
```
msnyder-msft, line 657: shouldn't this be specific to strings since we only allow numeric and string? "Note: string properties will be cast to numeric at query...."

msnyder-msft: I appreciate that it would flex to other types if we add them, but I can't imagine we want to support booleans and enums here so being specific seems pretty safe?

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
iremgokce: this is only the data history widget part

src/Models/Hooks/useTimeSeriesData.ts
iremgokce: cool, lets check what happens when the value is already `null`? Otherwise we will get bad request error due to query if `todouble(null)` throws error

WaywardHayward: If the value is already null todouble does not throw an error and the result remains null.

iremgokce (approved):
LGTM now, thanks for your contribution!

msnyder-msft (approved):

schemas/3DScenesConfiguration/v1.0.0/3DScenesConfiguration.schema.json
iremgokce: so just a note though. Since _now_ the propertyType is a required field, it will break the config files with existing data history widget in it, fyi

PR #773, opened 2022-11-10
Array support

montgomp:
image

### Testing 🧪

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
+++ b/src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
@@ -43,8 +43,10 @@ export const AdtTwin = (args, { globals: { theme, locale } }) => {
 AdtTwin.argTypes = {
     twinId: {
         control: { type: 'text' },
-        //        defaultValue: 'PasteurizationMachine_A01'
-        defaultValue: 'ezArray' //revert this
+        defaultValue: 'PasteurizationMachine_A01'
+        // Array testing
+        // defaultValue: 'widget1' //revert this
+        // defaultValue: 'ezArray' //revert this
     }
 };
 
--- a/src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
+++ b/src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
@@ -43,7 +43,8 @@ export const AdtTwin = (args, { globals: { theme, locale } }) => {
 AdtTwin.argTypes = {
     twinId: {
         control: { type: 'text' },
-        defaultValue: 'PasteurizationMachine_A01'
+        //        defaultValue: 'PasteurizationMachine_A01'
+        defaultValue: 'ezArray' //revert this
     }
 };
 
--- a/src/Components/PropertyInspector/PropertyInspectoryModel.ts
+++ b/src/Components/PropertyInspector/PropertyInspectoryModel.ts
@@ -77,6 +77,7 @@ abstract class PropertyInspectorModel {
             case dtdlPropertyTypesEnum.float:
                 return '';
             case dtdlPropertyTypesEnum.Array:
+                return [];
             default:
                 return null;
         }
@@ -100,7 +101,8 @@ abstract class PropertyInspectorModel {
         propertySourceObject,
         mapInfo = null,
         forceSet = false,
-        schemas
+        schemas,
+        isArrayChild = false
     }: {
         modelProperty: DtdlProperty;
         propertySourceObject: Record<string, any>;
@@ -111,15 +113,23 @@ abstract class PropertyInspectorModel {
         mapInfo?: { key: string };
         forceSet?: boolean;
         schemas?: DtdlInterfaceSchema[];
+        isArrayChild?: boolean;
     }): PropertyTreeNode => {
         // Check if property is using reusable schema from schemas set
+        if (isArrayChild) {
+            // eslint-disable-next-line no-debugger
+            debugger;
+        }
         if (
             typeof modelProperty.schema === 'string' &&
             Array.isArray(schemas) &&
             schemas.map((s) => s['@id']).includes(modelProperty.schema)
         ) {
             const targetSchema = schemas.find(
-                (s) => s['@id'] === modelProperty.schema
+                (s) =>
+                    s['@id'] ===
+                    ((modelProperty.schema as any).elementSchema ||
+                        modelProperty.schema)
             );
 
             return PropertyInspectorModel.parsePropertyIntoNode({
@@ -137,7 +147,8 @@ abstract class PropertyInspectorModel {
                 path,
                 schemas,
                 mapInfo,
-                forceSet
+                forceSet,
+                isArrayChild
             });
         } else if (
             typeof modelProperty.schema === 'string' &&
@@ -207,7 +218,7 @@ abstract class PropertyInspectorModel {
                                           )
                                         : PropertyInspectorModel.buildPath(
                                               path,
-                                              modelProperty.name
+                                              modelProperty.name || field.name
                                           ),
                                     isObjectChild: true,
                                     isMapChild: false,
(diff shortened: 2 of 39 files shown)


src/Components/PropertyInspector/PropertyInspectoryModel.ts
```diff
@@ -344,21 +384,116 @@ abstract class PropertyInspectorModel {
...
                                           }
                                       );
                                   })
                                 : null,
                         ...(schemas && { mapSchemas: schemas })
                     };
                 }
-                case DTDLSchemaType.Array: // TODO support arrays in future
+                case DTDLSchemaType.Array: {
+                    const arrayItems: any[] = PropertyInspectorModel.getPropertyValueOrDefault(
```
msnyder-msft, line 404: do we know anything more about these objects at all? Or is `any` really the best we can do here?

src/Components/PropertyInspector/PropertyInspectoryModel.ts
```diff
@@ -344,21 +384,116 @@ abstract class PropertyInspectorModel {
...
+                        (s) =>
+                            s['@id'] ===
+                            ((modelProperty.schema as any).elementSchema ||
+                                modelProperty.schema)
+                    );
+                    return {
+                        name: modelProperty.name,
+                        displayName: PropertyInspectorModel.parsePropertyTreeDisplayName(
+                            {
+                                name: modelProperty.name,
```
msnyder-msft, line 419: why do this spread here? If we pull name off the object and then spread the object after that, do we gain anything? Why not just pass the object itself?

src/Components/PropertyInspector/PropertyInspectoryModel.ts
```diff
@@ -344,21 +384,116 @@ abstract class PropertyInspectorModel {
...
+                                ...modelProperty
+                            }
+                        ),
+                        role: NodeRole.parent,
+                        schema: dtdlPropertyTypesEnum.Array,
+                        isCollapsed: true,
+                        type: DTDLType.Property,
+                        path: mapInfo
```
msnyder-msft, line 427: what is `mapInfo`? I kind of assumed when I saw it up above that it was tied to map objects, but now we are inside of knowing the item is an array so I'm questioning that assumption.

src/Components/PropertyInspector/PropertyInspectoryModel.ts
```diff
@@ -344,21 +384,116 @@ abstract class PropertyInspectorModel {
...
+                                  path,
+                                  modelProperty.name,
+                                  modelProperty.index
+                              ),
+                        parentObjectPath: isObjectChild && path,
+                        isInherited,
+                        isMapChild,
+                        isArrayItem,
+                        childSchema: modelProperty.schema.elementSchema,
+                        isRemovable: !(isMapChild || isArrayItem),
```
msnyder-msft, line 442: don't we know that this isn't a map now that we are inside teh Array case condition?

src/Components/PropertyInspector/PropertyTree/TreeComponents/TreeNode.tsx
```diff
@@ -24,7 +25,12 @@ const TreeNode: React.FC<NodeProps> = ({ node }) => {
                     </div>
                 )}
                 <TreeNodeMapTool node={node} />
```
msnyder-msft, line 27: should we only show the Array tools when it's an array type? I see we hide the other control based on type here so maybe it makes sense to do teh show/hide for both at this level?

src/Components/PropertyInspector/PropertyTree/TreeComponents/TreeNodeArrayItemTool.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const handleClearArray = () => {
+        onClearArray(node);
+    };
+    if (readonly) {
+        return null;
+    }
+
+    if (node.schema === 'Array') {
+        const clearButton = (
+            <div
```
msnyder-msft, line 40: can we make this a button instead of a div?

src/Components/PropertyInspector/PropertyTree/TreeComponents/TreeNodeArrayItemTool.tsx
```diff
@@ -0,0 +1,116 @@
...
+        }
+        return (
+            <>
+                {addButton}
+                <TreeNodeSetUnset node={node} />
+            </>
+        );
+    }
+    if (!node.isArrayItem) return null;
+    return (
```
msnyder-msft, line 93: I'm a bit confused why we have 3 paths here, It seems the first is handling the array case, the second is if it's not an array and then I'm wondering how we ever get to the final case. Can we add some comments that explains that for future me?

src/Components/PropertyInspector/StandalonePropertyInspector.state.ts
```diff
@@ -234,6 +252,115 @@ const StandalonePropertyInspectorReducer = produce(
...
+
+                const originalNode = PropertyInspectorModel.findPropertyTreeNodeRefRecursively(
+                    draft.originalPropertyTreeNodes,
+                    arrayNode.path
+                );
+
+                arrayTreeNode.children.splice(0, arrayTreeNode.children.length);
+
+                // Remove all edit status flags for map children
+                Object.keys(draft.editStatus).forEach((key) => {
```
msnyder-msft, line 354: do we have maps here? Is this in case it's an array of maps?

src/Components/PropertyInspector/StandalonePropertyInspector.tsx
```diff
@@ -185,6 +204,9 @@ const StandalonePropertyInspector: React.FC<StandalonePropertyInspectorProps> =
                         onNodeValueUnset={onNodeValueUnset}
                         onAddMapValue={onAddMapValue}
                         onRemoveMapValue={onRemoveMapValue}
+                        onRemoveArrayItem={onRemoveArrayItem}
```
msnyder-msft, line 207: it's a bummer that we have to duplicate all the callbacks just to add a type. Presumably it's a lot of surgery to simply pass back a type field on the existing callbacks and rename them be generic so it would scale to the next type we decide to add?

msnyder-msft (approved):
I'm certainly no expert in this domain but the code looks alright. I left a few comments. 
Hopefully we got a chance to try it with a real environment to make sure we aren't just living in the glorious happy path of mock data.

src/Components/PropertyInspector/PropertyInspectoryModel.ts
montgomp: Could be anything. Not just the DTDL primitives but complex object (maybe not even match the defined DTDL schema).

src/Components/PropertyInspector/PropertyInspectoryModel.ts
montgomp: I was having issues with the transpiler not understanding that `name` is a property on `modelProperty`. It seems more happy with passing it directly now. Will remove the extra ops.

src/Components/PropertyInspector/PropertyInspectoryModel.ts
montgomp: Good point. This'll always be falsey. Will remove.

src/Components/PropertyInspector/PropertyInspectoryModel.ts
montgomp: Good point. This'll always be false.

src/Components/PropertyInspector/PropertyTree/TreeComponents/TreeNode.tsx
montgomp: I'm hiding the Unset here because it has "special" visibility rules for array types. I'll put a ternary operator on it.

src/Components/PropertyInspector/PropertyTree/TreeComponents/TreeNodeArrayItemTool.tsx
montgomp: The first (`node.schema === 'Array'`) is for the setting the array root to empty. The final return is for removing the individual array item. I'll reconfigure the conditionals so they make more sense (and put in some comments).

src/Components/PropertyInspector/StandalonePropertyInspector.state.ts
montgomp: I think the confusion here is in the comment. The draft.editStatus is a map of the edited paths so wee need to clear the flagged child array items from the map when clearing the root.

src/Components/PropertyInspector/StandalonePropertyInspector.tsx
montgomp: Addressed comments and made some additional div-->IconButton migrations.

src/Components/PropertyInspector/PropertyInspector.stories.local.tsx
```diff
@@ -44,6 +44,9 @@ AdtTwin.argTypes = {
     twinId: {
         control: { type: 'text' },
         defaultValue: 'PasteurizationMachine_A01'
+        // Array testing
```
msnyder-msft, line 47: do we want another story that actually has these enabled?

msnyder-msft (approved):

darsney (approved):
LGTM, lets just make sure we didnt actually do anything that makes those tests break by changing mock data or something, then send it!

PR #774, opened 2022-11-11
Create flighted Visual rule viewer experience

pasanchMSFT:
### Summary of changes 🔍 
Added flighted capabilities to show Visual rules in viewer experience. With current mock data all should run smoothly. Since this functionality is still flighted and not forward compatible, I won't be able to update mock data until I remove flighting. Which will be done in next PR.

This PR is mostly duplicating and slightly modifying existing components that rendered the Elements list and Behavior section to be able to support Visual rules instead of Alerts/Status. Terminology for Alerts/Status is removed in the new components but is still around on the codebase, will target to remove that next PR. 

Since this PR contains a lot of duplicate components, the types and styles files haven't been updated to match the V2 name. Naming updates are targeted for next PR.

Here's a screenshot of all places that contain changes:
image
- Left navbar, shows callout when color pills are hovered over. (Tooltip host cannot be added to a Cardboard list item).
- Small modal when hovering over a mesh now shows improved color pills.
- Right modal, shows updated color pills and callout shows up when row is hovered.
- All three spots support and correctly render several badges.

### Testing 🧪
- Contact me to get a modified Mock data file to be able to see the changes more clearly.
- Open browser dev tools and run this in console: `localStorage.setItem('cardboard.feature.visualRulesFeature',true);`
- Interact with objects to see their new behavior.
  - Change twin properties to be able to see how badges/colors get updated.

### Checklist ✔️
- [x] Added relevant labels


src/Components/ElementsPanel/Internal/VisualRuleElementsList.tsx
```diff
@@ -0,0 +1,274 @@
...
+            visualRuleDisplayTitle: string;
+        }> = [];
+        let visualRules: IExpressionRangeVisual[] = [];
+
+        panelItem.behaviors.map((b) => {
+            // Add all visual rules to an array
+            visualRules = b.visuals.filter(ViewerConfigUtility.isVisualRule);
+
+            // Separate into their respective arrays
+            visualRules.forEach((vr) => {
```
msnyder-msft, line 122: sooooo many nested loops here. This is going to get realllly bogged down on large scenes.

src/Components/ElementsPanel/Internal/VisualRuleElementsList.tsx
```diff
@@ -0,0 +1,274 @@
...
+    const tooltipId = useId('cb-element-coloring-header-tooltip');
+    const theme = useExtendedTheme();
+
+    if (!props.isModal) {
+        const tooltipContent = (colorings: VisualColorings[]) => {
+            return <ColorPillsTooltip visualColorings={colorings} />;
+        };
+
+        return (
+            <TooltipHost
```
msnyder-msft, line 244: don't you have a ColorPillTooltip component? Why do we need this copied here as well as somewhere else earlier in the PR? They seem like nearly copies.

src/Components/ElementsPanel/Internal/VisualRuleElementsList.tsx
```diff
@@ -0,0 +1,274 @@
...
+                id={tooltipId}
+                tooltipProps={{
+                    onRenderContent: () => tooltipContent(props.colorings),
+                    targetElement: document.getElementById(props.rowId)
```
msnyder-msft, line 248: do you have to pass the entire object here? Can't you give it an id to bind to?

src/Components/ElementsPanel/Internal/VisualRuleElementsList.tsx
```diff
@@ -0,0 +1,274 @@
...
+    return listItems;
+}
+
+interface IElementColoringProps {
+    colorings: VisualColorings[];
+    isModal: boolean;
+    rowId: string;
+}
+
+const ElementColoring: React.FC<IElementColoringProps> = (props) => {
```
msnyder-msft, line 234: i'd rather see each of these components in their own file. IT seems like a lot is packed into a single file here.

src/Components/ElementsPanel/Internal/VisualRuleElementsList.tsx
```diff
@@ -0,0 +1,274 @@
...
+        return (
+            <TooltipHost
+                id={tooltipId}
+                tooltipProps={{
+                    onRenderContent: () => tooltipContent(props.colorings),
+                    targetElement: document.getElementById(props.rowId)
+                }}
+                directionalHint={DirectionalHint.rightCenter}
+                delay={TooltipDelay.zero}
+                calloutProps={{
```
msnyder-msft, line 252: please pull the styles out to the styles file.

src/Components/ElementsPanel/ViewerElementsPanel.types.ts
```diff
@@ -22,6 +22,7 @@ export interface IViewerElementsPanelProps {
 export interface IViewerElementsPanelListProps {
     isLoading: boolean;
     panelItems: Array<IViewerElementsPanelItem>;
+    isModal?: boolean; // TODO: Make this required when deprecated Elements list is removed
```
msnyder-msft, line 25: can we add a comment to this prop on when it should be true? I don't quite know just from the name. Is it that the list is in a modal or that the contents should be a modal? 🤔

src/Components/StatusPills/ColorPills.tsx
```diff
@@ -0,0 +1,63 @@
...
+            verticalAlign={'center'}
+            styles={{ root: styles.root }}
+        >
+            {trimmedVisualColorings.map((vc, idx) => {
+                const { color } = vc;
+                const pillStyles = getPillStyles(color);
+
+                return (
+                    <div
+                        key={`${pillId}-${idx}`}
```
msnyder-msft, line 51: what do we need to do here for accessibility (mainly screen readers)? I don't know what the guidance is since this is a color. It's not like we can read the hex code out. 😟

src/Components/StatusPills/ColorPills.tsx
```diff
@@ -0,0 +1,63 @@
...
+                        className={pillStyles.statusColorPill}
+                    ></div>
+                );
+            })}
+            {extraVisualColoringCount > 0 && (
+                <span
```
msnyder-msft, line 57: tip: if you don't want to use styles here, you can use the `<Text>` component and just pass the variant for font size directly to the component and it will do the sizing for you.

src/Components/ColorPillsTooltip/ColorPillsTooltip.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { useId } from '@fluentui/react-hooks';
+import { useTranslation } from 'react-i18next';
+import { useExtendedTheme } from '../../Models/Hooks/useExtendedTheme';
+
+const getClassNames = classNamesFunction<
+    IColorPillsTooltipStyleProps,
+    IColorPillsTooltipStyles
+>();
+
+const ColorPillsTooltip: React.FC<IColorPillsTooltipProps> = (props) => {
```
msnyder-msft, line 18: I would probably rethink the naming of this component, it doesn't actually have to be in a tooltip component. When I saw this, I thought it was a tooltip itself, but i'ts actually ColorPillsCalloutContents or something like that.

src/Components/ElementsPanel/Internal/VisualRuleElementsList.tsx
msnyder-msft, line 244: I would argue that the tooltip doesn't make sense here. I would bind this to a callout and show/hide based on the callbacks from CardboardList. The tooltip interaction here feels really weird given the location of the tooltip (showing up 300px away from where I hovered feels wrong). 
I also would expect the whole row to be the target instead of just the bars since hovering the row already has the side effect of highlighting the element in the viewer.

src/Components/StatusPills/StatusPills.styles.ts
```diff
@@ -1,4 +1,4 @@
-import { mergeStyleSets } from '@fluentui/react';
```
msnyder-msft, line 1: nit: this file is named StatusPills but the component is ColorPills

src/Components/StatusPills/StatusPills.types.ts
```diff
@@ -1,6 +1,7 @@
 import { BehaviorModalMode } from '../../Models/Constants/Enums';
 import { DTwin } from '../../Models/Constants/Interfaces';
 import { IExpressionRangeVisual } from '../../Models/Types/Generated/3DScenesConfiguration-v1.0.0';
+import { VisualColorings } from '../BehaviorsModal/Internal/BehaviorSection/BehaviorVisualRuleSection';
 
 export type StatusPillsContainerWidth = 'compact' | 'wide';
```
msnyder-msft, line 7: nit on the name of this file. It should be ColorPills. OR move your types to your own files so we can just delete the old component when you do your migration.

msnyder-msft (commented):
I think we should consider using a callout on the left panel instead of a tooltip so that we can adjust the trigger target as well as the position of the callout. 
The rest all feel like pretty minor stuff.

msnyder-msft:
Separate issue: The callout on your Condition form needs to auto set focus to the Label text field when opened using a keyboard. It's not accessible via keyboard right now since you can't get focus to transfer into it.

src/Components/StatusPills/ColorPills.tsx
```diff
@@ -0,0 +1,63 @@
...
+                : visualColorings
+        );
+        setExtraVisualColoringCount(
+            visualColorings && visualColorings.length > 3
+                ? visualColorings.length - 3
+                : 0
+        );
+    }, [visualColorings]);
+
+    return (
```
msnyder-msft, line 37: the ordering of the pills is not deterministic. I see Yellow, blue, green +1 on the elements panel and Blue, green, yellow (no plus 1???) on the modal.
image
image

msnyder-msft: ahh. it's cause there is another behavior giving it another yellow. Interesting. Ok. So that's probably fine.

src/Components/BehaviorsModal/Internal/BehaviorSection/BehaviorVisualRuleSection.tsx
```diff
@@ -0,0 +1,175 @@
...
+            <div className={styles.behaviorSection}>
+                <TooltipHost
+                    id={tooltipId}
+                    tooltipProps={{
+                        onRenderContent: () => tooltipContent(meshColorings)
+                    }}
+                    directionalHint={DirectionalHint.leftCenter}
+                    delay={TooltipDelay.zero}
+                    calloutProps={{
+                        isBeakVisible: false,
```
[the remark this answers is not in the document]
pasanchMSFT: Designs don't have beak, but I could double check that with Azima

src/Components/BehaviorsModal/Internal/BehaviorSection/BehaviorVisualRuleSection.tsx
```diff
@@ -0,0 +1,175 @@
...
+        return <ColorPillsTooltip visualColorings={colorings} />;
+    };
+
+    return (
+        <>
+            <div className={styles.behaviorSection}>
+                <TooltipHost
+                    id={tooltipId}
+                    tooltipProps={{
+                        onRenderContent: () => tooltipContent(meshColorings)
```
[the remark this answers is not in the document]
pasanchMSFT: This is a different way of creating a tooltip. TooltipHost wraps around the target element and brings onHover/onFocus capabilities with it. onHover of the element being wraped around it will show the tooltip, to render custom content you need to pass the onRenderFunction.

src/Components/BehaviorsModal/Internal/BehaviorSection/BehaviorVisualRuleSection.tsx
```diff
@@ -0,0 +1,175 @@
...
+                                background: theme.palette.glassyBackground75
+                            },
+                            calloutMain: {
+                                background: 'unset',
+                                paddingRight: 24
+                            }
+                        }
+                    }}
+                >
+                    <Stack
```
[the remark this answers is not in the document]
pasanchMSFT: Full row here.

src/Components/ElementsPanel/Internal/VisualRuleElementsList.tsx
```diff
@@ -0,0 +1,274 @@
...
+                    if (
+                        shouldShowVisual(
+                            vr.valueRangeType,
+                            panelItem.twins,
+                            vr.valueExpression,
+                            condition.values
+                        )
+                    ) {
+                        if (hasBadge(condition.visual.iconName)) {
+                            badges.push({
```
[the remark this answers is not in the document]
pasanchMSFT: There are 1 or 2 places, but the badges store different information based on the component they are in. I don't think it is worth consolidating into a single class or several util functions.

src/Components/ElementsPanel/Internal/VisualRuleElementsList.tsx
pasanchMSFT: ColorPillTooltip is the content of the tooltip, I can add Content to the name if that makes it clearer.

pasanchMSFT: I'll also remove this Tooltip host in favor of the callout + onHover of Cardboardlist item

src/Components/StatusPills/ColorPills.tsx
pasanchMSFT: aria-hide this?

msnyder-msft: should we set it to aria-hidden? I don't know whether screen readers will see this since nothing actually shows up in teh UI?

pasanchMSFT: Now that I think about it, this has no content, if it had a title or some aria label it would be mentioned by screen readers, but a div with a certain background should be ignored. The colors that can be used for visual rules are limited, I could do a map of color to aria label. Sounds a bit overkill

src/Resources/Locales/en.json
```diff
@@ -1305,5 +1305,8 @@
             "Last 180 days": "Last 180 days",
             "Last year": "Last year"
         }
+    },
+    "ColorPills": {
```
iremgokce, line 1309: is there a reason that the key starts with upper case?

src/Models/SharedUtils/VisualRuleUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+        );
+    } else {
+        // Return false since other property types are not yet supported
+        return false;
+    }
+}
+
+/** Utility that returns if a Condition contains or not a badge icon */
+export const hasBadge = (condition: IValueRange): boolean => {
+    return !!(condition.visual && condition.visual.iconName);
```
iremgokce, line 40: nit
```suggestion
    return !!(condition?.visual?.iconName);
```

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ActionItem.tsx
```diff
@@ -9,7 +9,6 @@ import {
 import React, { useCallback, useEffect, useState } from 'react';
```
iremgokce, line 9: Wasn't included in the scope of this PR, but can we make the margin left of the default button same as in the other places in ConditionsCallout with 8px?

image

iremgokce: again not in the scope of this PR, but when I change the visual type to Badge coloring, even if it is empty (no badge selected), the save button is still active and when I click on save nothing happens (it returns to the state before edit - element coloring). I think we should disable save when visual type selected for badge but no icon specified, or we can select a default value for icon:
image

iremgokce: again not in the scope and maybe you already fixed in in the subsequent PR, but can we make error messages red colored as we use in different places?
image

src/Components/ColorPills/ColorPills.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const pillId = useId('cb-visual-rule-pills');
+
+    // styles
+    const classNames = getClassNames(styles, {
+        theme: useExtendedTheme(),
+        width: props.width
+    });
+
+    const [trimmedVisualColorings, setTrimmedVisualColorings] = useState(
+        visualColorings && visualColorings.length > 3
```
iremgokce, line 35:
```suggestion
        visualColorings?.length > 3
```

src/Components/ColorPills/ColorPills.tsx
```diff
@@ -0,0 +1,93 @@
...
+            ? visualColorings.slice(0, 3)
+            : visualColorings
+    );
+    const [extraVisualColoringCount, setExtraVisualColoringCount] = useState(
+        visualColorings && visualColorings.length > 3
```
iremgokce, line 40:
```suggestion
        visualColorings?.length > 3
```

src/Components/ColorPills/ColorPills.tsx
```diff
@@ -0,0 +1,93 @@
...
+            ? visualColorings.length - 3
+            : 0
+    );
+
+    useEffect(() => {
+        setTrimmedVisualColorings(
+            visualColorings && visualColorings.length > 3
```
iremgokce, line 47:
```suggestion
            visualColorings?.length > 3
```

src/Components/ColorPills/ColorPills.tsx
```diff
@@ -0,0 +1,93 @@
...
+                ? visualColorings.slice(0, 3)
+                : visualColorings
+        );
+        setExtraVisualColoringCount(
+            visualColorings && visualColorings.length > 3
```
iremgokce, line 52:
```suggestion
             visualColorings?.length > 3
```

src/Components/ColorPills/ColorPills.tsx
iremgokce, line 35: since you already updating it in useEffect down below with the changes of visualColortins, I don't think you should set a default value here - repetition of the code

src/Components/ColorPills/ColorPills.tsx
iremgokce, line 40: since you already updating it in useEffect down below with the changes of visualColortins, I don't think you should set a default value here - repetition of the code

src/Components/ColorPillsCalloutContent/ColorPillsCalloutContent.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const { t } = useTranslation();
+    const coloringIdPrefix = useId('cb-visual-coloring-callout-content');
+
+    // styles
+    const classNames = getClassNames(styles, {
+        theme: useExtendedTheme()
+    });
+
+    const getLabel = (label: string | undefined) => {
+        if (label && label.trim().length > 0) {
```
iremgokce, line 33:
```suggestion
        if (label?.trim().length > 0) {
```

src/Components/ElementsPanel/Internal/VisualRuleElementsList.tsx
```diff
@@ -0,0 +1,282 @@
...
+                            });
+                        }
+                    }
+                });
+            });
+        });
+
+        //sort the alert within its own group by name
+        badges.sort((a, b) =>
+            a.visualRuleDisplayTitle.localeCompare(
```
iremgokce, line 213: I think there is a sort method in utils, lets use it: `sortAscendingOrDescending`

src/Components/ElementsPanel/ViewerElementsPanel.tsx
```diff
@@ -24,6 +31,28 @@ const ViewerElementsPanel: React.FC<IViewerElementsPanelProps> = ({
     const [filterTerm, setFilterTerm] = useState('');
     const nodeRef = React.useRef(null); // <Draggable> requires an explicit ref to avoid using findDOMNode
 
+    const getVisualRulesIncludingFilterTerm = (panelItem) => {
+        // Map all visuals from existing behaviors
+        const visuals = [].concat(...panelItem.behaviors.map((b) => b.visuals));
```
iremgokce, line 36: why not just: 
```suggestion
        const visuals = [...panelItem.behaviors.map((b) => b.visuals)];
```

or  const visuals = panelItem.behaviors.map((b) => deepCopy(b.visuals))

src/Models/Constants/VisualRuleTypes.ts
```diff
@@ -0,0 +1,10 @@
+export interface VisualColorings {
+    color: string;
+    label: string | undefined;
```
iremgokce, line 3: can we make it optional with `label?` instead of adding `undefined `as or value?

src/Models/Constants/VisualRuleTypes.ts
```diff
@@ -0,0 +1,10 @@
...
+}
+
+export interface VisualBadges {
+    color: string;
+    iconName: string;
+    labelExpression: string | undefined;
```
iremgokce, line 9: can we make it optional with `label?` instead of adding `undefined `as or value?

src/Models/SharedUtils/VisualRuleUtils.ts
```diff
@@ -0,0 +1,41 @@
...
+    const evaluatedExpression = parseLinkedTwinExpression(
+        valueExpression,
+        twins
+    );
+    if (propertyType === 'boolean') {
+        return values[0] === evaluatedExpression;
+    } else if (propertyType === 'string') {
+        return values.includes(evaluatedExpression);
+    } else if (isNumericType(propertyType)) {
+        return ViewerConfigUtility.getValueIsWithinRange(
```
iremgokce, line 28: nit: rename isValueWithinRange

iremgokce:
Lets check out the preview mode, the number of conditions are not overlapping:
image

darsney (approved):
LGTM

iremgokce (approved):
Great work!

PR #775, opened 2022-11-12
[Azure Resource Graph]: Changes in AzureManamentAdapter and EnvironmentPicker

iremgokce:
### Summary of changes 🔍 
- Fetching flow for Azure resources in ResourcePicker is changed to using [Azure Resource Graph REST API](https://learn.microsoft.com/en-us/rest/api/azureresourcegraph/resourcegraph(2021-03-01)/resources/resources?tabs=HTTP) 
which supports [advance queries](https://learn.microsoft.com/en-us/azure/governance/resource-graph/samples/advanced?tabs=azure-cli) instead of using service provider endpoints in management API to decrease the number of calls required to fetch resources and increase performance of EnvironmentPicker
- Unused code is removed, some null checks are added
- Added `getResourceByUrl` to `AzureManagementAdapter` to fetch resources by their url strings (no longer need for id) using querying in Resource Graph (note that for container type of resources we still need to use the Storage service provider endpoint since containers are not listed [supported type for querying in Resource Graph](https://learn.microsoft.com/en-us/azure/governance/resource-graph/reference/supported-tables-resources#resources))
- In `EnvironmentPicker`, permissions check flow is removed from loading of the options (so we list all the resources no matter the required permissions, no filtering applied), but moved to as we change the selected option in the ResourcePicker dropdown we check the role assignments for given permission sets. This improved the performance a lot since we used to make thousands of calls for checking role assignments for each resource before listing in the dropdown. Not, we check role assignments as user change the options in dropdown. And whenever user don't have access to the selected resource, we show the error message down below:

image

image

image

image

**_ToDo:_** Maybe using useAdapter in EnvironmentPicker for checking permissions method and displaying the isLoading state somewhere in ResourcePicker?

### Testing 🧪
Check it with either standalone ResourcePicker component or the EnvironmentPicker in ADT3DScenePage.

### Checklist ✔️
- [x] Added relevant labels


src/Adapters/AzureManagementAdapter.ts
```diff
@@ -256,166 +254,221 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
...
+                            ?.storageAccountId ||
+                        params.searchParams?.additionalParams
+                            ?.storageAccountBlobUrl
+                    ) {
+                        const storageAccounts: Array<IAzureStorageAccount> = await this.fetchAllResources(
+                            adapterMethodSandbox,
+                            token,
+                            {
+                                type: AzureResourceTypes.StorageAccount,
+                                query: params.searchParams?.additionalParams
```
msnyder-msft, line 371: suggestion: it might be more readable to pull this out to a variable.

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -256,166 +254,221 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
...
+                                    ?.storageAccountId
+                                    ? "id == '" +
+                                      params.searchParams?.additionalParams
```
msnyder-msft, line 374: I personally find it more readable when we use the other notation for building strings with the ` marks and the ${variable}

src/Adapters/__mockData__/MockAdapterData/MockSubscriptionData.json
```diff
@@ -1,22 +0,0 @@
-[
```
msnyder-msft, line 1: does it make sense to keep the mock data for our new responses in JSON files as well? That seems like a nice way to manage it so you can more easily see and add data to the response? It has the down side of not being typed, but might still be worth it.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -113,12 +114,14 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
...
     const loadingLabelText = useMemo(() => {
-        switch (resourceType) {
-            case AzureResourceTypes.DigitalTwinInstance:
+        switch (
+            resourceType.toLowerCase() // need to compare lowercase since for the resource types that we use Resource Graph api, it returns lowercased resource type in the response object
+        ) {
+            case AzureResourceTypes.DigitalTwinInstance.toLowerCase():
                 return t('resourcesPicker.loadingInstances');
-            case AzureResourceTypes.StorageBlobContainer:
+            case AzureResourceTypes.StorageBlobContainer.toLowerCase():
```
msnyder-msft, line 122: is there a reason we don't just lowercase the enum value itself?

src/Models/Services/Utils.ts
```diff
@@ -628,11 +628,13 @@ export const getResourceUrl = (
         if (typeof resource === 'string') {
             // it means the option is manually entered using freeform
             if (resourceType) {
-                switch (resourceType) {
-                    case AzureResourceTypes.DigitalTwinInstance:
-                    case AzureResourceTypes.StorageAccount:
-                        return resource;
-                    case AzureResourceTypes.StorageBlobContainer: {
+                switch (resourceType.toLowerCase()) {
```
msnyder-msft, line 631: I see we are adding a lot of ToLowerCase on exisitng code. Why do we need to do this now?

msnyder-msft (approved):
Looks solid to me. I didn't try out the actual experience but the code looks good from what I can see. Left a few quality of life ideas but nothing major.
This is a pretty major change though, so let's identify some test cases that we should have the team bang on before we roll this out to customers.

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: yes, I kept the enum as is since **other management service provider endpoint call return the enums as it was** (including upper case) but since now the resource graph api returns all lowercase, I needed to add these lowercase comparison because we don't know the return object based on the api call in management adapter. Since Resource Graph allows querying only [certain types](https://learn.microsoft.com/en-us/azure/governance/resource-graph/reference/supported-tables-resources#resources), we still use service provider calls for containers for example where the types include upper case for initial letters.

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -157,8 +178,8 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
      *  in their list of role assignments for a particular resource */
     async hasRoleDefinitions(
         resourceId: string,
-        uniqueObjectId: string,
-        accessRolesToCheck: AzureAccessPermissionRoleGroups
+        accessRolesToCheck: AzureAccessPermissionRoleGroups,
+        uniqueObjectId: string = this.uniqueObjectId
```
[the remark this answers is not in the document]
iremgokce: uniqueObjectId is a class variable for AzureManagementAdapter that we pass in the constructor (it originally comes from auth when user logs in in 3rd party app, but in cardboard the value is from secrets file) so the default value in this method is the adapters its own variable, otherwise for different users this adapter method still can check **role assignments** for different users if the consumer of this method wants to.

src/Models/Services/Utils.ts
iremgokce: as I mentioned in my other comment, since Resource Graph returns type all lower case while other management calls return as it is described as enum in our codebase I needed to do this lowercase checks, instead I can create a util which does lowercase comparsion.

iremgokce: for example Microsoft.Authorization roleAssignments type resources returns as usual in our enum using that service provider endpoint in management api:
image

whereas Resource Graph endpoint returns:
image

src/Adapters/__mockData__/MockAdapterData/MockSubscriptionData.json
iremgokce: hmm, since we dont use other fields in the Resource Graph response I was not using new responses in separate JSON file, but just using resource objects in MockAdapter like this:
image


but I can definitely create a file like this for Resource Graph response like this:
image

okay adding it not, I think it is valuable to see the return object of this new api to see full fields in the response

msnyder-msft: makes sense. It would make even more sense if the parsing of the response was shared by the mock and real adapters but alas that is not the case in our code base. :(

src/Adapters/AzureManagementAdapter.ts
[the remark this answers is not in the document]
msnyder-msft: gotcha. it's the User's object id then. Makes sense.

darsney (approved):
This looks good to me, much better experience (and it will get a lottt better with react-select)

The only caveat is that if i have selected a container i dont have access to, then i switch to one I do, the state flashes that i dont have access until the new permissions are determined.  I think it makes more sense to unset the permissions state when the container selection box clears, that way we default to assuming permissions exist and updating if they dont, rather than erroneously showing that the dont exist from the previous state

src/Adapters/__mockData__/MockAdapterData/MockSubscriptionData.json
iremgokce: added [MockResourceGraphData.json](https://github.com/microsoft/iot-cardboard-js/pull/775/files#diff-02d1609facc206356a237324ce3005f0e735ea47a31acf2d8a8fe9d3556ddd6e)

iremgokce:
> This looks good to me, much better experience (and it will get a lottt better with react-select)
> 
> The only caveat is that if i have selected a container i dont have access to, then i switch to one I do, the state flashes that i dont have access until the new permissions are determined. I think it makes more sense to unset the permissions state when the container selection box clears, that way we default to assuming permissions exist and updating if they dont, rather than erroneously showing that the dont exist from the previous state

I made the suggested change to set the default error message state to null, but now, in between selecting **two resources that I don't have access to** the button enables for a second until checking the permission

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -256,166 +254,221 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
...
-                { apiVersion: apiVersion }
-            );
-            return new AzureResourcesData(resources);
+            let resource: IAzureResource;
+            let query = '';
+            let urlObj: URL;
+            try {
+                switch (type.toLowerCase()) {
+                    case AzureResourceTypes.DigitalTwinInstance.toLowerCase():
+                        urlObj = new URL(urlString);
```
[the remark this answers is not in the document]
iremgokce: @msnyder-msft I am thinking about adding this Utils method to make a single source of URL obj creation and handling errors and just use this anywhere we need to create a URL object and get a field we need, does it sound good?:
image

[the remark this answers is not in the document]
msnyder-msft: Sure. I dont' mind that as long as we never want to actually handle those errors. But why don't we return the URL object?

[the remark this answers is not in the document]
iremgokce: okay added the util method and updated code wherever we construct a new URL [b8ffdd6](https://github.com/microsoft/iot-cardboard-js/pull/775/commits/b8ffdd67631115e3b9e34e1f01150ea1561f5d24)

PR #781, opened 2022-11-16
Removing flight and alert/status naming

pasanchMSFT:
### Summary of changes 🔍 
This change is large, but mainly because it renames, refactors or removes some files. It is not a very complex change. Here are the details of the changes made in the PR.
Things changed:
- Imports
- Alert and Status naming throughout the app
  - Component names
  - Css classes
  - Variables
  - Functions
  - i18n
- Things that are no longer needed
  - Flighting of Visual rules
- Components removed
  - AlertsTab
  - StatusTab
  - StatusPills
- Components refactored (deleted and v2 took their place)
  - BehaviorSection
  - Elements list
- Components renamed
  - AlertsModal -> VisualsModal
- Telemetry changed
  - count of visual rules instead of status or alert visuals
- Components unchanged
  - Alert badge (could rename, but its code I haven't manipulated)
- Other changes:
  - ViewerElementPanel utils update
  - Remove isStatusColorVisual and isAlertVisual
- Tests
  - BehaviorForm stories
  - Behavior form context
- Update to Demo environment mock data and scene page mock data

### Testing 🧪
Flighting should no longer be needed
Play around on the builder and viewer with Visual rules

### Checklist ✔️
- [ ] Added relevant labels


iremgokce:
hey @pasanchMSFT it seems UI tests are failing, can you check those? I don't see any changes in stories, maybe we didn't reflect component prop changes - if happened- in stories?

pasanchMSFT:
> hey @pasanchMSFT it seems UI tests are failing, can you check those? I don't see any changes in stories, maybe we didn't reflect component prop changes - if happened- in stories?

I try those stories locally and see no issue other than model is not loading. Could that be the cause of them failing?

src/Components/ADT3DSceneBuilder/ADT3DSceneBuilder.BehaviorForm.stories.tsx
```diff
@@ -173,77 +173,77 @@ EditTwinAliasesTabAddAlias.play = async ({ canvasElement }) => {
...
 
-export const EditStatusTab = Template.bind({});
-EditStatusTab.play = async ({ canvasElement }) => {
-    await EditElementsTabSelectItem.play({ canvasElement });
-    const canvas = within(canvasElement);
-    // Finds the tabs and clicks Status
-    const tab = await canvas.findAllByRole('tab');
-    await userEvent.click(tab[2]);
-};
+// export const EditStatusTab = Template.bind({});
```
msnyder-msft, line 176: do any of these scenarios make sense to wire up against the visual rules logic so we still have them covered?

src/Components/ElementsPanel/Internal/ElementsList.tsx
```diff
@@ -72,155 +145,131 @@ const ElementsList: React.FC<IViewerElementsPanelListProps> = ({
 
 function getListItems(
```
msnyder-msft, line 146: this looks like more than just renames. Is there functionality here?

src/Components/ElementsPanel/Internal/ElementsList.tsx
```diff
@@ -72,155 +145,131 @@ const ElementsList: React.FC<IViewerElementsPanelListProps> = ({
...
     panelItems: Array<IViewerElementsPanelItem>,
+    isModal: boolean,
+    openCalloutRowId: string,
     onItemClick: ElementsPanelCallback,
-    onItemHover?: ElementsPanelCallback,
-    onItemBlur?: ElementsPanelCallback
+    onItemHover: ElementsPanelCallback,
+    onItemBlur: ElementsPanelCallback,
+    onItemHoverExtended: ViewerElementsPanelCallback,
```
msnyder-msft, line 153: do we have to pass separate callbacks? Can't we just always pass the extended version and just handle the cases internally to those functions where there isn't a callout?

src/Components/VisualsModal/VisualsModal.tsx
```diff
@@ -0,0 +1,60 @@
...
+    onItemClick,
+    onItemBlur,
+    onItemHover
+}) => {
+    const styles = getStyles();
+
+    return (
+        <div
+            style={{ top: position.top, left: position.left }}
+            className={styles.boundaryLayer + ' cb-base-fade-in'}
```
msnyder-msft, line 45: nit: there's something about doing class names this way that sometimes causes issues. Can you use the `css` helper that `fluent-ui` exposes to combine these class names?

src/Models/Context/BehaviorFormContext/BehaviorFormContext.test.tsx
```diff
@@ -28,108 +27,95 @@ import {
 describe('BehaviorFormContext', () => {
     afterEach(cleanup);
     describe('Actions', () => {
-        describe('Alert visuals', () => {
-            const getAlertVisual = (
-                expression: string
+        // TODO: REMOVE
```
msnyder-msft, line 30: why is there a TODO to remove them?

msnyder-msft (approved):

src/Models/Context/BehaviorFormContext/BehaviorFormContext.test.tsx
pasanchMSFT: Nice catch, removed the comment

src/Components/ElementsPanel/Internal/ElementsList.tsx
pasanchMSFT: I need to send in both callbacks in because badge rows don't open the tooltip. Adding an argument to open or not the tooltip feels worse than two functions.

src/Models/Classes/3DVConfig.ts
```diff
@@ -98,38 +93,8 @@ export const defaultOnClickPopover: IPopoverVisual = {
...
-    ],
-    objectIDs: {
-        expression: 'objectIDs'
-    }
-};
-
-export const getDefaultVisualRule = (): IExpressionRangeVisual => ({
-    id: createGUID(),
+export const getDefaultVisualRule = (id?: string): IExpressionRangeVisual => ({
+    id: id ? id : createGUID(),
```
iremgokce, line 97: is there a reason why we pass id here as optional to get a default visual rule object? I see that you use it for passing mock_id but we could get a default visual rule as a variable and pass its guid generated id in the below state action instead. The reason is that the getDefault... methods should be similar by setting id to either empty string (to generate the guid later at the time of save for example) or a createGuid as we have been doing in that file (for consistency).

image

iremgokce (approved):
Great refactoring work and as far as played with it, it works fine (except checking twin aliases that are created on the fly - without updating the behavior - to be listed in the twin property dropdown while adding/editing a visual rule. We see those twin aliases and their properties while adding/editing widgets though)

pasanchMSFT:
> Great refactoring work and as far as played with it, it works fine (except checking twin aliases that are created on the fly - without updating the behavior - to be listed in the twin property dropdown while adding/editing a visual rule. We see those twin aliases and their properties while adding/editing widgets though)

Found what the issue was with that and fixed it. I was grabbing the behavior for the dropdown from Scene context and not Behavior form context.

PR #784, opened 2022-11-18
[React-select for ResourcePicker]: Component refactors

iremgokce:
### Summary of changes 🔍 
- Replaced fluent combobox with [react-select](https://react-select.com/components) component in `ResourcePicker ` component to get nice features like search&filtered list of options and made necessary changes in parent component and data interaction flow
- Since react-select's perf is really bad with long lists, as suggested [here ](https://github.com/JedWatson/react-select/issues/3128) I added [react-window](https://www.npmjs.com/package/react-window) package to render only visible part of long list with thousands of options in dropdown (called virtualized list - I used `VirtualizedComboBox` from fluent before to achieve this) as I pull all the resources in single call in `AzureManagementAdapter` (**not** making async calls as user scrolls down by loading more options, if this is requested I need to change AzureManagementAdapter `getResourcesByPermission` method)

image
image
image
image
image
image
image

### Testing 🧪
Test it with either standalone ResourcePicker stories or by opening EnvironmentPicker in ADT3DScenePage. Try:
- search&auto complete, menu list interactions
- clear input
- create additional options with "No match found, create ..." option
- check the dependency between storage account and container pickers (whenever you change, clear a storage account - container picker should fetc/update its options accordingly)

### Checklist ✔️
- [x] Added relevant labels


pasanchMSFT:
Should we update current react-select components to include react-window? Does react-window affect the search functionality of react-select or it just limits number of rendered items?

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -77,7 +77,12 @@ const dialogStyles: Partial<IModalStyles> = {
     main: {
         width: '640px !important',
         maxWidth: 'unset !important',
-        minHeight: 'fit-content'
+        minHeight: 'fit-content',
+        overflow: 'visible'
+    },
+    scrollableContent: {
+        overflow: 'visible',
+        '> div:first-child': { overflow: 'visible' }
```
pasanchMSFT, line 85: Does this code make the overflow of the dropdown options actually show up? I believe we have been using Fluent's Callout in other places, if this works in the other places it could improve performance.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -547,47 +514,122 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
...
+            >
+                {({ index, style }) => (
+                    <div style={style}>{children[index]}</div>
+                )}
+            </FixedSizeList>
+        );
+    };
+
+    const CustomOption = (props) => {
+        if (props.data.type === 'header') {
```
pasanchMSFT, line 543: Why do you prefer to create header options instead of using react-select groups?

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -547,47 +514,122 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
...
+                    value={selectedOption}
+                    defaultValue={selectedOption ?? undefined}
+                    defaultInputValue={searchValue ?? ''}
+                    placeholder={placeholder()}
+                    formatCreateLabel={(inputValue: string) =>
+                        `${t(
+                            'resourcesPicker.useNonExistingResource'
+                        )} "${inputValue}"`
+                    }
+                    noOptionsMessage={() => t('resourcesPicker.noOption')}
```
pasanchMSFT, line 612: you can store the value in a variable instead of doing an inline function here.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -547,47 +514,122 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
...
+                    isLoading={resourcesState.isLoading}
+                    isDisabled={disabled}
+                    components={{
+                        Option: CustomOption,
+                        MenuList: CustomMenuList
+                    }}
+                    onChange={handleOnChange}
+                    onInputChange={handleOnInputChange}
+                    filterOption={createFilter({ ignoreAccents: false })}
+                    loadingMessage={() => loadingLabelText}
```
pasanchMSFT, line 624: Same as above inline function

pasanchMSFT (commented):
Solid changes, left some comments, let me know if you want to chat about them Irem.

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: "Does this code make the overflow of the dropdown options actually show up" - yep that is why I did that. I am using Dialog fluent component in EnvironmentPicker and there is no prop to control overflow hidden that is why I did it with styling. 

You mentioned "callout" component, it seems there is a prop for Callout fluent component that you can use: `hideOverflow`. In fact, I noticed that TwinPropertySearchDropdown defines a custom Menu for react-select component where the menu is a Callout fluent component, I think in that case "hideOverflow" could be set to false to show the overflow maybe since Callout is rendered outside of the parent component in DOM? But I didn't use a custom menu in my case, so I achieve this with styling.

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: yeah, I tried it there was a whole trial-error about using select groups. The reason I didn't use it was because I was not able to render each item correctly in a virtualized list when I use react-window as it expects a one-dimensional flat list of options. Since each children in DOM has might have sub-children (in react-select group there are sub options field), the rendered option texts were overlapping.

iremgokce:
> Should we update current react-select components to include react-window? Does react-window affect the search functionality of react-select or it just limits number of rendered items?

react-window only enables list-virtualization which means only rendering what is visible to the user. It is not affecting the search functionality, you only use a wrapper component from that react-window library to wrap your list and tell the length of the options and height of each option to that wrapper component and it handles the rendering of that list with performance. If you also experienced lagging in your react-select components we might need to use this as well. [It seems](https://github.com/JedWatson/react-select/issues/2970) react-select starts slowing down with around 1000 options, so we might need to use this even if we use infinite scroll (loading items async as we scroll down) as in Twin search dropdown as there might be thousands if twins if user do not use search functionality.

So, maybe we can create a react-select standalone component which is wrapped by react-window and wherever we use react-select we can consume that? But since react-select props are very customizable it may not worth it as each component is very custom in our cases where we use react-select (CreatableSelect).

pasanchMSFT (approved):
Thanks for the detailed responses. LGTM

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -547,47 +514,122 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
...
+            delete props.innerProps.onMouseMove;
+            delete props.innerProps.onMouseOver;
+        }
+        return (
+            <div className={classNames.optionWrapper}>
+                <components.Option
+                    {...props}
+                    isDisabled={props.data.type === 'header'}
+                >
+                    {props.data.__isNew__ ? (
```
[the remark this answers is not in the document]
iremgokce: react-select marks the option data with `__isNew__` key if the option is not found after searching when it offers to create a new option with "Not found, create new..." message with the entered value. So, as far as I can tell it is safe to use.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -547,47 +514,122 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
...
+                    isSearchable
+                    isLoading={resourcesState.isLoading}
+                    isDisabled={disabled}
+                    components={{
+                        Option: CustomOption,
+                        MenuList: CustomMenuList
+                    }}
+                    onChange={handleOnChange}
+                    onInputChange={handleOnInputChange}
+                    filterOption={createFilter({ ignoreAccents: false })}
```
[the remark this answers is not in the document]
iremgokce: it was one of the [suggested methods](https://github.com/JedWatson/react-select/issues/3128) to improve rendering performance when there are many options in the list. I added this before implementing react-window to render a virtualized list with overriding MenuList of react-select. It basically reduces the lag when you're typing into the select, but now with using react-window, I can remove it.

src/Models/Constants/Types.ts
```diff
@@ -249,6 +250,7 @@ export type AzureResourceSearchParams = {
         storageAccountBlobUrl?: string;
         [key: string]: any;
     };
+    isAdditionalSearchParamsRequired?: boolean;
```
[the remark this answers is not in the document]
iremgokce: it is the `additionalParams` one above in that typing definition. Instead I can make two types like WithAdditionalParams | WithoutAdditionalParams to ensure dependency in parameters?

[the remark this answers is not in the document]
msnyder-msft: that's always my personal preference since it makes it more obvious to someone consuming which args are dependent on eachother. It's kind of self-documenting and stupid proofs someone from using it in a way you didn't intend by mistake.

darsney (approved):
This feels _very good_ and is a significant improvement to the original implementation.  Honestly the onlyyy thing thats a bit weird is that when you click the down arrow there isnt anything in the dropdown besides the resource you selected, so you have to clear it to see everything, but i think thats also fine, just takes 1 shot to learn how to do it.  

I'm fine to have a conversation about what all changed here just as an FYI since its a big change and brought in a new package, but experience wise this LGTM for sure

src/Models/Constants/Types.ts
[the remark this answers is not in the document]
iremgokce: added discriminating union types here too

iremgokce:
> This feels _very good_ and is a significant improvement to the original implementation. Honestly the onlyyy thing thats a bit weird is that when you click the down arrow there isnt anything in the dropdown besides the resource you selected, so you have to clear it to see everything, but i think thats also fine, just takes 1 shot to learn how to do it.
> 
> I'm fine to have a conversation about what all changed here just as an FYI since its a big change and brought in a new package, but experience wise this LGTM for sure

@darsney it happens because back then you asked to have the vertical line character at the end of the search term [by default](https://react-select.com/home) which is not the case in default behavior of react-select - it is set to beginning since default input value is empty by default:
image
 So that is why I set the `defaultInputValue` to the search term which brings the focus to the end of the search term as we click on the input field/open menu. That causes to filter the list to the search term which ends up only seeing that option when you open the menu using down arrow.

So, which one feels more important for you, default search term focus position or menu listing all?

iremgokce:
> > This feels _very good_ and is a significant improvement to the original implementation. Honestly the onlyyy thing thats a bit weird is that when you click the down arrow there isnt anything in the dropdown besides the resource you selected, so you have to clear it to see everything, but i think thats also fine, just takes 1 shot to learn how to do it.
> > I'm fine to have a conversation about what all changed here just as an FYI since its a big change and brought in a new package, but experience wise this LGTM for sure
> 
> @darsney it happens because back then you asked to have the vertical line character at the end of the search term [by default](https://react-select.com/home) which is not the case in default behavior of react-select - it is set to beginning since default input value is empty: image So that is why I set the `defaultInputValue` prop to the search term which brings the focus to the end of the search term in the input field as we click on the input field/open menu. That causes to filter the list to the search term which ends up only seeing that option when you open the menu using down arrow.
> 
> So, which one feels more important for you, default search term focus position or menu listing all?

Hi @darsney I made a change to show all the list with subscription headers when there is no default input (reverted setting default input prop to search term which is now the default react-select behavior) and added initial offset to the selected item. Please pull and test it again and let me know if this feels better.

src/Components/ResourcePicker/ResourcePicker.tsx
```diff
@@ -547,47 +515,149 @@ const ResourcePicker: React.FC<IResourcePickerProps> = ({
...
-                onChange={(_e, option, _idx, value) => {
-                    handleOnChange(option, value);
-                }}
-                onRenderLabel={handleOnRenderLabel}
-                onRenderOption={handleOnRender}
-            />
+            <Stack tokens={{ childrenGap: 4 }}>
+                {label && <Stack horizontal>{dropdownLabel}</Stack>}
+                <CreatableSelect
+                    aria-labelledby="resource-picker-dropdown-label"
```
msnyder-msft, line 634: doesn't there need to be an element with this id? I think that hsould be the label on the line above right?

msnyder-msft (approved):

src/Components/ResourcePicker/ResourcePicker.tsx
iremgokce: yes, it is the id of that label

PR #789, opened 2022-12-01
Create twin search component and use where needed

pasanchMSFT:
### Summary of changes 🔍 
Created `TwinSearch` component that includes `TwinPropertySearchDropdown`, `PropertyInspector`, and `AdvancedSearch`. Updated Element form and Twin alias forms to use it.

### Testing 🧪
- `TwinSearch` component story
- Update a behavior by adding a new twin alias, the form should now have `TwinSearch` component.
  - Editing the twin alias should also have `TwinSearch`.
- Element form should now have `TwinSearch` component and functionality should be the same.

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/ADT3DSceneBuilder/Internal/Elements/Internal/ElementTwinAliasForm.tsx
+++ b/src/Components/ADT3DSceneBuilder/Internal/Elements/Internal/ElementTwinAliasForm.tsx
@@ -129,18 +129,6 @@ const ElementTwinAliasForm: React.FC = () => {
                         }
                     }}
                 />
-                {/* <TwinPropertySearchDropdown
-                    key={'aliased-twin'}
-                    adapter={adapter}
-                    label={t('twinId')}
-                    labelIconName="Shapes"
-                    onChange={(selectedTwinId: string) => {
-                        handleTwinSelect(selectedTwinId);
-                    }}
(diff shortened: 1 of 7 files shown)


montgomp (approved):

src/Components/ADT3DSceneBuilder/Internal/Elements/Internal/ElementTwinAliasForm.tsx
```diff
@@ -132,7 +129,7 @@ const ElementTwinAliasForm: React.FC = () => {
                         }
                     }}
                 />
-                <TwinPropertySearchDropdown
+                {/* <TwinPropertySearchDropdown
```
msnyder-msft, line 132: do you want to remove teh commented section now?

msnyder-msft (approved):
This is a really good refactor! Nice job. I don't see any issues with the experience. Code looks good. 
Only thing would be if we wanted a few other stories for the new component if there are different variants.

src/Components/ADT3DSceneBuilder/Internal/Elements/Internal/ElementTwinAliasForm.tsx
pasanchMSFT: so close to a no comments PR :(

PR #793, opened 2022-12-05
Fix bugs in Visual rules

pasanchMSFT:
### Summary of changes 🔍 
Fixes following bugs:
- Condition range turns to null after user inputs Infinity
- Condition amount does not change Save button disabled

Also removes some unused lines.

### Testing 🧪
- Create a new visual rule
  - When adding a condition input -Infinity or Infinity in the number range input fields. (This should lead to the range having infinity set as one the values)
  - Add a condition when the rule has none. (This should remove the disabled state of the save button)
  - Remove a condition if the rule only has one. (This should add the disabled state of the save button)

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
+++ b/src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
@@ -119,6 +119,11 @@ const ConditionsCallout: React.FC<IConditionsCalloutProps> = (props) => {
                     payload: { values: newValues }
                 });
             } else {
+                // If user types infinity or -infinity handle that case (JSON parsing turns numerical infinities to null)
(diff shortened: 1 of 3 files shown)


msnyder-msft (approved):

src/Components/ADT3DSceneBuilder/Internal/VisualRuleForm/Internal/ConditionsCallout/ConditionsCallout.tsx
```diff
@@ -119,6 +119,11 @@ const ConditionsCallout: React.FC<IConditionsCalloutProps> = (props) => {
                     payload: { values: newValues }
                 });
             } else {
+                // If user types infinity or -infinity handle that case (JSON parsing turns numerical infinities to null)
+                if (newValues[0] === Infinity || newValues[0] === -Infinity) {
```
iremgokce, line 123: isn't newValues [min, max] tuple? shouldn't we check index 1 as well and do casting for Infinity if it is?

iremgokce (approved):

PR #796, opened 2022-12-07
[Environment Picker]: Null check for container url and removed tenant ID from ARG query

iremgokce:
### Summary of changes 🔍 
- Added null checks for optional storage prop containerUrl (on flights app on the pages where the <EnvironmentPicker/> component is used without props like /models /twins pages, there is error boundary popping up, this fixes it)
- Removed tenantId filtering from ARG calls not to restrict the get resources query to the logged in user's tenant id, otherwise they get empty array from the call even though they have the resource under their subscription in other tenants.
image
image

### Testing 🧪
Ideally it should be tested with logging in different tenant (by building cardboard and npm install on flights app).

### Checklist ✔️
- [x] Added relevant labels
- [] Request UI review from design / PM


msnyder-msft (approved):

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -392,11 +397,11 @@ const EnvironmentPicker = ({
...
     const containerDisplayText = useMemo(() => {
         const displayText = getContainerDisplayText(
-            getContainerNameFromUrl(storage.containerUrl),
-            getStorageAccountUrlFromContainerUrl(storage.containerUrl)
+            getContainerNameFromUrl(storage?.containerUrl),
+            getStorageAccountUrlFromContainerUrl(storage?.containerUrl)
         );
         return displayText || t('environmentPicker.noContainer');
-    }, [t, storage.containerUrl]);
+    }, [t, storage?.containerUrl]);
```
pasanchMSFT, line 404: how safe is this? I don't think I've used it before.

iremgokce: yep it is possible, React keeps track of changes in the dependency array, storage?.containerUrl can be either `undefined` or a value. [undefined] is same as []. So in either way it treats it as a change and call useEffect again: https://stackoverflow.com/questions/61439712/react-conditional-dependency-on-useeffect-or-how-to-prevent-redundancy

pasanchMSFT (approved):

src/Adapters/AzureManagementAdapter.ts
```diff
@@ -99,9 +99,7 @@ export default class AzureManagementAdapter implements IAzureManagementAdapter {
             },
             data: isResourceGraphCall(params)
                 ? {
-                      query: `Resources | where type =~ '${
```
msnyder-msft, line 102: does it make sense to roll this behind a local storage flag? They can test it out and it it works, we can remove the flag and send it for real? That would help to avoid any potential regression for other folks

darsney: my preference is to check in as is, and send to dev.  i think the chance of this regressing others is low from what i can tell reading the docs (the filter is basically optional), and it means we can roll to prod that much faster without another code change.  plus it makes validating it for customers a lot simpler, not everyone is trying to crack the console open like we are haha

PR #799, opened 2022-12-07
Gauge widget should be configurable to show decimals

abalogun-dev:
### Testing 🧪
 > Take a look at the Gauge Widget Stories and also, create gauge widgets with decimal or large numbers and verify that the appropriate number is displayed.

### Checklist ✔️
- [ ] Added relevant labels

https://msazure.visualstudio.com/One/_workitems/edit/14610114/

### DIFF
--- a/src/Components/BehaviorsModal/Internal/Widgets/GaugeWidget/GaugeWidget.tsx
+++ b/src/Components/BehaviorsModal/Internal/Widgets/GaugeWidget/GaugeWidget.tsx
@@ -6,6 +6,7 @@ import { getStyles } from './GaugeWidget.styles';
 import GaugeChart from 'react-gauge-chart';
 import { BehaviorsModalContext } from '../../../BehaviorsModal';
 import { BehaviorModalMode } from '../../../../../Models/Constants';
+import { formatNumber } from '../../../../../Models/Services/Utils';
 
 interface IProp {
     widget: IGaugeWidget;
@@ -17,6 +18,7 @@ const GaugeWidget: React.FC<IProp> = ({ widget }) => {
     const label = widget.widgetConfiguration.label;
     const units = widget.widgetConfiguration.units || '';
     let value = 0;
+    let format = '';
     try {
         if (mode === BehaviorModalMode.preview) {
             // In preview mode, gauge uses min value range as value
@@ -32,12 +34,11 @@ const GaugeWidget: React.FC<IProp> = ({ widget }) => {
             if (!value) {
                 value = 0;
             }
-            value = Math.floor(value);
         }
     } catch {
         value = 0;
     }
-
+    format = formatNumber(value);
     const { valueRanges } = widget.widgetConfiguration;
 
     // Get active color from value range -- if value not in defined range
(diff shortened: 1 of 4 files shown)


src/Components/BehaviorsModal/Internal/Widgets/GaugeWidget/GaugeWidget.tsx
```diff
@@ -17,6 +18,7 @@ const GaugeWidget: React.FC<IProp> = ({ widget }) => {
     const label = widget.widgetConfiguration.label;
     const units = widget.widgetConfiguration.units || '';
     let value = 0;
+    let format = '';
```
iremgokce, line 21: nit: renaming this var to reflect that it is a formatted "value", not format itself.
```suggestion
    let formatedValue = '';
```

src/Models/Services/Utils.ts
```diff
@@ -831,3 +832,30 @@ export const getMockTimeSeriesDataArrayInLocalTime = (
         })).sort((a, b) => (a.timestamp as number) - (b.timestamp as number));
     });
 };
+
+export function formatNumber(val: number) {
```
iremgokce, line 836: can we add a block comment before this method to explain how it formats the passed number with e.g. input and output?

src/Models/Services/Utils.ts
```diff
@@ -36,6 +36,7 @@ import {
     IConsoleLogFunction,
     TimeSeriesData
 } from '../Constants/Types';
+import * as d3 from 'd3';
```
iremgokce, line 39: hmm, instead of importing whole d3, can we try to only import the format method from that package? of I noticed that you can import it directly from submodule like:
```suggestion
import { format } from 'd3-format';
```

can you please try this and confirm it still works?

src/Models/Services/Utils.ts
```diff
@@ -831,3 +832,30 @@ export const getMockTimeSeriesDataArrayInLocalTime = (
         })).sort((a, b) => (a.timestamp as number) - (b.timestamp as number));
     });
 };
+
+export function formatNumber(val: number) {
+    if (Math.abs(val) < 1000000) {
+        if (Math.abs(val) < 0.000001) {
+            // display as scientific notation if greater than 100 thousandths
+            return d3.format('.1n')(val);
+        } else {
```
iremgokce, line 841: if the value is 0.000001 this is how it looks like: is this what we want? should it be like using e-6 notation instead of ellipsis?
image

pasanchMSFT (approved):
I've got no comments other than what Irem pointed out. I'll approve but please resolve those.

montgomp (approved):

src/Components/BehaviorsModal/Internal/Widgets/GaugeWidget/GaugeWidget.tsx
```diff
@@ -32,12 +34,11 @@ const GaugeWidget: React.FC<IProp> = ({ widget }) => {
             if (!value) {
                 value = 0;
             }
-            value = Math.floor(value);
         }
     } catch {
         value = 0;
     }
-
+    formatedValue = formatNumber(value);
```
msnyder-msft, line 41: super nit. can we just instantiate the variable here since it's not set before this?
`const formattedValue`
Another super nit, formatted has 2 't's. 🙂

msnyder-msft (approved):

PR #803, opened 2022-12-12
colorPicker theming

### LINKED ISSUE #312: Feat: Color picker styling should align with icon picker
**Is your feature request related to a problem? Please describe.**
The color picker uses the default fluent picker which doesn't quite match with the styling of the icon picker when items are selected or being tabbed.

Designs
https://www.figma.com/file/LsmZeVOZJgwZs75HPTQmgU/3D-Scenes-Mock-Screens?node-id=44214%3A160541
image

Actual 
image

**Describe the solution you'd like**
The focus targets should be squared off and not shrink the colored circles when selected

montgomp:
image
image
image

### Testing 🧪

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Components/Pickers/Internal/Picker.base.styles.ts
+++ b/src/Components/Pickers/Internal/Picker.base.styles.ts
@@ -8,22 +8,34 @@ const classNames = {
 export const getStyles = (_props: IPickerBaseStyleProps): IPickerBaseStyles => {
     return {
         root: [classNames.root],
-        button: [
-            classNames.button,
-            {
-                border: '1px solid var(--cb-color-input-border)',
-                borderRadius: '50%',
-                cursor: 'pointer',
-                height: 28,
-                width: 28
-            }
-        ],
+        button: [],
         subComponentStyles: {
             callout: {
                 calloutMain: {
                     // this is not exposed on the styles directly
                     '& .ms-swatchColorPickerBodyContainer': {
-                        minWidth: 'unset'
+                        minWidth: 'unset',
+                        ' .ms-Button': {
+                            cursor: 'pointer',
+                            backgroundColor: 'transparent',
+                            width: 28,
+                            height: 28,
+                            margin: 0,
+                            padding: '2px',
+                            borderRadius: 0,
+                            borderStyle: 'solid',
+                            borderWidth: 0,
+                            ':hover': {
+                                backgroundColor:
+                                    'var(--cb-color-card-contrast)',
+                                '::before': {
+                                    boxShadow: 'none'
+                                },
+                                borderRadius: 0,
+                                margin: 0,
+                                padding: '2px'
+                            }
+                        }
                     },
                     // there is no good class to grab here
                     '& table': {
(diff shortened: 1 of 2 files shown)


src/Components/Pickers/Internal/Picker.base.styles.ts
```diff
@@ -8,22 +8,34 @@ const classNames = {
...
-            }
-        ],
+        button: [],
         subComponentStyles: {
             callout: {
                 calloutMain: {
                     // this is not exposed on the styles directly
                     '& .ms-swatchColorPickerBodyContainer': {
-                        minWidth: 'unset'
+                        minWidth: 'unset',
```
msnyder-msft, line 17: did you yank this from teh fluent repo or how did you come up with all these styles?

src/Components/Pickers/Internal/Picker.base.styles.ts
```diff
@@ -8,22 +8,34 @@ const classNames = {
...
-            classNames.button,
-            {
-                border: '1px solid var(--cb-color-input-border)',
-                borderRadius: '50%',
-                cursor: 'pointer',
-                height: 28,
-                width: 28
-            }
-        ],
+        button: [],
```
msnyder-msft, line 11: you can remove this style entirely now I think? Or leave it but keep the className.button part in there.

montgomp: nope, the IPickerBaseStyles return needs to have the `button` array or there's complier errors

src/Components/Pickers/Internal/Picker.base.styles.ts
montgomp: I looked at Fluent repo but the majority came from opening dev tools and looking at the DOM generated from the ColorSwatchPicker

src/Components/Pickers/Internal/Picker.base.styles.ts
```diff
@@ -8,22 +8,34 @@ const classNames = {
...
+                            width: 28,
+                            height: 28,
+                            margin: 0,
+                            padding: '2px',
+                            borderRadius: 0,
+                            borderStyle: 'solid',
+                            borderWidth: 0,
+                            ':hover': {
+                                backgroundColor:
+                                    'var(--cb-color-card-contrast)',
```
iremgokce, line 30: do you think we can do better with on hover background in darker theme colors? In the screenshots it looks like more white bg color on hover, but generally it is more dark grey on hover of icon buttons in fluent design, can we make it here too? https://developer.microsoft.com/en-us/fluentui#/controls/web/button 
image

montgomp: We definitely can. The color picker is mirroring the values used for icon picker so if we're changing the coloration, we should change in both. I'll chase this down with design.

montgomp: Moved to semantic color in theme as the light theme variable was the only value applying.

src/Components/Pickers/Internal/Picker.base.styles.ts
```diff
@@ -6,24 +6,36 @@ const classNames = {
     button: `${classPrefix}-button`
 };
 export const getStyles = (_props: IPickerBaseStyleProps): IPickerBaseStyles => {
+    const { theme } = _props;
```
pasanchMSFT, line 9: nit: _props can be renamed to props if it is being used. The underscore is used to avoid a lint warning for unused arguments.

pasanchMSFT (approved):
Nit comment, otherwise, looks good

msnyder-msft (approved):

PR #806, opened 2022-12-15
[Resource Picker]: Reverted back to react-select allow input edit on focus and bug fixes

iremgokce:
### Summary of changes 🔍 

- Reverted back to the original experience for initial focus and input value in the react-select component for easy edit.

So, currently it is like (listing all resources initially since the input value is set to null and therefore the focus is at the beginning when on lick of input field):
image

Reverted back to this (setting the input field value always to be able to let user edit the value easily, but this way we will not be able to see other options in the list since react-select filters based on the input value internally):
image

- Fixed a main bug that I noticed on **demo** page in flights. Steps to reproduce:
  - open /3dscenes and configure your environment using environment picker
  - go to /demo page
  - come back to /3dscenes and see that your configuration is messed in the picker (shows demo related values)

- Fixed another bug for default selected value:
  - open environment picker on demo page
  - clearing a selected option for storage account (which is an additional/not coming from ARG results)
  - click on close to close the environment picker
  - re-open the environment picker and see that the storage account is not selected by default

- Fixed another bug in `getUrlFromString` method
- Fixed another bug for case insensitive adt instance hostname filter in ARG call:
image

### Testing 🧪

### Checklist ✔️
- [x] Added relevant labels


montgomp (approved):
LGTM

pasanchMSFT (approved):

msnyder-msft:
@iremgokce  Is this good to complete?

PR #807, opened 2022-12-17
[Data History Explorer]: Builder components

iremgokce:
### Summary of changes 🔍 
You can find the design doc [here](https://microsoft.sharepoint.com/:fl:/s/7e83e18c-fe3f-4f63-8c39-8b65b78daa06/EZDob3PnLNBNhEG3Kyath6IBiGzyEB2J7b9dVSmmEXrsMA?e=8oZ4UN&nav=cz0lMkZzaXRlcyUyRjdlODNlMThjLWZlM2YtNGY2My04YzM5LThiNjViNzhkYWEwNiZkPWIheHBmYWJCRUVCVWFMcF90MTJRbDV0ck04dzZyZHV4NU5xdi1PNWdJLU1MaEpNUU9qVkItUVE3dFhGcm9KbGlDSCZmPTAxUEFYQ0Q0VVE1QlhYSFpaTTJCR1lJUU5YRk1USzNCNUMmYz0lMkYmZmx1aWQ9MSZhPUxvb3BBcHAmcD0lNDBmbHVpZHglMkZsb29wLXBhZ2UtY29udGFpbmVyJng9JTdCJTIydyUyMiUzQSUyMlQwUlRVSHh0YVdOeWIzTnZablF1YzJoaGNtVndiMmx1ZEM1amIyMThZaUY0Y0daaFlrSkZSVUpWWVV4d1gzUXhNbEZzTlhSeVRUaDNObkprZFhnMVRuRjJMVTgxWjBrdFRVeG9TazFSVDJwV1FpMVJVVGQwV0VaeWIwcHNhVU5JZkRBeFVFRllRMFEwVWpSTVRVTmFWRVJPU0ZsU1IwdFdOMU5KTnpSS1JGaFpWakklM0QlMjIlMkMlMjJpJTIyJTNBJTIyNzczYTcwZDctMTYwOS00YmJlLTgyMjctN2Q5MjA3ZDdhOGY4JTIyJTdE).
- Added `DataHistoryExplorer` and `DataHistoryExplorerModal `standalone components
- DataHistoryExplorer has `TimeSeriesBuilder `and `TimeSeriesCallout `internal components
- Added `ADTDataHistoryAdaper `which is a mixin of ADT and ADX adapters
- Added `DataHistoryUtils `shared file and moved some files to there since now these components are also using some methods that data history widget related components used to use only before.
- Moved some constants and types around and did some renamings **that is why it touches many files**

**TODO**: I will validate the design with @azimamansuri. The items to be discussed:
- [ ] Is coloring the data types okay instead of using a circle and adding a data type icon next to it as we originally discussed with @darsney 
- [ ] Is keeping the "Add new" stacked at the bottom of the builder on the left hand side okay (even if the list is empty)? Otherwise it keeps moving as we add another item using the callout. Alternatively, I can change it to primary button and use the style as we use for Elements and Behaviors stacked to bottom with border top.
- [ ] Should we enable "Property Inspector" and "Advance Twin Search" buttons on the TimeSeriesCallout? Since it is already a callout, another modal or callout on top of it may not be good for the user as it was part of our discussion with @darsney?

image
image

### Testing 🧪
I added both local and mock stories for each of those components, please search for their names in storybook

### Checklist ✔️
- [x] Added relevant labels


src/Components/CardboardList/CardboardListItem.tsx
```diff
@@ -125,6 +125,9 @@ export const CardboardListItem = <T extends unknown>(
                         (typeof iconStart.name === 'string' ? (
                             <FontIcon
                                 iconName={iconStart.name}
+                                style={{
+                                    color: (iconStart as any).color
```
montgomp, line 129: Any chance that casting iconStart to any wouldn't have a `.color`? Should we provide a default here?

iremgokce: ahh okay it was due to build error since in the type definition iconStart can either be a object (even if I added color?: string there) or react element. So my fix is not working still, I will fix it with type checking.

image
image

src/Components/DataHistoryExplorerModal/DataHistoryExplorerModal.styles.ts
```diff
@@ -0,0 +1,21 @@
...
+    contentStack: `${classPrefix}-content-stack`
+};
+export const getStyles = (
+    _props: IDataHistoryExplorerModalStyleProps
+): IDataHistoryExplorerModalStyles => {
+    return {
+        root: [classNames.root],
+        contentStack: [classNames.contentStack, { height: '100%' }],
+        subComponentStyles: {
+            modal: { titleContainer: { i: { paddingTop: 0 } } }
```
msnyder-msft, line 18: there is a property name called "i"?

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.types.ts
```diff
@@ -0,0 +1,41 @@
...
+    ITheme,
+    IToggleStyles
+} from '@fluentui/react';
+import { IDataHistoryTimeSeriesTwin } from '../../../../Models/Constants';
+
+export interface ITimeSeriesTwinCalloutProps {
+    timeSeriesTwin?: IDataHistoryTimeSeriesTwin;
+    target: string;
+    onDismiss?: () => void;
+    onPrimaryActionClick: (timeSeriesTwin: IDataHistoryTimeSeriesTwin) => void;
```
msnyder-msft, line 15: isn't this always an "add" operation? if so, can we rename to `onAdd`? If not, shouldn't we pass the operation type so the parent knows what to do with the information?

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.tsx
```diff
@@ -0,0 +1,235 @@
...
+    return (
+        <DataHistoryExplorerContext.Consumer>
+            {({ adapter }) => (
+                <div className={classNames.root}>
+                    <Callout
+                        target={`#${target}`}
+                        styles={classNames.subComponentStyles.callout}
+                        onDismiss={onDismiss}
+                        directionalHint={DirectionalHint.bottomCenter}
+                        isBeakVisible={false}
```
msnyder-msft, line 136: why no beak?

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.tsx
```diff
@@ -0,0 +1,235 @@
...
+                    >
+                        <Stack tokens={{ childrenGap: 12 }}>
+                            <Stack tokens={{ childrenGap: 4 }}>
```
msnyder-msft, line 139: don't we typically use 8 as our spacing?

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+            <DefaultButton
+                text="Target button"
+                id="mock-time-series-twin-callout-target"
+            />
+            <TimeSeriesTwinCallout {...args} />
+        </DataHistoryExplorerContext.Provider>
+    );
+};
+
+export const Mock = Template.bind({}) as TimeSeriesTwinCalloutStory;
```
msnyder-msft, line 36: Are there more scenarios we should validate or show for this callout?

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.stories.local.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { ComponentStory } from '@storybook/react';
+import { getDefaultStoryDecorator } from '../../../../Models/Services/StoryUtilities';
+import TimeSeriesTwinCallout from './TimeSeriesTwinCallout';
+import { ITimeSeriesTwinCalloutProps } from './TimeSeriesTwinCallout.types';
```
msnyder-msft, line 5: I noticed that we have stories.local for most of these components. Does every component really make calls to live endpoints?

src/Components/DataHistoryExplorer/DataHistoryExplorer.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { ComponentStory } from '@storybook/react';
+import { getDefaultStoryDecorator } from '../../Models/Services/StoryUtilities';
+import DataHistoryExplorer from './DataHistoryExplorer';
+import { IDataHistoryExplorerProps } from './DataHistoryExplorer.types';
+import MockAdapter from '../../Adapters/MockAdapter';
+
+const wrapperStyle = { width: '100%', height: '500px', padding: 8 };
+
+export default {
+    title: 'Components/DataHistoryExplorer',
```
msnyder-msft, line 11: Can we add a folder for DataHistoryExplorer and then put the modal and the component itself inside the folder?

msnyder-msft:
Interaction comment: 
It feels like the callout should go to the right instead of down.
image

msnyder-msft:
the double scrollbars are not good. I think I saw you made a style change to the twin search. is that a side effect?
image

msnyder-msft:
The label for the toggle feels very loud. Did we make that text bold ourselves or is that default? It draws my eyes a lot and I feel like it shouldn't. Not a deal breaker but I'd be curious if it might look better as normal weight.
image

msnyder-msft:
The spacing between the toggle and the next field looks larger than the standard. Can you check if maybe there is something extra going on there?
image

msnyder-msft:
How would you feel about pushing the callout to the right of the list item?
image

pasanchMSFT (approved):
Code looks good, lightweight context threw me off a little bit but after giving that a second look it makes sense why you added it. Haven't played with the experience but will take a look at that.

msnyder-msft:
Opening twin search from the callout of a list item is a rouggggh experience. It seems to close the original modal, the callout is still there and now we have a modal on top of that. I think this is what you referenced in your question to @darsney but it doesn't feel great.
image

msnyder-msft:
There hsould be a primary button on the modal IMO, even if it doesn't do anything right now.
image

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
```diff
@@ -0,0 +1,89 @@
...
+                        </span>
+                    </div>
+                )}
+                <Stack
+                    horizontal
+                    tokens={{ childrenGap: 8 }}
+                    className={classNames.contentStack}
+                >
+                    <TimeSeriesBuilder
+                        onTimeSeriesTwinListChange={(twins) =>
```
msnyder-msft, line 70: can we avoid the full set operation here and instead have add/remove? that feels like the child has more granular control and doesn't just fully control the entire list.

msnyder-msft (commented):
Generally looks good. left some nits. I agree with your comments in the description that the stacking of modals and callouts doesn't feel good. Let's see what @darsney has to say on that. Biggest experience thing for me is the position of the callouts. I think they should push right and not down.

src/Components/CardboardModal/CardboardModal.types.ts
```diff
@@ -24,7 +24,7 @@ export interface ICardboardModalProps {
         linkProps?: ILinkProps;
     };
     /** click handle for the primary footer button */
-    footerPrimaryButtonProps: IButtonProps;
+    footerPrimaryButtonProps?: IButtonProps;
```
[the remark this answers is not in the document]
pasanchMSFT: Should it be possible to have a Modal without buttons? Maybe that is what we should have, in case of having buttons always have primary and secondary, and in case we don't have a primary don't show either.

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.tsx
```diff
@@ -0,0 +1,235 @@
...
+                                    onChange={handleLabelChange}
+                                />
+                            </Stack>
+                            <Stack
+                                horizontal
+                                horizontalAlign={'end'}
+                                tokens={{ childrenGap: 8 }}
+                            >
+                                <PrimaryButton
+                                    text={t('add')}
```
darsney, line 213: Text could be "Save" and potentially also have "Cancel" in the edit flow here?

image

src/Components/DataHistoryExplorer/Internal/TimeSeriesBuilder/TimeSeriesBuilder.tsx
```diff
@@ -0,0 +1,227 @@
...
+            <Stack tokens={{ childrenGap: 8 }}>
+                <h2 className={classNames.header}>
+                    {t('dataHistoryExplorer.builder.title')}
+                </h2>
+                <span className={classNames.description}>
+                    {t('dataHistoryExplorer.builder.description')}
+                </span>
+            </Stack>
+            <Separator />
+            <CardboardList
```
darsney, line 130: In the event that we cast, I think we should make the type numeric in the List

image

darsney (approved):
The experience feels remarkably complete for the first PR, nice stuff!  I left a couple comments that should be addressed, and I believe callouts should anchor on the right side of their target (not bottom), and we should figure out what the double scroll bar on the TSID autocomplete issue is.  

Advanced twin search is clunky right now because it is modal in modal, that is known and I don't think we should address it in this PR

For "Add New" my preference would be that if the list is empty, we put the button where the list is (perhaps with placeholder text/image and primary button), and if the list is nonempty we place it at the bottom using the current styling.  Please validate with @azimamansuri 

Coloring data types looks cool as well, but I think if its cast to numeric we should show "#" that way people know that things that are "#" can be viewed in the line chart, and other things cant

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
```diff
@@ -0,0 +1,89 @@
...
+>();
+
+const DataHistoryExplorer: React.FC<IDataHistoryExplorerProps> = (props) => {
+    const { adapter, hasTitle = true, styles } = props;
+
+    // hooks
+    const { t } = useTranslation();
+
+    // state
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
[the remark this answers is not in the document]
iremgokce: this was kept for the forked branch for viewer but will remove it here.

src/Components/DataHistoryExplorer/DataHistoryExplorer.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+    ]
+};
+
+type DataHistoryExplorerStory = ComponentStory<typeof DataHistoryExplorer>;
+
+const Template: DataHistoryExplorerStory = (args) => {
+    return <DataHistoryExplorer {...args} />;
+};
+
+export const Mock = Template.bind({}) as DataHistoryExplorerStory;
```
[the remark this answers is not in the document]
iremgokce: added stories for using `hasTitle` prop with and without to capture the snapshot.

src/Components/DataHistoryExplorerModal/DataHistoryExplorerModal.styles.ts
iremgokce: this is for aligning the text with the icon in the modal title:
image

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.tsx
iremgokce: as far as I know we have not been using beak in callouts in the builder components, no? Not sure though, I can confirm it with Azima.

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.stories.tsx
iremgokce: Added "New" and "Edit" stories to show `timeSeriesTwin` prop

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.stories.local.tsx
iremgokce: yes?

src/Components/CardboardModal/CardboardModal.types.ts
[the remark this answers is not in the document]
iremgokce: I don't think that we should always have the primary button but disable it if not needed. Since this is a "modal" component, it is possible that we do not need any button but a close button on the top right corner which is the common practice for fluent Modal component. In places where we use it as a "dialog" component, than having buttons (primary & secondary) make sense. In my case, I want to use it as a pure "modal" where only close button is needed with no primary button action. I am in favor of having the flexibility in our custom CardboardModal component which can be either used as a pure modal or dialog by customizing the props. What do you think?

src/Components/DataHistoryExplorerModal/DataHistoryExplorerModal.types.ts
```diff
@@ -0,0 +1,34 @@
+import { IStyle, IStyleFunctionOrObject, ITheme } from '@fluentui/react';
+import ADTDataHistoryAdapter from '../../Adapters/ADTDataHistoryAdapter';
+import MockAdapter from '../../Adapters/MockAdapter';
+import { ICardboardModalStyles } from '../CardboardModal/CardboardModal.types';
+
+export interface IDataHistoryExplorerModalProps {
+    adapter: ADTDataHistoryAdapter | MockAdapter;
+    isOpen: boolean;
+    onDismiss?: () => void;
```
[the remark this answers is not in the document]
iremgokce: for "modal" component we do not have to provide a dismiss callback, modal visibility is controlled by another prop (`isOpen`), onDismiss callback is just another level of method which is called when the modal is dismissed. It should be optional as in the original modal fluent component: 
image

src/Models/Constants/Interfaces.ts
```diff
@@ -939,8 +940,13 @@ export interface IBlobServiceCorsRule {
...
 }
 
-export interface IDataHistoryWidgetTimeSeriesTwin {
+export interface IDataHistoryTimeSeriesTwin {
     label?: string;
     twinId: string;
     twinPropertyName: string;
+    twinPropertyType: PropertyValueType;
+    chartProps?: {
+        color?: string;
```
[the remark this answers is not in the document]
iremgokce: no, these individual chart related props should be separately controlled and all optional. chartProps itself is also optional if you want to specify a property in it for the series. does it make sense?

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.types.ts
iremgokce: it is either "add" or "edit". Since we pass the timeSeriesTwin object as param, the consumer would know what to do with this information via the selected time series twin id if exist. you can check it out [here](https://github.com/microsoft/iot-cardboard-js/pull/807/files#diff-50759147c90dc1b8570a05ea0cae82a017031e59269992398ccd1ea23d0abf0eR57).

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.tsx
iremgokce: added beak and positioned the callout to right center

src/Components/DataHistoryExplorer/Internal/TimeSeriesBuilder/TimeSeriesBuilder.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+    // styles
+    const classNames = getClassNames(styles, {
+        theme: useTheme()
+    });
+
+    return (
+        <div className={classNames.root}>
+            <Stack tokens={{ childrenGap: 8 }}>
+                <h2 className={classNames.header}>
```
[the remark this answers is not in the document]
iremgokce: oh wow, good catch!

src/Components/DataHistoryExplorer/Internal/TimeSeriesTwinCallout/TimeSeriesTwinCallout.stories.local.tsx
msnyder-msft: hmm ok. I thought they'd be a bit more encapsulated and only the top parent would be making those fetches. What kind of data are we pushing or pulling from the list of series? I would have expected that to be mostly just rendering a list of items.

src/Models/Constants/Interfaces.ts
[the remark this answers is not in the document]
msnyder-msft: ok. I'm fine with the chartProps being optional, that makes sense that you wouldn't always want them, I just figured that if you were going to provide that object, you would want to provide all of the children since you're now saying you want to configure the chart. I'm generally hesitant to use a lot of optional props since it's not as clear to the consumer when they are actually required and when they aren't (and what happens when they are/aren't), so I tend to be fairly critical of them when I see them.

src/Components/DataHistoryExplorerModal/DataHistoryExplorerModal.types.ts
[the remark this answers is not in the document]
msnyder-msft: righttttt.... but if you don't provide the onDismiss, how can you ever close the modal?

src/Components/DataHistoryExplorerModal/DataHistoryExplorerModal.styles.ts
msnyder-msft: gotchaaa.... Could we just put those in a horizontal stack instead?

src/Components/CardboardModal/CardboardModal.types.ts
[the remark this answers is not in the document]
msnyder-msft: I think there should always be a primary action on the screen though right? Even if that action is to close the modal? Maybe we should make it possible to hide the cancel button instead? I can't think of a scenario we have in our app today where we would not want a primary button. What is the scenario you're thinking of here?

msnyder-msft (commented):
Round one looks great.

src/Components/CardboardModal/CardboardModal.types.ts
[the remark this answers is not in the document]
iremgokce: not necessarily since this is a pure modal which does not require a button at all: https://developer.microsoft.com/en-us/fluentui#/controls/web/modal

The scenario here is that, the "DataHistoryExplorerModal" is going to be triggered from property inspector to visualize some twin data history which does not require a primary action at all, just the close button (or only close button on the top right). This component is self contained in terms of interaction and no need to pass a callback or something. Only requires to be opened and dismiss which is what a modal does (this is not intended to be a dialog type interaction).

src/Models/Constants/Interfaces.ts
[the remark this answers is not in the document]
iremgokce: I see, but most of the time chartProps are like optional/customizable chart options that you want or don't want to set. It is like optional settings for user's time series. On the contrary, I think they should be all optional in our case since we already have a default way of rendering them in the chart. If a user want to override or specify anything, they will use this chartProps as entry point for those parameters to set. I prefer to keep it as optional here.

msnyder-msft:
I follow now. That totally makes sense. It could almost be chartOverrides or something like that. But not required. Thanks for explaining.

iremgokce:
> The label for the toggle feels very loud. Did we make that text bold ourselves or is that default? It draws my eyes a lot and I feel like it shouldn't. Not a deal breaker but I'd be curious if it might look better as normal weight. image

I don't think it feels like a action item with normal font weight, but maybe?:
image

msnyder-msft:
> > The label for the toggle feels very loud. Did we make that text bold ourselves or is that default? It draws my eyes a lot and I feel like it shouldn't. Not a deal breaker but I'd be curious if it might look better as normal weight. image
> 
> @msnyder-msft I don't think it feels like a action item with normal font weight, but maybe?: image

I actually think this looks pretty good. I feel like it's kind of a "sub-item" to the dropdown above so it doesn't need it's own visually bold label. I'm not deadset on it, but I like this screenshot.

iremgokce:
> the double scrollbars are not good. I think I saw you made a style change to the twin search. is that a side effect? image

I think I did not change styling for TwinPropertySearchDropdown specifically, but I fixed the double scroll issue here in this component and also fixed another issue which was rendering double "MenuList" Reach select component:
image

src/Components/DataHistoryExplorer/Internal/TimeSeriesBuilder/TimeSeriesBuilder.types.ts
```diff
@@ -0,0 +1,38 @@
+import { IStyle, IStyleFunctionOrObject, ITheme } from '@fluentui/react';
+import { IDataHistoryTimeSeriesTwin } from '../../../../Models/Constants';
+import { ITimeSeriesTwinCalloutStyles } from '../TimeSeriesTwinCallout/TimeSeriesTwinCallout.types';
+
+export const MAX_NUMBER_OF_TIME_SERIES_TWINS = 3;
+export const TIME_SERIES_TWIN_LIST_ITEM_ID_PREFIX =
+    'cb-data-history-explorer-series';
+
+export interface ITimeSeriesBuilderProps {
+    onTimeSeriesTwinListChange?: (
```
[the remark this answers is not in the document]
iremgokce: it is to update the shared context with the viewer with the recent timeSeriesTwinList. So, whenever the list is changed in the builder, the new list of series is passed to viewer through the shared context.

iremgokce:
> > the double scrollbars are not good. I think I saw you made a style change to the twin search. is that a side effect? image
> 
> I think I did not change styling for TwinPropertySearchDropdown specifically, but I fixed the double scroll issue here in this component and also fixed another issue which was rendering double "MenuList" Reach select component: UPDATE: this is a weird issue, still debugging... image

@msnyder-msft it seems like this requires a separate PR, I do not want to include the fixes in this. React select in twin search requires some revision.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
iremgokce: As I explained in another comment, there is no need to distinguish for add/remove here. onTimeSeriesTwinListChange is used to update the shared context with the viewer whenever the timeSeriesTwinList changes regardless add/remove. So, whenever the list is changed in the builder, the new list of series is passed to viewer through the shared context to render the series in chart.

msnyder-msft (approved):
LGTM

PR #816, opened 2023-01-06
Check for correct snapshot after delay

pasanchMSFT:
### Summary of changes 🔍 
Checking if Chromatic snapshot yields the same results as Storybook after some delay.

### Testing 🧪
View 3DV/ADT3DViewer/Engine story in Storybook and compare with the result Chromatic gives us.

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Components/ADT3DViewer/ADT3DViewer.stories.tsx
+++ b/src/Components/ADT3DViewer/ADT3DViewer.stories.tsx
@@ -36,19 +36,19 @@ type ADT3DViewerStory = ComponentStory<typeof ADT3DViewer>;
 const Template: ADT3DViewerStory = (args) => {
     const authenticationParameters = useAuthParams();
     const scenesConfig = mockVConfig as I3DScenesConfig;
+    const adapter = new MockAdapter();
 
     return !authenticationParameters ? (
         <div></div>
     ) : (
-        <ADT3DViewer scenesConfig={scenesConfig} {...args} />
+        <ADT3DViewer adapter={adapter} scenesConfig={scenesConfig} {...args} />
     );
 };
 
 export const Engine = Template.bind({});
 Engine.args = {
     title: '3D Viewer',
-    adapter: new MockAdapter(),
-    sceneId: { mockSceneId },
+    sceneId: mockSceneId,
     connectionLineColor: '#000'
 };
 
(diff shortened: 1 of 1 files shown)


src/Components/ADT3DViewer/ADT3DViewer.stories.tsx
```diff
@@ -22,7 +23,10 @@ export default {
 
 const mockSceneId = 'f7053e7537048e03be4d1e6f8f93aa8a';
 
-export const Engine = (_args, { globals: { theme, locale } }) => {
+export const Engine: ComponentStoryFn<typeof ADT3DViewer> = (
```
iremgokce, line 26: can we change it to the new style of decorating for theme and localization and using ComponentStory instead of ComponentStoryFn? not sure the difference though...
image

src/Components/ADT3DViewer/ADT3DViewer.stories.tsx
```diff
@@ -41,6 +45,10 @@ export const Engine = (_args, { globals: { theme, locale } }) => {
         </div>
     );
 };
+Engine.play = async ({ _canvasElement }) => {
+    //👇 This sets a timeout of 2s
+    await new Promise((resolve) => setTimeout(resolve, 2000));
```
iremgokce, line 50: we might need to extend this timeout depending on the model file size, right, to wait it to load before taking the snapshot? if so, we can add a comment for it to remember if we change the mock file.

iremgokce (approved):

src/Components/ADT3DViewer/ADT3DViewer.stories.tsx
pasanchMSFT: Looks like ComponentStoryFn === ComponentStory, in our Storybook version
image

src/Components/ADT3DViewer/ADT3DViewer.stories.tsx
pasanchMSFT: I just set arbitrary time, but yeah, we might need to change this based on file size. I'll add a comment

src/Components/ADT3DViewer/ADT3DViewer.stories.tsx
iremgokce: ohh interesting

PR #821, opened 2023-01-11
Ensure that we write the configuration value for local storage on deeplink as a JSON object using appAdtUrl instead of as a string literal

darsney:
See title, simple as that


pasanchMSFT (approved):

msnyder-msft (approved):

montgomp (approved):
LGTM

PR #832, opened 2023-01-14
Iteratively remove models when they fail the parser

darsney:
### Summary of changes 🔍 
ADT allows models to have duplicate property names in object definitions in DTDL.  This is a problem, because our parser implementation forbids this.  The parser may also fail in other cases that are supported in ADT, but this is the only known case thus far.

When parsing models, if any model fails, the parser will return null.  In order to at least resolve _some_ of the valid models, we attempt to remove offending models by iteratively calling the parser by removing the models that do not parse.  In general, the parser implementation fails only one model at a time, so this may require multiple iterations.  This is a performance concern if the environment has a large number of offending models, but I think its a good balance to attempt to gracefully degrade with this recursive function since the experience degrades _so poorly_ if one of the models doesn't parse.

We limit iteration to 20 iterations, and only iterate if we successfully removed at least one model and there are models left

### Testing 🧪
I have added models with duplicate properties to MockModelData.json.  You can test the ModelledPropertyBuilder and see the associated console logging that describes the parse issues.  If you were to use this mock data in main, the ModelledPropertyBuilder dropdown would be empty because the parser would fail.  

See here: 
https://601c6b2fcd385c002100f14c-xayuksuoqh.chromatic.com/?path=/story/components-modelledpropertybuilder--toggle-mode

The console warnings for parsing are pretty printed like this...
image

Try it for yourself in a real environment by adding the models from the MockModelData file to your environment, and compare the hosted version with the work that I've done here.

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Models/Services/Utils.ts
+++ b/src/Models/Services/Utils.ts
@@ -83,13 +83,13 @@ export const parseDTDLModelsAsync = async (dtdlInterfaces: DtdlInterface[]) => {
                 }
             );
 
-            /* Recursively parse models if we successfully removed some models, and there are still some left */
+            /* Iteratively parse models if we successfully removed some models, and there are still some left */
             if (
                 interfacesWithoutParserErrors.length <
                     previousInterfaces.length &&
                 interfacesWithoutParserErrors.length > 0
             ) {
-                console.log('Removing model that failed to parse and retrying');
+                console.log('Removing models that failed to parse and retrying');
                 modelDict = await tryParseWorkerFunction(
                     interfacesWithoutParserErrors
                 );
--- a/src/Models/Services/Utils.ts
+++ b/src/Models/Services/Utils.ts
@@ -43,10 +43,65 @@ const parser = createParser(ModelParsingOption.PermitAnyTopLevelElement);
 
 /** Parse DTDL models via model parser */
 export const parseDTDLModelsAsync = async (dtdlInterfaces: DtdlInterface[]) => {
-    const modelDict = await parser.parse(
-        dtdlInterfaces.map((dtdlInterface) => JSON.stringify(dtdlInterface))
-    );
+    let modelDict = null;
+    try {
+        modelDict = await parser.parse(
+            dtdlInterfaces.map((dtdlInterface) => JSON.stringify(dtdlInterface))
+        );
+    } catch (e) {
+        /* Some models failed to parse, we will try to remove them and parse the rest of the models */
+        logDtdlParserError(e);
+        try {
+            const parsingErrorCauses = e?._parsingErrors.map(
+                (pe) => pe?.cause
+            ) as Array<string>;
+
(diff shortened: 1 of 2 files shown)


src/Models/Services/Utils.ts
```diff
@@ -43,12 +43,67 @@ const parser = createParser(ModelParsingOption.PermitAnyTopLevelElement);
...
+            console.log(e);
+        }
+    }
     return modelDict;
 };
 
+/* Pretty prints parser errors to the console to aid customer debugging */
+export const logDtdlParserError = (parserErrors) => {
+    try {
+        console.group('DTDL Parser Errors');
```
msnyder-msft, line 87: I never knew you could do something like this. Fascinating!!

src/Models/Services/Utils.ts
```diff
@@ -43,12 +43,67 @@ const parser = createParser(ModelParsingOption.PermitAnyTopLevelElement);
...
+                ).length;
+            });
+
+            // Recursively parse models if we successfully removed some models, and there are still some left
+            if (
+                modelsWithoutParserErrors.length < dtdlInterfaces.length &&
+                modelsWithoutParserErrors.length > 0
+            ) {
+                console.log('Removing model that failed to parse and retrying');
+                return await parseDTDLModelsAsync(modelsWithoutParserErrors);
```
msnyder-msft, line 74: do we want to keep track of some kind of depth so we can bust out if we get too far in?

src/Models/Services/Utils.ts
```diff
@@ -43,12 +43,67 @@ const parser = createParser(ModelParsingOption.PermitAnyTopLevelElement);
...
+        );
+    } catch (e) {
+        /* Some models failed to parse, we will try to remove them and parse the rest of the models */
+        logDtdlParserError(e);
+        try {
+            const parsingErrorCauses = e?._parsingErrors.map(
+                (pe) => pe?.cause
+            ) as Array<string>;
+
+            /* The parsing error cause is not a model ID, but it contains a model ID
```
msnyder-msft, line 59: Can't we look for a particular message instead of just seeing that it contains an id?

msnyder-msft (approved):

msnyder-msft (commented):
Sounds like you're going to take a different approach so I'll wait to see that.

src/Models/Services/Utils.ts
darsney: See the raw error log, the only thing that contains the actual DTID is unfortunately the error cause message

src/Models/Services/Utils.ts
darsney: yup, added depth of 20

pasanchMSFT (approved):
LGTM, recursive approach was the only concern for me previously, but this looks good.

msnyder-msft (approved):

PR #840, opened 2023-01-25
[Data History Explorer]: Telemetry and Logging

iremgokce:
### Summary of changes 🔍 
- Added telemetry for Data History Explorer (for both system and user actions) and logging of actions. Note that I added a separate section for `DataHistoryExplorer` in the `TelemetryConstants` file since it is a independent entity as an orthogonal tool to the 3D scene app (different than Adapters, Builder and Viewer app regions). If you have concerns about it we can discuss more.
- **To-do:** I am going to emit `OpenModal` user action telemetry when I bind the modal to property inspector soon.

image
image

Here is some examples:
image

### Testing 🧪

1. Enable telemetry logging feature in your local storage using `localStorage.setItem('cardboard.debug.telemetryLogging',true)`;
2. Enable "Verbose" to see the telemetry logging in the inspector on the page
3. Go to Components/DataHistoryExplorer/Modal ADT story and see the logs and telemetry

### Checklist ✔️
- [ ] Linked associated issue (if present)
- [x] Added relevant labels
- [x] Requested developer reviews
- [ ] Request UI review from design / PM
- [x] Verify all code checks are passing

PR #842, opened 2023-01-27
[Data History Explorer]: Error Handling

iremgokce:
### Summary of changes 🔍 
_Note:_ This branch is off the [Telemetry branch](https://github.com/microsoft/iot-cardboard-js/pull/840), merge that one first for less file diffs.
- Created a standalone component **`DataHistoryErrorHandlingWrapper`** to handle ADX service specific errors and `TimeSeriesDatabaseConnectionFetchFailed` adapter error.
   - This component is consumed by `DataHistoryWidget` and `DataHistoryExplorer` now. 
- Added series level "noData" errors if there is no ADX data to display in the current view (either in Chart or Table)
   - For already numeric properties it might be due to the selected time range
   - For non numeric properties but which is set as "cast to numeric", the ADX still may return to null since with this toggle we add toDouble() in the ADX query for the Value column but which can still return null if the value is not castable to double (e.g. even though a boolean property is casted to numeric in setting, the toDouble('True') will return null since it is not castable to a number, but when a string property which stores a numeric value is casted to numeric toDouble('123') will return data like 123).
- Changed some translation for messages to better reflect the errors
- Here is the list of handled errors (I put the references into the DataHistoryErrorHandlingWrapper component):
   - **Cluster connection** (this happens when ADX request fails with status 0 which might be due to invalid cluster url, stopped cluster or network connectivity error) 
   - **Database connection** (this happens with specific ADX return error named '_BadRequest_EntityNotFound_' with status 400)
   - **Table or query error** (since table is part of the query payload in the request, and this happens with specific ADX return error named '_General_BadRequest_' with status 400)
   - Missing permission to query (this happens when ADX request fails with error code 403)
   - Additionally, I added to handle '_TimeSeriesDatabaseConnectionFetchFailed_' custom adapter error (this happens when `getTimeSeriesConnectionInformation` from AzureManagementAdapter fails to fetch the connection information of the selected ADT instance which might be due to control plane access or if there is no connection setup for the instance)

**To-do:** I am going to do content pass with @azimamansuri and @rinisbet 
image
image
image
image
image

### Testing 🧪
- Check `DataHistoryErrorHandlingWrapper` individual stories, or
- Check Components/DataHistoryExplorer/Internal/TimeSeriesViewer/Errors stories or
- Components/DataHistoryExplorer/Modal/TimeSeriesConnectionInformationFetchFailed or
- Check the whole integration on Components/DataHistoryExplorer/Modal/ADT with actual data since MockAdapter always return a mock data currently.
- You can test some of the errors by:
   - Set a no data history enabled adt instance url in your secrets file
   - Stop your cluster that your adt instance connected
      -  It is hard to test database and table errors since you cannot change the name of those or stop the service but you can try to delete a data connection which I didn't try yet.
   - Try to select series which you think don't have data in selected time frame or data type would not return to numeric even if it is set as casted (see my comment above about those). 
   - Try to remove your permission to query from the cluster which is again hard to test because we all have Contributor and Owner assignments due to security group access. You can try to login the stories with a guest user account though. In that case you would probably see the _TimeSeriesDatabaseConnectionFetchFailed_ error though.

### Checklist ✔️
- [x] Added relevant labels


src/Resources/Locales/en.json
```diff
@@ -1392,11 +1395,20 @@
...
+                    "noData": "No historized data found in the selected time range.",
                     "noSeries": "No twin selected to query, please add twin properties to view historized data."
                 }
             },
             "noSeriesDescription": "To view data history for this twin, select a property to view."
         }
+    },
+    "dataHistoryErrorHandlingWrapper": {
+        "errors": {
+            "genericNetwork": "Cluster cannot be connected which may be due to invalid url, stopped service or a network error.",
```
msnyder-msft, line 1407: suggestion: `Cluster cannot be reached. This may be due....`

src/Resources/Locales/en.json
```diff
@@ -1392,11 +1395,20 @@
...
+            "genericBadRequest": "Bad request exception occured.",
```
msnyder-msft, line 1408: do we have to include the bad request part? THat's very jargony. Can we be generic and say `Something went wrong, please try again` or `An exception occurred with the request`. I personally prefer #1.

src/Resources/Locales/en.json
```diff
@@ -1392,11 +1395,20 @@
...
+            "generalBadRequestMessage": "Request is invalid and cannot be executed.",
```
msnyder-msft, line 1409: Is this because of something the user did?

src/Resources/Locales/en.json
```diff
@@ -1392,11 +1395,20 @@
...
+            "entityNotFound": "Database not found, failed to execute the query.",
```
msnyder-msft, line 1410: instead of failed, can we say `... cannot execute..`

src/Resources/Locales/en.json
```diff
@@ -1376,11 +1373,17 @@
                 "edit": "Edit time series",
                 "remove": "Remove time series",
                 "property": "Property",
-                "propertyTypeCastLabel": "Cast property type to number"
+                "propertyTypeCastLabel": "Cast property type to number",
```
msnyder-msft, line 1376: suggestion: `Cast property value to number`

src/Resources/Locales/en.json
```diff
@@ -1376,11 +1373,17 @@
...
+                "noData": "This time series contains no data to display"
```
msnyder-msft, line 1377: can we trim this? `This time series contains no data`

src/Resources/Locales/en.json
```diff
@@ -1376,11 +1373,17 @@
...
             }
         },
         "viewer": {
-            "chart": "Chart",
+            "chart": {
+                "title": "Chart",
+                "messages": {
+                    "noData": "No numeric historized data to display, try selecting a different time range."
```
msnyder-msft, line 1384: can we shorten this up a bit? `No data was found for the selected time range`

src/Resources/Locales/en.json
```diff
@@ -1392,11 +1395,20 @@
                 "messages": {
                     "noColor": "Not set",
                     "loading": "Loading historized data...",
-                    "noData": "No historized data found in selected time frame.",
+                    "noData": "No historized data found in the selected time range.",
```
msnyder-msft, line 1398: can we remove `historized`? `No data found for the selected time range`.
This seems like it should/could match up with the message above (in the loc file) for `viewer.chart.messages.noData`

src/Resources/Locales/en.json
```diff
@@ -116,6 +116,10 @@
         "schemaValidationFailed": {
             "type": "JSON schema validation failed",
             "message": "JSON is invalid. Check syntax matches required schema."
+        },
+        "timeSeriesDatabaseConnectionFetchFailed": {
+            "type": "Time series database connection validation failed",
+            "message": "Data history is not enabled for selected instance or time series database connection could not be verified."
```
msnyder-msft, line 122: Can we add in 'the' in a few spots to make it a little smoother? 
`Data history is not enabled for _the_ selected instance or _the_ time series database connection could not be verified.`

src/Resources/Locales/en.json
```diff
@@ -116,6 +116,10 @@
         "schemaValidationFailed": {
             "type": "JSON schema validation failed",
             "message": "JSON is invalid. Check syntax matches required schema."
+        },
+        "timeSeriesDatabaseConnectionFetchFailed": {
+            "type": "Time series database connection validation failed",
```
msnyder-msft, line 121: should we say "failed" in this case? That makes it sound fatal but we really just mean "we couldn't validate it, keep on doing your thing if you like" Just seems like it's a dead end when I believe it's more of a warning?

src/Models/Hooks/useTimeSeriesData.ts
```diff
@@ -137,16 +138,20 @@ export const useTimeSeriesData = ({
         }
     };
 
+    useEffect(() => {
+        setIsLoading(timeSeriesData.isLoading);
```
msnyder-msft, line 142: do we still need local state if it will always refelct the time series data value? Can we just use it directly?

src/Components/DataHistoryExplorer/Internal/TimeSeriesViewer/Internal/TimeSeriesTable/TimeSeriesTable.tsx
```diff
@@ -222,6 +227,16 @@ const TimeSeriesTable: React.FC<ITimeSeriesTableProps> = (props) => {
         );
         setItems(adxTimeSeriesTableRows);
     }, [data]);
+    useEffect(() => {
+        if (data && !isLoading) {
+            const seriesWithNoData = timeSeriesTwinList
+                ?.filter((ts) => !data?.find((d) => d.seriesId === ts.seriesId))
+                .map((ts) => ts.seriesId);
+            if (onMissingSeriesData) {
+                onMissingSeriesData(seriesWithNoData);
```
msnyder-msft, line 236: why do we need to do this? and do we really want to call it on every render where something changes? It has a weird feel that we need this... 🤔

src/Components/DataHistoryExplorer/Internal/TimeSeriesViewer/Internal/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -114,28 +131,41 @@ const TimeSeriesChart: React.FC<ITimeSeriesChartProps> = (props) => {
...
-                        legendPadding: 0,
-                        hasMultipleAxes:
-                            chartOptions.yAxisType === 'independent',
-                        dataGrouping: chartOptions.aggregationType,
-                        xMinInMillis: xMinDateInMillisRef.current,
-                        xMaxInMillis: xMaxDateInMillisRef.current,
-                        maxLegendHeight: 160
-                    }}
+            {errors.length > 0 ? (
+                <DataHistoryErrorHandlingWrapper
```
msnyder-msft, line 135: so the wrapper is a different component that is a sibling to the other stuff? I guess the name `Wrapper` implies that it is like an error boundary and surrounds the child components. If that's not hte case, could we call it "DataHistoryErrorModal/Dialog" or something like that?

src/Components/DataHistoryExplorer/Internal/TimeSeriesViewer/Internal/TimeSeriesChart/TimeSeriesChart.tsx
```diff
@@ -106,6 +113,16 @@ const TimeSeriesChart: React.FC<ITimeSeriesChartProps> = (props) => {
             onChartOptionsChange(chartOptions);
         }
     }, [chartOptions]);
+    useEffect(() => {
```
msnyder-msft, line 116: it feels weird that we need to do this same logic in the chart as the table component. Could we instead feed the data into those components and do this in the parent? That way we don't have both components fetching data separately? That seems like a good way to end up with the chart doing some different manipulation than the table and they can get out of sync.

msnyder-msft (commented):
I think this seems pretty solid overall. As we discussed on the call, I think it's worth thinking about pulling our data fetching up higher in the stack. Right now, the table & chart both fetch data and now that the builder cares about the data that was fetched (to show the no data message) it seems prudent to pull that fetching and state for the fetched data up higher (potentially all the way into the context since it's needed in both Viewer and Builder) and then we can pass it down to the rendering components. That will ensure that all components are working off the same data and won't trigger data fetches when switching between the tabs on the right.
If you decide to go that route, I would probably do that in a separate PR from this one though unless you think the error handling would be heavily impacted by this change.
One added bonus here is that if we split off the fetching from the rendering, it will let us reuse the table & chart components in the future and plug in a different query if we need to use them elsewhere in the app without having to rebuild them or plumb through weird props that let you change the query.

src/Resources/Locales/en.json
iremgokce: changed to "Azure Data Explorer instance cannot be connected, make sure the resource is not stopped and your network connectivity is stable." from content review by Azima and Riley

src/Components/DataHistoryErrorHandlingWrapper/DataHistoryErrorHandlingWrapper.tsx
```diff
@@ -136,18 +157,18 @@ export const DataHistoryWidgetErrorHandling: React.FC<IDataHistoryWidgetErrorHan
...
 
     return (
-        <div style={classNames.errorContainer.root}>
+        <div className={classNames.root}>
             <IllustrationMessage
                 descriptionText={errorObj.description}
                 type={'info'}
                 width={'compact'}
                 imageProps={{
                     src: errorObj.imgSrc,
```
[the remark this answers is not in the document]
iremgokce: it is for to be consistent in height no matter what the error and its image is. errorObjs is only to decide the img src and description. height is pass as prop to this component

src/Resources/Locales/en.json
iremgokce: Since I got the errors from [here](https://learn.microsoft.com/en-us/azure/data-explorer/kusto/api/netfx/kusto-data-client-errors#bad-request-exceptions), I wanted to reflect the error messages as is but I can change it to that one

src/Resources/Locales/en.json
iremgokce: To capture General_BadRequest error code from Kusto: https://learn.microsoft.com/en-us/azure/data-explorer/kusto/api/netfx/kusto-data-client-errors#request-exceptions

The content of the messages what the error message says in the response.
image

src/Resources/Locales/en.json
iremgokce: we used "validation failed" in somewhere else, that is why I used it this way. I can change it to "Time series database connection could not be validated".

src/Models/Hooks/useTimeSeriesData.ts
iremgokce: since I am exposing isLoading from the useTimeSeriesData hook it needs to be a state variable to let the parent know about a change in the exposed property. Otherwise if I expose isLoading as timeSeriesData.isLoading it would not be triggering consumer of the hook to catch this change.

src/Resources/Locales/en.json
iremgokce: it is changed to something else based on the content feedback by Azima and Riley

src/Resources/Locales/en.json
iremgokce: based on content feedback we changed it to "No historized data found in the selected time range." but I kept the numeric part in the translation for chart since it might be misleading if we say "no historized data", in fact it is no numeric data found to display in the chart. fyi @azimamansuri @rinisbet

src/Resources/Locales/en.json
iremgokce: after content pass with Azima and Riley it is "No numeric historized data found in the selected time range" for chart, "No historized data found in the selected time range" for Table. I think it is good to mentioned historized since this is timeseries data stored in Kusto. What do you think @azimamansuri @rinisbet , I can remove that part if you all agree.

src/Resources/Locales/en.json
iremgokce: after content pass it is changed to "Data history connection information could not be found, make sure data history is enabled for this Azure Digital Twins instance and that you have at least '{{permission}}' access to the Azure Data Explorer instance."

src/Components/DataHistoryExplorer/Internal/TimeSeriesViewer/Internal/TimeSeriesTable/TimeSeriesTable.tsx
iremgokce: as we discussed, since in the current design we fetch data separately in Chart and Table, whenever there is change in data, or list we need to make the parent know about series with no data, we need to bubble up the missing series to the parent so that higher up sibling component (TimeSeriesBuilder) would update the list to show the warning icon with noData label.

When I implement single data fetch logic in the upcoming PR, I will not need that since I will feed all the internal components with the same data and the do not need to communicate with each other to show the consistent noData message.

src/Components/DataHistoryExplorer/Internal/TimeSeriesViewer/Internal/TimeSeriesChart/TimeSeriesChart.tsx
iremgokce: I did the same thing in ScenePageErrorHandlingWrapper. I think it make sense since it is wrapping the IllustrationMessage component based on the component errors of adapter calls. So there is a specific internal logic to control the IllustrationMessage child component. Does it make sense? I do not have a strong preference though, can change it to "DataHistoryErrorHandlingIllustration"?

src/Components/DataHistoryExplorer/Internal/TimeSeriesViewer/Internal/TimeSeriesChart/TimeSeriesChart.tsx
iremgokce: Yeah since both Chart and Table is children of TimeSeriesViewer which needs to pass this missing data information to TimeSeriesBuilder on the left to show the triange noData icon for individual series, I needed to bubble up series with no data from places where I do data fetching, in which case it is both Chart and Table. After I make the refactor there will be no need for this, wait for the following PR please...

iremgokce:
> I think this seems pretty solid overall. As we discussed on the call, I think it's worth thinking about pulling our data fetching up higher in the stack. Right now, the table & chart both fetch data and now that the builder cares about the data that was fetched (to show the no data message) it seems prudent to pull that fetching and state for the fetched data up higher (potentially all the way into the context since it's needed in both Viewer and Builder) and then we can pass it down to the rendering components. That will ensure that all components are working off the same data and won't trigger data fetches when switching between the tabs on the right. If you decide to go that route, I would probably do that in a separate PR from this one though unless you think the error handling would be heavily impacted by this change. One added bonus here is that if we split off the fetching from the rendering, it will let us reuse the table & chart components in the future and plug in a different query if we need to use them elsewhere in the app without having to rebuild them or plumb through weird props that let you change the query.

As we discussed, will change the data fetching from individual components to single one in the TimeSeriesViewer parent component in the next PR.

src/Resources/Locales/en.json
msnyder-msft: I'm confused. Why do we have it localized if it's not sured for showing to the user?

src/Models/Classes/Errors.ts
```diff
@@ -56,7 +56,8 @@ class ComponentError extends Error implements IComponentError {
                 return i18n.t('errors.schemaValidationFailed.message');
             case ComponentErrorType.TimeSeriesDatabaseConnectionFetchFailed:
                 return i18n.t(
-                    'errors.timeSeriesDatabaseConnectionFetchFailed.message'
+                    'errors.timeSeriesDatabaseConnectionFetchFailed.message',
+                    { permission: 'Reader' }
```
msnyder-msft, line 60: this should be localized no?

src/Components/DataHistoryExplorer/Internal/TimeSeriesViewer/Internal/TimeSeriesChart/TimeSeriesChart.tsx
msnyder-msft: Yeah. I personally prefer that, but don't feel strongly.

src/Components/DataHistoryErrorHandlingWrapper/DataHistoryErrorHandlingWrapper.types.ts
```diff
@@ -7,11 +7,15 @@ import {
...
-import { IllustrationMessageStyles } from '../IllustrationMessage/IllustrationMessage.types';
+import {
+    IllustrationMessageStyles,
+    IllustrationMessageWidth
+} from '../IllustrationMessage/IllustrationMessage.types';
 
 export interface IDataHistoryErrorHandlingWrapperProps {
     error: IComponentError;
     imgHeight?: number;
+    messageWidth?: IllustrationMessageWidth;
```
msnyder-msft, line 18: why do we need this? Can't the message be different lengths based on teh language?

msnyder-msft (approved):

src/Models/Classes/Errors.ts
iremgokce: I wanted to keep the permission name out of the translation since it might be a bad translation. In the portal right now, we can see the name of permissions translated if another language is selected but a very specific way. We can have a list of permanent translation section in our language files or another place like enums.

src/Components/DataHistoryErrorHandlingWrapper/DataHistoryErrorHandlingWrapper.types.ts
iremgokce: this is the enum for the width of the message container, either "compact" or "wide". It is originally a prop to the IllustrationMessage component, just exposed it in the parent to pass along.

msnyder-msft: gotcha. ok. makes sense.

src/Models/Classes/Errors.ts
msnyder-msft: yep. makes sense. Let's leave it as is since that's the permission name.

PR #853, opened 2023-02-03
OAT: Refactor import/export logic to public utils file

msnyder-msft:
### Summary of changes 🔍 
Pulling out the import, export and parsing logic to a public utility file. It doesn't change the functionality, just moves it to a common place so we can iterate and expose to the ADT Explorer as well in the future.

### Testing 🧪
- import a file
- import a folder
- Import a file that already exists in the ontology, see the errors
- Import a file that is not a JSON file
- Import a file that is a JSON file but the contents don't parse properly
- export ontology

### DIFF
--- a/src/Components/OATHeader/OATHeader.tsx
+++ b/src/Components/OATHeader/OATHeader.tsx
@@ -19,7 +19,7 @@ import {
 import { useTranslation } from 'react-i18next';
 import JSZip from 'jszip';
 import { CommandHistoryContext } from '../../Pages/OATEditorPage/Internal/Context/CommandHistoryContext';
-import { getDebugLogger, parseModels } from '../../Models/Services/Utils';
+import { getDebugLogger } from '../../Models/Services/Utils';
 import {
     HeaderModal,
     IOATHeaderProps,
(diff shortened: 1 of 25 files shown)


src/Components/OATHeader/OATHeader.tsx
```diff
@@ -76,204 +92,92 @@ const OATHeader: React.FC<IOATHeaderProps> = (props) => {
...
+            zipResult.file.generateAsync({ type: 'blob' }).then((content) => {
+                logDebugConsole(
+                    'info',
+                    '[END] Export models to file. {content}',
+                    content
+                );
+                downloadFile(
+                    content,
+                    `${
+                        oatPageState.currentOntologyProjectName || 'ontology'
```
montgomp, line 153: Do we need to be concerned with ProjectName containing invalid filename characters here?

msnyder-msft: oh that's an interesting point. What characters would that be? I guess I could stick to the hard coded name to be safe. This was just a nice cosmetic thing but I doubt there is a good regex that would help us to sanitize those.

montgomp: Something like `^[a-zA-Z0-9](?:[ a-zA-Z0-9._-]*[a-zA-Z0-9])?$`  allows upper/lower/numeric to start; allows numbers, spaces, etc, within the string; doesn't allow trailing spaces or periods

msnyder-msft: that seems fair.

msnyder-msft: Added.

montgomp (approved):
LGTM

pasanchMSFT (approved):
LGTM, will circle back to test utils/files to re-use patterns.

PR #856, opened 2023-02-06
Remove all cruft. cards, boards, and irrelevant pages and irrelevant packages

darsney:
### Summary of changes 🔍 
This PR removes a lot of the early work in cardboard that is no longer relevant.  Stuff like Cards, Boards, BIM experiments, references to TSI and Central, old components that use patterns we no longer use like SaSS, stuff like that.

### Testing 🧪
Nothing should be impacted by these changes, but @iremgokce should note that I did need to update one snapshot


src/Components/ModelCreate/ElementsList.scss
```diff
@@ -1,11 +0,0 @@
-.cb-elementslist-row {
```
darsney, line 1: Okay i gotta undelete this somehow, seems like modelcreate is used?

src/Components/CardboardListCallout/__snapshots__/CardboardListCallout.test.tsx.snap
```diff
@@ -2,13 +2,13 @@
 
 exports[`CardboardListCallout entered search term passed successfully on primary button click 1`] = `
 <div
-  class="ms-Callout root-131"
+  class="ms-Callout root-129"
```
darsney, line 5: @iremgokce I am really not sure why this file changed, but I did need to update the snapshot to get it to work.  Seemed to me like it was kind of a fluke, maybe just a function of me deleting a million things that changed the numbers here, but the stories looked correct to me

msnyder-msft (approved):

pasanchMSFT (approved):
LGTM

src/Components/CardboardListCallout/__snapshots__/CardboardListCallout.test.tsx.snap
iremgokce: so yeah this is what I mentioned by saying className change. So I thought I already updated the snapshot with the change in classnames changes. So lets see what you have after merge main

PR #857, opened 2023-02-06
OAT: Various bug fixes for the graph nodes & property setting

msnyder-msft:
### Summary of changes 🔍 
- Refactor utility to simplify logic around detecting version of a model
- Remove trashcan from "Untargeted" nodes since we can't know which "untargeted" node to delete. They can only remove that node by removing the reference instead. Which gets selected when you click the node so that should be fine.
image

- Fix bug where model versions can be set to negative values
- Fix bug where userare are unable to rename a property after adding a reference to a model
- Removed outdated logic for updating all existing models when the default path is changed. This should have been removed when we changed from using this as a global namespace but was missed.

### Testing 🧪
- add a reference to a model and then try to add and rename a property on the parent model. This used to not work, it renamed the reference instead.
- Try setting a model version to 0 or negative
- Try removing an untargeted node from the graph

### Checklist ✔️
- [ ] Added relevant labels


montgomp (approved):
LGTM

src/Components/OATGraphViewer/Internal/OATGraphCustomNode.tsx
```diff
@@ -92,6 +93,7 @@ const OATGraphCustomNode: React.FC<IOATGraphCustomNodeProps> = (props) => {
 
     const onDelete = () => {
         const deletion = () => {
+            console.log('***Data', data);
```
montgomp, line 96: Should this be using debug logger?

montgomp (approved):
Reapproving post updates

pasanchMSFT (approved):

PR #858, opened 2023-02-07
[Data History Explorer]: Revised data fetching logic and components

iremgokce:
### Summary of changes 🔍 
- Data fetching and state management are moved up to single parent model component (DataHistoryExplorer) and all internal components are used as view components to render the data using the props all the way passed down
   - Moved code around and removed internal components which no longer needed
- Added **refetch** command action
- Added mock time series data
- Design cleanup based on final feedback

Here is the link for model-view components: [Figma](https://www.figma.com/file/Voxv1SVyF3jSvMneqP6F76/Data-History-Widget?node-id=704%3A242&t=RqmS0sJMtEuLCHdz-4)
image
image
image

### Testing 🧪
You can try it on DataHistoryExplorerModal component stories.

### Checklist ✔️
- [x] Added relevant labels


src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
```diff
@@ -39,24 +80,371 @@ const DataHistoryExplorer: React.FC<IDataHistoryExplorerProps> = (props) => {
...
+            dispatch,
+            sendDataHistoryExplorerUserTelemetry,
+            getDefaultSeriesLabel,
+            getHighChartColor,
+            state.timeSeriesTwins,
+            state.selectedTimeSeriesId
+        ]
+    );
+
+    const handleOnAddSeriesClick = useCallback(
```
msnyder-msft, line 278: great stuff. love what you did here with the callbacks and reducing the dependency on the context in the lower components. Great idea.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
```diff
@@ -39,24 +80,371 @@ const DataHistoryExplorer: React.FC<IDataHistoryExplorerProps> = (props) => {
...
+                            (tsTwin) => tsTwin.seriesId === seriesId
+                        ).chartProps.color
+                ),
+                1
+            );
+        },
+        [state.timeSeriesTwins, dispatch, sendDataHistoryExplorerUserTelemetry]
+    );
+
+    const handleOnViewModeChange = useCallback(
```
msnyder-msft, line 338: love how you have coupled the handlers to the telemetry. This is great. Very nicely done. Love to see the instrumentation.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
```diff
@@ -39,24 +80,371 @@ const DataHistoryExplorer: React.FC<IDataHistoryExplorerProps> = (props) => {
...
+            logDebugConsole(
+                'debug',
+                `Chart options changed: ${JSON.stringify(chartOptions)}`
+            );
+            const telemetry =
+                TelemetryEvents.Tools.DataHistoryExplorer.UserAction
+                    .ChangeChartOption;
+            sendDataHistoryExplorerUserTelemetry(telemetry.eventName, [
+                {
+                    [telemetry.properties.chartOptions]: chartOptions
```
msnyder-msft, line 371: great idea to capture these, just make sure we can't end up with anything sensitive or data from their chart in this like a series name or something that might be related to their actual twin.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
```diff
@@ -39,24 +80,371 @@ const DataHistoryExplorer: React.FC<IDataHistoryExplorerProps> = (props) => {
...
+        [
+            state.timeSeriesTwins,
+            dispatch,
+            logDebugConsole,
+            sendDataHistoryExplorerUserTelemetry
+        ]
+    );
+
+    const handleOnDownloadTableClick = useCallback(() => {
+        downloadText(
```
msnyder-msft, line 384: nit: this helper is downloading the file as a CSV  but you are serializing the data to JSON. Not sure of the implications there. Just wanted to call it out.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
```diff
@@ -39,24 +80,371 @@ const DataHistoryExplorer: React.FC<IDataHistoryExplorerProps> = (props) => {
...
+            JSON.stringify(
+                tableData.map((d) => {
+                    const { property, ...rest } = d;
+                    delete rest.seriesId;
```
msnyder-msft, line 388: how robust is this? Should we instead map over the properties we know we want? I'm just wondering if it will break in the future if we add another property to the object.

msnyder-msft (approved):
THIS IS AWESOME!!! Really fantastic work @iremgokce. I couldn't find much to say other than great job. I love how this refactor turned out and I think it's in a super solid spot. Thanks for taking the time to do this.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
```diff
@@ -39,24 +80,371 @@ const DataHistoryExplorer: React.FC<IDataHistoryExplorerProps> = (props) => {
...
+    }, [data]);
+
+    const viewerData: TimerSeriesViewerData = useMemo(
+        () => ({ chart: chartData, table: tableData }),
+        [tableData, chartData]
+    );
+
+    //callbacks
+    const updateXMinAndMax = useCallback(() => {
+        const nowInMillis = Date.now();
```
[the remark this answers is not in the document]
iremgokce: yeah I will do it in a separate PR.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
iremgokce: yeah, there should be no sensitive information related to series, only their "count" and if they are set to "casted to numeric". I will add the id for the currently opened modal to tie all the user interactions in a later PR, it is in my list to do as the final wrap up PR.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
iremgokce: good catch, I think the mime type of csv is required to have the same keys across all items, and since we didn't have any different keys before it was all good. but to be more accurate I changed it to application/json and changed the method name to downloadJSON since we all use .json format to download files across the cardboard now.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
iremgokce: yeah just in case if we extend table columns in the future, I changed it to assign what we want

src/Components/DataHistoryExplorer/Internal/TimeSeriesViewer/TimeSeriesViewer.types.ts
```diff
@@ -1,14 +1,38 @@
...
+    chart: Array<ADXTimeSeries>;
+    table: Array<TimeSeriesTableRow>;
+}
 export interface ITimeSeriesViewerProps {
-    timeSeriesTwinList: Array<IDataHistoryTimeSeriesTwin>;
-    onMissingSeriesData?: (seriesIds: Array<string>) => void;
+    timeSeriesTwins: Array<IDataHistoryTimeSeriesTwin>;
+    data: TimerSeriesViewerData;
+    isLoading: boolean;
+    viewerModeProps: IViewerModeProps;
```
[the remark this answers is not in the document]
iremgokce: IViewerModeProps is also a field for ITimeSeriesCommandBarProps which is props of the child here. So since it is not fully overlapping in terms of types (ITimeSeriesCommandBarOptions is not same as IDataHistoryExplorerChartOptions), I am going to leave it as is here.

montgomp (approved):
This is some really great work.

PR #859, opened 2023-02-07
OAT: Allow uploads of non-json files & fixing co-type handling of relationships

msnyder-msft:
### Summary of changes 🔍 
Today we stop the entire import process if there is a file in the uploaded set that is not a JSON file. Now we only alert the user if there is not a single file we can process.
image

Co-type relationships were previously not supported. But now we do! So this will work:
```JSON
{
    "@context": "dtmi:dtdl:context;2",
    "@id": "dtmi:model0;1",
    "@type": "Interface",
    "contents":
    [
        {
            "@type":
            [
                "Property",
                "AnotherType"
            ],
            "name": "nedfdfdfddfw_property_0dfddfd",
            "schema": "boolean"
        },
        {
            "@type": ["Relationship","Type2"],
            "name": "Relationship_0",
            "target": "dtmi:model2;1"
        },
        {
            "@type": "Relationship",
            "name": "Relationship_1",
            "target": "dtmi:model1;1"
        }
    ],
    "displayName": "Model0"
}
```

### Testing 🧪
- Try to import some file that is not JSON
- Try to import a folder containing non-JSON files but at least 1 valid DTDL JSON
- Try to import a folder with only non-JSON file(s)

### Checklist ✔️
- [ ] Added relevant labels


src/Models/Services/OatPublicUtils.ts
```diff
@@ -14,7 +15,7 @@ import {
 import { convertModelToDtdl, parseModelId, safeJsonParse } from './OatUtils';
 import { deepCopy, getDebugLogger, isDefined } from './Utils';
 
-const debugLogging = false;
+const debugLogging = true;
```
montgomp, line 18: I think we want to keep this false

msnyder-msft: damn! we really ned to build the eslint rule for this.

src/Components/OATGraphViewer/OATGraphViewer.tsx
```diff
@@ -91,8 +91,12 @@ import { DTDLType } from '../../Models/Classes/DTDL';
 import { IReactFlowInstance } from '../../Pages/OATEditorPage/Internal/Classes/OatTypes';
 import { ElementEdge } from './Internal/Classes/ElementEdge';
 import { ElementNode } from './Internal/Classes/ElementNode';
+import {
+    isDTDLComponentReference,
+    isDTDLRelationshipReference
+} from '../../Models/Services/DtdlUtils';
 
-const debugLogging = false;
+const debugLogging = true;
```
montgomp, line 99: I think we want this false

msnyder-msft: done

montgomp (approved):
New changes look good

PR #860, opened 2023-02-07
Added CORS enabled for ADT adapter

pasanchMSFT:
### Summary of changes 🔍 
To allow CORS enabled requests, I've added a variable in ADT Adapter that decides if proxy header and URL will be appended or not to all requests. Cors enabled variable is set to true by default in the constructor. It also checks for the current `window.origin` and `localStorage` to determine if it will stay true or not. Changes here allow for Storybook testing with CORS enabled, left a comment to allow that in `middleware.js` file.

image

### Testing 🧪
Local testing requires setting up permissions on dev cluster. Contact me if you want to test this yourself.

### Checklist ✔️
- [x] Added relevant labels


msnyder-msft (approved):
Looks solid to me. Let's just make sure we have easy ways to turn it off or that it's off by default till we are ready to go broad with it.

src/Models/Services/Utils.ts
```diff
@@ -999,3 +1000,18 @@ export function capitalizeFirstLetter(str: string) {
         return str;
     }
 }
+
+/**
+ * Validate if URL is Explorer for CORS enabling
+ */
+export function validateExplorerOrigin(origin: string) {
+    if (
```
montgomp, line 1008: could we return directly here `return [ADT_ALLOW_LISTED.DEV, ADT_ALLOW_LISTED_URLS.PROD].includes(origin);`

montgomp (approved):
looks good

pasanchMSFT:
> Looks solid to me. Let's just make sure we have easy ways to turn it off or that it's off by default till we are ready to go broad with it.

Off by default in Storybook, just change from false -> true to enable.

On by default in Adapters, just change value in constructor or add proxy override in localStorage.

.storybook/middleware.js
```diff
@@ -9,6 +9,8 @@ const { createProxyMiddleware } = require('http-proxy-middleware');
 const retryNumber = 3;
 
 module.exports = function (app) {
+    // Set to true when you want to test CORS in Storybook
+    const isCorsEnabled = false;
```
iremgokce, line 13: as we discussed on your PublickRepoFork PR, I think isCorsEnabled might be confusing variable name since you are referring if this feature is enabled on the backend, but here you use in a way to basically do cors related stuff if **not** enabled, lets change the name to "shouldProxy" or something.

PR #868, opened 2023-02-14
OAT: Add auto layout on import flow

msnyder-msft:
### Summary of changes 🔍 
Added a dialog that shows a progress indicator for the import as well as a confirmation screen at the end. When the dialog is closed, the graph gets laid out.
image

image

Graph Auto layout on import

### Testing 🧪
- Import a single file --> see the dialog flash quickly to complete
- Import a large ontology --> see the dialog stays up for a while and then gives a confirmation


src/Components/ConfirmDialog/ConfirmDialog.styles.ts
```diff
@@ -0,0 +1,27 @@
...
+    IConfirmDialogStyles
+} from './ConfirmDialog.types';
+import { CardboardClassNamePrefix } from '../../Models/Constants/Constants';
+
+const classPrefix = `${CardboardClassNamePrefix}-confirmdialog`;
+const classNames = {
+    root: `${classPrefix}-root`
+};
+
+// export const CONFIRMDIALOG_CLASS_NAMES = classNames;
```
pasanchMSFT, line 12: I remember this from another PR as well, what was the reason for keeping these? I thought we had considered not leaving them commented out but using the actual different export name.

src/Components/OATImportProgressDialog/OATImportProgressDialog.styles.ts
```diff
@@ -0,0 +1,20 @@
...
+    IOATImportProgressDialogStyles
+} from './OATImportProgressDialog.types';
+import { CardboardClassNamePrefix } from '../../Models/Constants/Constants';
+
+const classPrefix = `${CardboardClassNamePrefix}-oatimportprogressdialog`;
+const classNames = {
+    root: `${classPrefix}-root`
+};
+
+// export const OATIMPORTPROGRESSDIALOG_CLASS_NAMES = classNames;
```
pasanchMSFT, line 12: same here as in the other file

pasanchMSFT (approved):
LGTM, just to small comments

src/Components/ConfirmDialog/ConfirmDialog.styles.ts
msnyder-msft: the reason I don't change the name itself from 'classNames' is because then everywhere you use that internally to this file it's way longer. I could remove from the template though if you feel like it's just bloat.

msnyder-msft: Removed from here though.

PR #869, opened 2023-02-14
OAT: Add max model count limit

msnyder-msft:
### Summary of changes 🔍 
Adding a limit on the number of models that can be added. Set to 70 for now.
image

### Testing 🧪
Added test stories - /story/components-oat-oatimportprogressdialog--loading
- Import folder with more than 70 models
- Add 69 models and then try to "Add model" one more time from the header


pasanchMSFT (approved):
Code looks good

PR #870, opened 2023-02-14
OAT: Enable editing of Maps (previously missed)

msnyder-msft:
### Summary of changes 🔍 
Users can now edit the key and values of a Map property type.
This was missed in earlier iterations so wiring up the callbacks to store the data.

image

### Testing 🧪
Use this story - /story/components-oat-oatpropertyeditor-propertylist--complex-model
- Add a map to a model with primitive types
- Rename `key` of the Map
- Rename `value` property of primitive Map value
- Rename `value` property of complex Map property
- Change type of `value` field
- Change name and type of children of complex `value` field


src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyListItemChildHost/Internal/PropertyListItemMapChild/PropertyListItemMapChild.tsx
```diff
@@ -26,6 +39,20 @@ const PropertyListItemMapChild: React.FC<IPropertyListItemMapChildProps> = (
     // hooks
 
     // callbacks
+    const onUpdateKeyInternal = useCallback(
+        (key: DtdlMapKey) => {
+            //
```
pasanchMSFT, line 44: comment?

src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyListItemChildHost/Internal/PropertyListItemMapChild/PropertyListItemMapChild.tsx
```diff
@@ -26,6 +39,20 @@ const PropertyListItemMapChild: React.FC<IPropertyListItemMapChildProps> = (
...
+            onUpdateKey(key);
+        },
+        [onUpdateKey]
+    );
+    const onUpdateValueInternal = useCallback(
+        (value: DtdlMapValue) => {
+            //
```
pasanchMSFT, line 51: same here

src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyListItemChildHost/Internal/PropertyListItemMapChild/PropertyListItemMapChild.tsx
```diff
@@ -45,8 +72,12 @@ const PropertyListItemMapChild: React.FC<IPropertyListItemMapChildProps> = (
                 onCopy={undefined}
                 onRemove={undefined}
                 onReorderItem={undefined}
-                onUpdateName={undefined}
-                onUpdateSchema={undefined}
+                onUpdateName={(args) => {
+                    const keyCopy = deepCopy(item.mapKey);
```
pasanchMSFT, line 76: You use a very similar function in the following onUpdateName and onUpdateSchema, could we pull this out into a shared function?

src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyListItemChildHost/PropertyListItemChildHost.tsx
```diff
@@ -129,6 +129,20 @@ const PropertyListItemChildHost: React.FC<IPropertyListItemChildHostProps> = (
...
                     item={propertyItem.schema}
                     level={level}
+                    onUpdateKey={(key) => {
+                        // update the schema for the item
+                        const schemaCopy = deepCopy(propertyItem.schema);
+                        schemaCopy.mapKey = key;
+                        // send updated schema to parent
+                        onUpdateSchema(schemaCopy);
+                    }}
+                    onUpdateValue={(value) => {
```
pasanchMSFT, line 139: This function is also used in `PropertyListItemMapChild.tsx` it could even be part of a utils file

src/Models/Constants/dtdlInterfaces.ts
```diff
@@ -122,8 +122,8 @@ export interface DtdlArray {
 export interface DtdlMapKey {
     /** The "programming" name of the map's value. The name may only contain the characters a-z, A-Z, 0-9, and underscore, and must match this regular expression `^[a-zA-Z](?:[a-zA-Z0-9_]*[a-zA-Z0-9])?$.` */
     name: string;
-    /** The data type of the map's key */
-    schema: DTDLSchema;
+    /** The data type of the map's key. Always `string` */
+    schema: 'string';
```
pasanchMSFT, line 126: Why do we need 'string' literally defined here?

pasanchMSFT (commented):
Couple of comments here

src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyListItemChildHost/Internal/PropertyListItemMapChild/PropertyListItemMapChild.tsx
msnyder-msft: one updates the name property and one updates schema so I think this is about as consolidated as they can be.

src/Models/Constants/dtdlInterfaces.ts
msnyder-msft: becuase it has this property called `Schema` but the value is always defined as 'string'. I keep it here cause we do type checking on the `schema` field a lot. And everything in DTDL has a schema, it's just that sometimes, that field is a fixed constant. It's the same way with the @type field.

src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyListItemChildHost/PropertyListItemChildHost.tsx
msnyder-msft: in this case I think it is fine to be in both spots. It follows the same pattern as the rest of these callbacks for the other child properties so I'd rather not obfuscate it by moving to a utility.

pasanchMSFT (approved):

PR #879, opened 2023-02-17
Added CORS condition to Blob adapter

pasanchMSFT:
### Summary of changes 🔍 
Blob adapter required some changes to be able to make CORS enabled requests. This PR addresses that need and also modifies the existing CORS conditionals to fix forcing CORS with a local storage flag.

### Testing 🧪
Setting `useBlobProxy` enabled
1. Modify either `BlobAdapter` or `ADT3DScenesAdapter` to set `useBlobProxy` to true by default.
2. Run the project locally to be able to make requests.
3. Load containers from storage accounts that have CORS enabled.

Setting `forceProxy` flag
1. Set `useBlobProxy` to false.
2. Run this command in your console `localStorage.setItem('cardboard.feature.forceCORS', 'true')`.
3. Run the project locally to be able to make requests.
4. Load containers from storage accounts that have CORS enabled.

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Adapters/ADT3DSceneAdapter.ts
+++ b/src/Adapters/ADT3DSceneAdapter.ts
@@ -46,7 +46,7 @@ export default class ADT3DSceneAdapter {
         uniqueObjectId?: string,
         adtProxyServerPath = '/proxy/adt',
         blobProxyServerPath = '/proxy/blob',
-        useProxy = true,
+        useAdtProxy = true,
         useBlobProxy = true
     ) {
         this.adtHostUrl = adtHostUrl;
@@ -67,8 +67,9 @@ export default class ADT3DSceneAdapter {
          * Check if class has been initialized with CORS enabled or if origin matches dev or prod explorer urls,
          * override if CORS is forced by feature flag
          *  */
-        this.useProxy =
-            (useProxy || !validateExplorerOrigin(window.origin)) && !forceCORS;
+        this.useAdtProxy =
+            (useAdtProxy || !validateExplorerOrigin(window.origin)) &&
+            !forceCORS;
 
         this.useBlobProxy =
             (useBlobProxy || !validateExplorerOrigin(window.origin)) &&
--- a/src/Adapters/ADT3DSceneAdapter.ts
+++ b/src/Adapters/ADT3DSceneAdapter.ts
@@ -33,6 +33,10 @@ import {
 import ADTInstanceTimeSeriesConnectionData from '../Models/Classes/AdapterDataClasses/ADTInstanceTimeSeriesConnectionData';
 import ADTDataHistoryAdapter from './ADTDataHistoryAdapter';
 
+const forceCORS =
+    localStorage.getItem(LOCAL_STORAGE_KEYS.FeatureFlags.Proxy.forceCORS) ===
+    'true';
+
 export default class ADT3DSceneAdapter {
     constructor(
         authService: IAuthService,
@@ -64,16 +68,11 @@ export default class ADT3DSceneAdapter {
          * override if CORS is forced by feature flag
          *  */
         this.useProxy =
-            (useProxy || !validateExplorerOrigin(window.origin)) &&
-            localStorage.getItem(
-                LOCAL_STORAGE_KEYS.FeatureFlags.Proxy.forceCORS
-            ) !== 'true';
+            (useProxy || !validateExplorerOrigin(window.origin)) && !forceCORS;
 
         this.useBlobProxy =
             (useBlobProxy || !validateExplorerOrigin(window.origin)) &&
-            localStorage.getItem(
-                LOCAL_STORAGE_KEYS.FeatureFlags.Proxy.forceCORS
-            ) !== 'true';
+            !forceCORS;
 
         if (blobContainerUrl) {
             try {
--- a/src/Adapters/ADT3DSceneAdapter.ts
+++ b/src/Adapters/ADT3DSceneAdapter.ts
@@ -65,11 +65,9 @@ export default class ADT3DSceneAdapter {
          *  */
         this.useProxy =
             (useProxy || !validateExplorerOrigin(window.origin)) &&
-            !(
-                localStorage.getItem(
-                    LOCAL_STORAGE_KEYS.FeatureFlags.Proxy.forceCORS
-                ) === 'true'
-            );
+            localStorage.getItem(
+                LOCAL_STORAGE_KEYS.FeatureFlags.Proxy.forceCORS
+            ) !== 'true';
 
         this.useBlobProxy =
             (useBlobProxy || !validateExplorerOrigin(window.origin)) &&
--- a/src/Adapters/ADT3DSceneAdapter.ts
+++ b/src/Adapters/ADT3DSceneAdapter.ts
@@ -60,14 +60,15 @@ export default class ADT3DSceneAdapter {
         );
         /**
          * Check if class has been initialized with CORS enabled or if origin matches dev or prod explorer urls,
-         * override if proxy is forced by feature flag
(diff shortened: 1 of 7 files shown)


src/Adapters/ADTDataHistoryAdapter.ts
```diff
@@ -57,14 +57,15 @@ export default class ADTDataHistoryAdapter implements IADTDataHistoryAdapter {
...
-            !validateExplorerOrigin(window.origin) ||
-            localStorage.getItem(
-                LOCAL_STORAGE_KEYS.FeatureFlags.Proxy.forceProxy
-            ) === 'true';
+            (useProxy || !validateExplorerOrigin(window.origin)) &&
+            !(
+                localStorage.getItem(
+                    LOCAL_STORAGE_KEYS.FeatureFlags.Proxy.forceCORS
+                ) === 'true'
+            );
```
montgomp, line 68: This would read better as `localStorage.getItem( LOCAL_STORAGE_KEYS.FeatureFlags.Proxy.forceCORS ) !== 'true'` removing the wrapping parenthesis and negation.

src/Adapters/ADT3DSceneAdapter.ts
```diff
@@ -60,14 +61,23 @@ export default class ADT3DSceneAdapter {
...
+            (useProxy || !validateExplorerOrigin(window.origin)) &&
+            !(
+                localStorage.getItem(
+                    LOCAL_STORAGE_KEYS.FeatureFlags.Proxy.forceCORS
+                ) === 'true'
+            );
+
+        this.useBlobProxy =
+            (useBlobProxy || !validateExplorerOrigin(window.origin)) &&
+            !(
```
montgomp, line 76: This would read better moving the negation to `!==`

msnyder-msft, line 74: can we move the local storage read up to the module scope so we don't need to do it on every call? it's gonna be the same value for the life of the app hopefully so we can save the perf and the ease of reading.

src/Adapters/BlobAdapter.ts
```diff
@@ -36,16 +40,54 @@ export default class BlobAdapter implements IBlobAdapter {
...
+
+    generateBlobUrl(path: string) {
+        if (this.useBlobProxy) {
+            return `${this.blobProxyServerPath}${path}`;
+        } else {
+            return `${this.getBlobContainerURL()}${path}`;
+        }
+    }
+
+    generateBlobHeaders(headers: AxiosObjParam = {}) {
```
msnyder-msft, line 82: more dupe code makes me saaaaaaad. We should make this a base class.

src/Adapters/BlobAdapter.ts
```diff
@@ -405,15 +445,18 @@ export default class BlobAdapter implements IBlobAdapter {
...
             {
                 method: 'get',
-                url: `${this.blobProxyServerPath}`,
-                headers: {
-                    'x-ms-version': '2021-06-08',
-                    'x-blob-host': this.storageAccountHostName
-                },
+                // URL here requires storage account with container name removed
+                url: getStorageAccountUrlFromContainerUrl(
+                    this.generateBlobUrl('')
```
msnyder-msft, line 457: intentionally passing ''?

msnyder-msft: Seems like it makes sense, just double checking.

src/Adapters/ADT3DSceneAdapter.ts
pasanchMSFT: what do you mean module scope?

msnyder-msft: top of da file.

src/Adapters/BlobAdapter.ts
pasanchMSFT: BlobHeaders != ADTHeaders even if we made this a base class this method would be "duped"

src/Adapters/ADT3DSceneAdapter.ts
pasanchMSFT: done

montgomp (approved):

msnyder-msft (approved):

src/Adapters/ADT3DSceneAdapter.ts
```diff
@@ -42,7 +46,8 @@ export default class ADT3DSceneAdapter {
         uniqueObjectId?: string,
         adtProxyServerPath = '/proxy/adt',
         blobProxyServerPath = '/proxy/blob',
-        useProxy = true
+        useProxy = true,
```
iremgokce, line 49: now, can we make it explicity useAdtProxy to make it more clear as we did for adtProxyServerPath and blobProxyServerPath?
```suggestion
        useAdtProxy = true,
```

src/Adapters/BlobAdapter.ts
```diff
@@ -206,17 +245,18 @@ export default class BlobAdapter implements IBlobAdapter {
...
                 const filesData = await axios({
                     method: 'GET',
-                    url: `${this.blobProxyServerPath}/${this.containerName}`,
-                    headers: {
-                        authorization: 'Bearer ' + token,
-                        'Content-Type': 'application/json',
-                        'x-ms-version': '2017-11-09',
-                        'x-blob-host': this.storageAccountHostName
-                    },
+                    url: this.generateBlobUrl(''),
```
iremgokce, line 258: Seeing this with empty string makes me think that we should change the name of the method to something like generateBlobUrlWithPath or something.
```suggestion
                    url: this.generateBlobUrlWithPath(''),
```

src/Adapters/BlobAdapter.ts
```diff
@@ -36,16 +44,50 @@ export default class BlobAdapter implements IBlobAdapter {
...
+            !forceCORS;
+    }
+
+    getBlobContainerURL() {
+        return this.storageAccountHostName && this.containerName
+            ? `https://${this.storageAccountHostName}/${this.containerName}`
+            : '';
+    }
+
+    generateBlobUrl(path: string) {
```
iremgokce, line 74: nit: can we change the name to reflect the argument?
```suggestion
    generateBlobUrlWithPath(path: string) {
```

src/Models/Constants/Constants.ts
```diff
@@ -93,7 +93,7 @@ export const LOCAL_STORAGE_KEYS = {
             showExplorer: 'cardboard.feature.dataHistoryExplorer' // enables data history explorer feature if supported
         },
         Proxy: {
-            forceProxy: 'cardboard.feature.forceProxy' // force proxy to run instead of CORS
+            forceCORS: 'cardboard.feature.forceCORS' // force CORS to run instead of proxy
```
iremgokce, line 96: is there a reason why we changed the key name in local storage? Also, does this require small migration? But since it is a feature flag I guess it should be good.

iremgokce (commented):
I added some small comments. I also noticed there are some GitHub build warning messages under the files with "Unchanged files with check annotations" section. Can you also check what those are? - nevermind, it seems these are in unrelated fiels to this PR/

src/Models/Constants/Constants.ts
pasanchMSFT: Yes, I changed the logic to be default proxy and force CORS if needed. I thought it was better to change the key in local storage to reflect that change in logic.

src/Adapters/BlobAdapter.ts
pasanchMSFT: I like the naming for generateBlobUrl better. I think the naming in the argument (path: string) is enough to give people understanding of the intention of the function.

PR #882, opened 2023-02-21
fix: OAT: add custom validation for renaming properties to avoid duplicate property names

msnyder-msft:
### Summary of changes 🔍 
Refactoring validation logic out to a common place utility such that we can call it with the current state and selection and give it an updated item to insert into the ontology temporarily so we can validate and then decide whether to commit the change or not.

In particular, we are using this logic to validate when the name of a property is changed so that it can avoid introducing duplicate property names. Validation is now run `onBlur` of the name field.
image

### Testing 🧪
- Add two properties to a Relationship and try to change the name to be the same 
- Add two properties to a Model and try to change the name to be the same 

### Checklist ✔️
- [x] Added relevant labels


src/Components/OATPropertyEditor/Internal/EditorJsonTab/Internal/JSONEditor/JSONEditor.tsx
```diff
@@ -155,85 +133,44 @@ const JSONEditor: React.FC<IJSONEditorProps> = (props) => {
...
-        const parsingError = await parseModels(models);
+        const validationResult = await validateItemChangeBeforeSaving({
+            content: content,
+            originalItem: selectedItem,
+            existingModels: oatPageState.currentOntologyModels,
+            selectedModelId: oatPageState.selection.modelId
+        });
 
-        if (parsingError) {
+        if (validationResult.isValid === true) {
```
iremgokce, line 146: is there a reason why it is not only `if (validationResult.isValid)`

src/Components/OATPropertyEditor/Internal/PropertyList/PropertyList.tsx
```diff
@@ -170,9 +183,25 @@ const PropertyList: React.FC<IPropertyListProps> = (props) => {
...
+                };
+
+                // validate changes
+                validateOntology({
+                    existingModels: oatPageState.currentOntologyModels,
+                    selectedModelId: oatPageState.selection?.modelId,
+                    originalItem: selectedItem,
+                    updatedItem: updatedModel
+                }).then((validationResult) => {
+                    if (validationResult.isValid === true) {
```
iremgokce, line 198: same comment here, why not just `if (validationResult.isValid)`

src/Components/OATPropertyEditor/Internal/PropertyList/PropertyList.tsx
```diff
@@ -170,9 +183,25 @@ const PropertyList: React.FC<IPropertyListProps> = (props) => {
...
                     contents: updatedContents
+                };
+
+                // validate changes
+                validateOntology({
+                    existingModels: oatPageState.currentOntologyModels,
+                    selectedModelId: oatPageState.selection?.modelId,
+                    originalItem: selectedItem,
+                    updatedItem: updatedModel
+                }).then((validationResult) => {
```
iremgokce, line 197: should you also have `.catch` block for this promise to call `showErrorDialog(error);` in case validateOntology fails?

src/Components/OATPropertyEditor/Internal/PropertyList/PropertyList.tsx
```diff
@@ -182,9 +211,24 @@ const PropertyList: React.FC<IPropertyListProps> = (props) => {
...
                     ...selectedItem,
                     properties: updatedProperties
+                };
+                // validate changes
+                validateOntology({
+                    existingModels: oatPageState.currentOntologyModels,
+                    selectedModelId: oatPageState.selection?.modelId,
+                    originalItem: selectedItem,
+                    updatedItem: updatedReference
+                }).then((validationResult) => {
```
iremgokce, line 224: came here, should you also have `.catch` block for this promise to call `showErrorDialog(error);` in case validateOntology fails?

src/Resources/Locales/en.json
```diff
@@ -927,6 +927,11 @@
             "unhandledExceptionMessage": "Something unexpected happened. Please try again.",
             "versionDocumentationLink": "Learn more about DTDL version 3"
         },
+        "Errors": {
```
iremgokce, line 930: do we start with upper case in translation keys?

iremgokce (approved):
added some nit comments but did not run the code to test live.

src/Models/Services/DtdlUtils.ts
```diff
@@ -516,6 +521,95 @@ export const isValidDtmiPath = (path: string, isFinal: boolean): boolean => {
...
+    content: string;
+    originalItem: T;
+    selectedModelId: string;
+    existingModels: DtdlInterface[];
+}): Promise<IValidationResult> {
+    const isJsonStringValid = (value: string): T | null => {
+        try {
+            return JSON.parse(value);
+        } catch (e) {
+            return null;
```
montgomp, line 554: Should we return an object here with a message that the json didn't parse?

montgomp (approved):
One small question that I'll leave to you, but otherwise looks great.

src/Components/OATPropertyEditor/Internal/EditorJsonTab/Internal/JSONEditor/JSONEditor.tsx
msnyder-msft: Yeah. I don't know why but typescript only figures out the discriminated union if I do the explicit check. I feel like it should be smart enough but alas I haven't figured out an alternative here.

src/Components/OATPropertyEditor/Internal/PropertyList/PropertyList.tsx
msnyder-msft: same as above, the discriminated union barks without the explicit check.

src/Components/OATPropertyEditor/Internal/PropertyList/PropertyList.tsx
msnyder-msft: hmmm. fair enough. I can add that.

src/Resources/Locales/en.json
msnyder-msft: At least in OAT my pattern has been that each section block is caps and the actual key is lowercase.

src/Models/Services/DtdlUtils.ts
msnyder-msft: I actually stole this block from the vendors, but basically, they null check the result to know if it parsed correctly. I could make it fancier but that is what they do in a couple places to avoid bubbling up errors from the parsing operation.

src/Components/OATPropertyEditor/Internal/PropertyList/PropertyList.tsx
msnyder-msft: done

PR #884, opened 2023-02-21
fix: OAT: Handle targets that are arrays

msnyder-msft:
### Summary of changes 🔍 
DTDL does allow for the target field to be an array of length 1 or to be a string. Today we had always assumed it was a string, but now we will be resilient to receiving an array.

Before
image

After
image

### Testing 🧪
Sample model from the bug report - [Link](https://github.com/Azure/iot-plugandplay-models/blob/main/dtmi/com/socomec/socomecgateway-1.json)

- update the name of the targeted model, see that the relationship between the nodes is retained
- Import a model that has a target with an array instead of string


src/Components/OATGraphViewer/Internal/Utils.tsx
```diff
@@ -53,7 +53,9 @@ export const addTargetedRelationship = (
         '',
         sourceId,
         OAT_RELATIONSHIP_HANDLE_NAME,
-        relationship['target'],
+        Array.isArray(relationship['target'])
```
iremgokce, line 56: is there a reason why we dont use `relationship.target` as below? Is it due to type casting might be needed if we use?
```suggestion
        Array.isArray(relationship.target)
```

iremgokce (approved):
added a nit comment

src/Components/OATGraphViewer/Internal/Utils.tsx
msnyder-msft: I think this is just legacy formatting from when the vendors wrote it. I can update it.

msnyder-msft: done.

PR #885, opened 2023-02-22
fix: OAT: block adding self-referencing relationships for Extend or Components

msnyder-msft:
### Summary of changes 🔍 
Users should not be able to a self-referencing relationship of type Inheritance or Component.

Before
image
image

After
Nada, can't get that to happen

### Testing 🧪
- Try to draw a Component reference to itself
- Try to draw an Inheritance reference to itself

### Checklist ✔️
- [x] Added relevant labels


.eslintrc.json
```diff
@@ -3,18 +3,20 @@
...
         "prettier",
-        "prettier/@typescript-eslint"
+        "plugin:iot-cardboard-js/recommended"
     ],
     "rules": {
+        "iot-cardboard-js/debug-logging-disabled": "warn",
         "no-debugger": "warn",
-        "no-multiple-empty-lines": "error",
-        "react-hooks/exhaustive-deps": "off",
+        "no-multiple-empty-lines": "warn",
```
iremgokce, line 18: just to confirm, are we sure about turning those off?

src/Components/3DV/SceneViewWrapper.tsx
```diff
@@ -184,8 +184,8 @@ const SceneViewWrapper: React.FC<ISceneViewWrapperProps> = (props) => {
                 objectColorOptions={sceneThemeState.objectColorOptions}
                 objectStyle={sceneThemeState.objectStyle}
                 onCameraMove={addInProps?.onCameraMove ? cameraMove : undefined}
-                {...svp}
                 outlinedMeshitems={sceneViewState.outlinedMeshItems}
+                {...svp}
```
iremgokce, line 188: Didn't we already merged Paul's changes today?

.eslintrc.json
```diff
@@ -3,18 +3,20 @@
...
-        "react-hooks"
+        "react-hooks",
+        "iot-cardboard-js"
     ],
     "extends": [
         "eslint:recommended",
         "plugin:@typescript-eslint/recommended",
         "prettier",
-        "prettier/@typescript-eslint"
+        "plugin:iot-cardboard-js/recommended"
```
iremgokce, line 13: ohh do we have a plugin now? Can you walk us through about those eslint related changes?

.eslintrc.prod.json
```diff
@@ -0,0 +1,15 @@
+{
+    "extends": "./.eslintrc.json",
+    "rules": {
+        "iot-cardboard-js/debug-logging-disabled": "error",
+        "no-debugger": "error",
+        "no-multiple-empty-lines": "error",
+        "react-hooks/exhaustive-deps": "off",
```
iremgokce, line 7: so when it is prod, these rules will be overridden with those values? but how would it help us in dev as we send PRs?

.eslintrc.json
msnyder-msft: I wrote it on Friday. I can certainly go over it for folks that are interested.

.eslintrc.json
msnyder-msft: I'm not actually turning anything off. I'm just setting them to warnings so that it won't fail your local compilation. WHat we see today is people adding the `eslint-ignore` statements just to be able to develop and then those can accidentally get checked in. This allows us to get away with it locally (with yellow squiggles) and then it will throw errors on the PR if you forget to remove them.

.eslintrc.prod.json
msnyder-msft: "prod" here is just the CI pipeline. This is the file that is used anytime you're not local.

pasanchMSFT (approved):

PR #888, opened 2023-02-22
ESLint: Adding recommended rule set for hooks

msnyder-msft:
### Summary of changes 🔍 
Adding the recommended rule set for [eslint-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks). We already have the package and use it for the dependency lists, but adding the recommended defaults as well now.

Most of the violations came in the form of calling hooks conditionally. If I made changes in a file, I also tried to fix any dependency lists for `useCallback` while I was there to use the recommended entries. I did not fix `useMemo` or `useEffect` as those could change behavior. 

I only disabled the rules in the context files since we use a pattern for checking that the context does not exist before wrapping it in another context. The one exception to this rule was the breadcrumbs where the context is undefined if the build mode is not set so we had to keep that on conditionally as well.

### Testing 🧪
- Clicked around in 3D scenes and things seem to work working well
- Clicked around in OAT and things seem fine as well


src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
```diff
@@ -496,6 +497,7 @@ const DataHistoryExplorer: React.FC<IDataHistoryExplorerProps> = (props) => {
         ) {
             updateConnectionAdapterData.callAdapter();
         }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
pasanchMSFT, line 500: do we need to keep track of these to fix them later? or could you add a comment on the ones that we won't fix.

src/Components/OATPropertyEditor/Internal/TemplateColumn.tsx
```diff
@@ -29,9 +30,10 @@ export const TemplateColumn: React.FC<TemplateColumnProps> = (props) => {
     const { oatPageDispatch } = useOatPageContext();
 
     // styles
-    const propertyInspectorStyles = getPropertyInspectorStyles();
-    const textFieldStyles = getPropertyEditorTextFieldStyles();
-    const templateColumnStyles = getTemplateColumnStyles();
+    const theme = useTheme();
+    const propertyInspectorStyles = getPropertyInspectorStyles(theme);
```
pasanchMSFT, line 34: Why is theme here sent without the `{ }` like the other two functions?

src/Components/OATPropertyEditor/OATPropertyEditor.styles.ts
```diff
@@ -226,8 +224,8 @@ export const getPropertyEditorTextFieldStyles = () => {
 
 /* Template Column */
 
-export const getTemplateColumnStyles = () => {
-    const theme = useTheme();
+export const getTemplateColumnStyles = (args: { theme: Theme }) => {
```
pasanchMSFT, line 227: shouldn't the type be ExtendedTheme?

src/Components/SceneBreadcrumb/SceneBreadcrumbFactory.tsx
```diff
@@ -91,6 +91,7 @@ const SceneBreadcrumbFactory: React.FC<ISceneBreadcrumbFactoryProps> = ({
             setBehaviorTwinAliasFormInfo,
             setVisualRuleFormMode,
             visualRuleFormMode
+            // eslint-disable-next-line react-hooks/rules-of-hooks
```
pasanchMSFT, line 94: How will we keep track of these to fix?

pasanchMSFT (commented):
In general, looks good, but I've got a couple of questions about tracking the eslint ignore comments.

src/Components/DataHistoryExplorer/DataHistoryExplorer.tsx
msnyder-msft: I'm not planning to fix any of these actually. This one intentionally doesn't have any dependencies becaues it's supposed to only fire on mount.

src/Components/OATPropertyEditor/OATPropertyEditor.styles.ts
msnyder-msft: this one didn't need our extended theme so I left it as just a basic theme. I could change it, but it doesn't need it.

src/Components/SceneBreadcrumb/SceneBreadcrumbFactory.tsx
msnyder-msft: ehh. we don't really need to? But if you wnat to find them, they all have the same disable comment so you can just search those out.

PR #891, opened 2023-02-23
[Data History Explorer]: fix mock data

iremgokce:
### Summary of changes 🔍 
Added check for adapter's mockData type to be returned from getTimeSeriesData method. Ideally we should specify what is the mockData is for.

### Testing 🧪
You can test it in DataHistoryExplorer LargeData mock story since it is using mockData assigned when the new MockAdapter is initialized vs in any story which uses data history where mockData is initialized as a different type for the MockAdapter like ADT3DScenePage mock stories with mock scenes config:
image
image
image

### Checklist ✔️
- [x] Added relevant labels

### DIFF
--- a/src/Adapters/MockAdapter.ts
+++ b/src/Adapters/MockAdapter.ts
@@ -86,7 +86,10 @@ import ViewerConfigUtility from '../Models/Classes/ViewerConfigUtility';
 import ADTInstanceTimeSeriesConnectionData from '../Models/Classes/AdapterDataClasses/ADTInstanceTimeSeriesConnectionData';
 import { handleMigrations } from './BlobAdapterUtility';
 import ADXTimeSeriesData from '../Models/Classes/AdapterDataClasses/ADXTimeSeriesData';
-import { getMockTimeSeriesDataArrayInLocalTime } from '../Models/SharedUtils/DataHistoryUtils';
+import {
+    getMockTimeSeriesDataArrayInLocalTime,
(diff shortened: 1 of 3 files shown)


pasanchMSFT (approved):

montgomp (approved):
LGTM

src/Adapters/MockAdapter.ts
```diff
@@ -1049,7 +1052,17 @@ export default class MockAdapter
...
-        let mockTimeSeriesData: Array<Array<TimeSeriesData>> = this.mockData;
+        let mockTimeSeriesData: Array<Array<TimeSeriesData>>;
+        if (
+            this.mockData &&
+            Array.isArray(this.mockData) &&
+            this.mockData.every(
+                (arr) => Array.isArray(arr) && arr.every(isTimeSeriesData)
+            )
+        ) {
+            mockTimeSeriesData = this.mockData;
```
msnyder-msft, line 1063: why not just use the mockData object in the mocked out adapter call? Do we need multiple sets of mocked data in the same test?

iremgokce: mockData is a any type of value that you can set for MockAdapter. As you can see in the screenshot above, for ADT3DScenePage the mockData is set to a mock config object. So it is hard to structure the mockData type now, e.g. the mock time series is not a part of the mockData object that we send to the adapter. Please check out my screenshots above to see the issue.

PR #895, opened 2023-03-01
OAT: Give maps custom icon and remove menu

msnyder-msft:
### Summary of changes 🔍 
Fixing 2 requests from Private Preview to clarify the Maps property experience
1. Give the Map Key field a different icon since it's always a string and to differentiate it from other property items.
2. Remove the overflow menu for the Map Key field since it can't do any of those things.

Also bumping the reference limit up from 200 to 250 as a point of feedback to support some out of the box ontologies.

Before
image

After
image

### Testing 🧪
Check out this story to see the new experience `components-oat-oatpropertyeditor-propertylist--complex-model`

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -32,7 +32,7 @@
         "@typescript-eslint/no-var-requires": 0,
         "@typescript-eslint/no-explicit-any": 0,
         "@typescript-eslint/no-unused-vars": [
-            "error",
+            "warn",
             {
                 "argsIgnorePattern": "^_"
             }
--- a/src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyIcon/PropertyIcon.tsx
+++ b/src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyIcon/PropertyIcon.tsx
@@ -18,7 +18,7 @@ const getClassNames = classNamesFunction<
 >();
 
 const PropertyIcon: React.FC<IPropertyIconProps> = (props) => {
-    const { schema, styles } = props;
+    const { overrideIcon, schema, styles } = props;
 
     // contexts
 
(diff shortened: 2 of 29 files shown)


montgomp (approved):
Looks good

.eslintrc.json
```diff
@@ -32,7 +32,7 @@
         "@typescript-eslint/no-var-requires": 0,
         "@typescript-eslint/no-explicit-any": 0,
         "@typescript-eslint/no-unused-vars": [
-            "error",
+            "warn",
```
pasanchMSFT, line 35: Did you intend to check-in this change?

src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyIcon/PropertyIcon.tsx
```diff
@@ -38,6 +38,13 @@ const PropertyIcon: React.FC<IPropertyIconProps> = (props) => {
 
     // data
     const { iconSource, iconTitle, iconType } = useMemo(() => {
+        if (overrideIcon) {
+            return {
+                iconSource: overrideIcon.name,
+                iconTitle: overrideIcon.title,
+                iconType: 'Fluent'
```
pasanchMSFT, line 45: why not overrideIconType?

pasanchMSFT (approved):
Small nits, looks good overall.

src/Components/OATPropertyEditor/Internal/PropertyList/Internal/PropertyListItem/Internal/PropertyIcon/PropertyIcon.tsx
msnyder-msft: cause today it will always be an icon by name

.eslintrc.json
msnyder-msft: ype. Local we don't need to fail when you have an extra import

PR #899, opened 2023-03-09
[Unified MSAL]: Updated MSAL service and exported for consumption

iremgokce:
### Summary of changes 🔍 
I used the auth service that we use in public repo fork & flights app. On top of it, made changes in the MSALAuthService class and enums/interfaces/parameters and exported it to be consumed externally (e.g. in public repo fork or flights apps). @darsney I left some comments where I need you review specifically and confirm the changes.

I followed some patterns from the example typescript file of msal-browser package: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/samples/msal-browser-samples/TypescriptTestApp2.0/src/AuthModule.ts 

### Testing 🧪
Now I changed the secrets structure, please update your secrets.user as appropriate in the placeholder file. Now the MSALAuthService expects a login url and client id along with a target tenant id - if exists. Update your secrets and test it with ADT3DScenePage local story. You do not need to clear your cache or something. It should just work if you are already logged in.

### Checklist ✔️
- [x] Added relevant labels


src/Models/Services/MsalAuthService.ts
```diff
@@ -1,81 +1,189 @@
...
+                    } else {
+                        let activeAccount;
+                        if (this.tenantId === DEFAULT_TENANT_ID) {
+                            if (
+                                this.targetTenantId &&
+                                this.targetTenantId !== DEFAULT_TENANT_ID
+                            ) {
+                                activeAccount = allAccounts.find(
+                                    (a: any) =>
+                                        a.tenantId === this.targetTenantId
```
iremgokce, line 144: I added target Id to capture intended tenant id passed in at the time of construction (e.g. if there is tenant id from url or local storage etc.)

src/Models/Services/MsalAuthService.ts
```diff
@@ -1,81 +1,189 @@
...
+                            } else {
+                                activeAccount = allAccounts[0];
+                            }
+                        } else {
+                            activeAccount = allAccounts.find(
+                                (a: any) => a.tenantId === this.tenantId
+                            );
+                        }
+                        this.setActiveAccount(activeAccount);
+                        if (continuation) {
```
iremgokce, line 155: I added this check for continuation if the ssoSilent is called externally and tenant id is captured by the consumer app in a specific way. You can check the usage example in my public repo fork PR.

src/Models/Services/MsalAuthService.ts
```diff
@@ -1,81 +1,189 @@
...
-        );
+        this.MSALObj = new PublicClientApplication(this.MSALConfig);
+
+        this.loginPopupRequest = {
+            scopes: []
+        };
+        this.loginRedirectRequest = {
+            ...this.loginPopupRequest,
+            redirectStartPage: window.location.href,
+            prompt: 'select_account'
```
iremgokce, line 69: This one was Matt's last change

src/Adapters/ADTAdapter.ts
```diff
@@ -100,16 +98,12 @@ export default class ADTAdapter implements IADTAdapter {
     constructor(
         adtHostUrl: string,
         authService: IAuthService,
-        tenantId?: string,
```
msnyder-msft, line 103: @darsney is this gonna cause a lot of pain for consumers who already provide these today? Should we send a comm to Paul about this so they aren't surprised if they pull latest versions in?

src/Models/Constants/Interfaces.ts
```diff
@@ -141,10 +142,26 @@ export interface IConsumeCompositeCardProps extends ICardBaseProps {
 }
 
 export interface IAuthService {
-    login: () => void;
-    getToken: (
-        tokenFor?: 'azureManagement' | 'adx' | 'storage' | 'powerBI'
-    ) => Promise<string>;
+    /** the unique object id of the logged in user */
+    userObjectId?: string;
```
msnyder-msft, line 146: do we have to expose teh value directly or can we expose a setter and getter? Feels weird for a service to let other components set a property directly.

src/Models/Constants/Interfaces.ts
```diff
@@ -141,10 +142,26 @@ export interface IConsumeCompositeCardProps extends ICardBaseProps {
...
+    setActiveAccount: (account: any) => void;
+    login: (continuation?: (params: any) => void) => void;
```
msnyder-msft, line 148: should the continuation function take the IAuthParams object? or do we really not know the type?

src/Models/Constants/Interfaces.ts
```diff
@@ -141,10 +142,26 @@ export interface IConsumeCompositeCardProps extends ICardBaseProps {
...
+    /** base login url for the auth service authority */
+    login: string;
+    clientId: string;
+    tenantId?: string;
+}
+
+export interface IMSALLoginContinuationParams {
+    name: string;
+    userName: string;
+    tenants: Array<any>;
```
msnyder-msft, line 163: can we get any more specific than `any`? Can it be Record<string, string> or something? Do we at least know it's an object?

src/Models/Services/MsalAuthService.ts
```diff
@@ -149,58 +268,130 @@ export default class MsalAuthService implements IAuthService {
...
+        const setAccountAndContinue = (response) => {
+            if (response.account) {
+                this.setActiveAccount(response.account);
+            }
+            continuation?.();
+        };
+        const catchLogic = (err) => {
+            console.error('SSO failed');
+            console.error(err);
+            continuation?.();
```
msnyder-msft, line 380: why do we call continuation when we fail?

src/Models/Constants/Interfaces.ts
```diff
@@ -141,10 +142,26 @@ export interface IConsumeCompositeCardProps extends ICardBaseProps {
 }
 
 export interface IAuthService {
-    login: () => void;
-    getToken: (
-        tokenFor?: 'azureManagement' | 'adx' | 'storage' | 'powerBI'
-    ) => Promise<string>;
+    /** the unique object id of the logged in user */
+    userObjectId?: string;
+    setActiveAccount: (account: any) => void;
```
ptallettms, line 147: I think changing this interface breaks anyone that has implemented their own AuthService. I'm not suggesting we shouldn't change it, but all uses of 3D Scenes I have been involved with do this, so maybe we want to give a heads up.

ptallettms: We have various issues with our implementation of IAuthService which was based on the old cardboard code. We have started rewriting it over the next few days but will probably wait for this PR to pick up the new IAuthService definition.

1. When logged into a default domain (eg corpnet), attempting to auth against another directory where this user is a guest with the same email address, we get a ton of auth timeouts and popup dialogs. Eventually it does usually seem to work but its messy.
2. After a while of being logged in, the stream of popups occurs again (presumably refreshing a token or something).
3. Usually after getting this sequence a few times, we get browser error 431 (form fields too long) and the only fix is to restart the browser or clear the cache
4. We don't really want to use popups at all as all our users have to enable popups for the site. Preferable would be one redirect call at login that fetches all the scopes we are going to need at once (maybe pass in the scopes we want  as a login() parameter). The app would then only ever have one token to manage, so the issue of avoiding multiple simultaneous token requests goes away, dramatically simplifying the code. The getToken call would then just refresh "the" single super-token.
5. In our implementation, all the variables used for auth are set in a .env file (eg authority, tenant id, client id etc.). Its not clear to me if this PR supports that or not

Any new auth code should be tested with a parallel AAD directory where the user is a guest, but also where their login is different (select login account).

src/Models/Constants/Interfaces.ts
iremgokce: based on [getTenants ](https://learn.microsoft.com/en-us/rest/api/resources/tenants/list?tabs=HTTP)response I added the interface which has the keys that I am interested in.

src/Models/Constants/Interfaces.ts
iremgokce: no, continuation is a optional callback method after login is successful with purely customizable in terms of parameters. For example, in adt app we pass username, name, tenants and accounts as params to show the user name in user menu and pick up the certain account with a certain tenant id which is read from local storage or url. So it should be any in this case.

src/Models/Services/MsalAuthService.ts
iremgokce: this catch logic is called if the ssosilent fails with error which is not type of "InteractionRequiredAuthError" which triggers popup login. In all other cases, if ssosilent fails with other type of errors, we should not break the flow if user is setting "isloading" to false etc. in the continuation as we do in the public repo fork. @darsney can confirm though.

src/Models/Constants/Interfaces.ts
iremgokce: I added getter instead

src/Adapters/ADTAdapter.ts
iremgokce: Yeah, I moved the tenantId to the authService constructor which makes more sense. Lets discuss the breaking changes on Monday.

src/Models/Constants/Interfaces.ts
iremgokce: I changed the IAuthService not to break the current implementations with latest commit. You can review it and we can talk about all other breaking changes on Monday. Thanks for listing all the points!

PR #902, opened 2023-03-15
[forceCORS and use proxy]: Fixes for storage account url and handling missing CORS

iremgokce:
### Summary of changes 🔍 
Here is the problem and solution definition [document](https://microsoft.sharepoint.com/:fl:/s/7e83e18c-fe3f-4f63-8c39-8b65b78daa06/ETuL9In0nvhJuQL1ySg_MdEBSUiUVclra5ITBg-Ew1lIpg?e=eZhWa9&nav=cz0lMkZzaXRlcyUyRjdlODNlMThjLWZlM2YtNGY2My04YzM5LThiNjViNzhkYWEwNiZkPWIheHBmYWJCRUVCVWFMcF90MTJRbDV0ck04dzZyZHV4NU5xdi1PNWdJLU1MaEpNUU9qVkItUVE3dFhGcm9KbGlDSCZmPTAxUEFYQ0Q0UjNSUDJJVDVFNjdCRTNTQVhWWkVVRDZNT1ImYz0lMkYmZmx1aWQ9MSZhPUxvb3BBcHAmeD0lN0IlMjJ3JTIyJTNBJTIyVDBSVFVIeHRhV055YjNOdlpuUXVjMmhoY21Wd2IybHVkQzVqYjIxOFlpRjRjR1poWWtKRlJVSlZZVXh3WDNReE1sRnNOWFJ5VFRoM05uSmtkWGcxVG5GMkxVODFaMGt0VFV4b1NrMVJUMnBXUWkxUlVUZDBXRVp5YjBwc2FVTklmREF4VUVGWVEwUTBValJNVFVOYVZFUk9TRmxTUjB0V04xTkpOelJLUkZoWlZqSSUzRCUyMiUyQyUyMmklMjIlM0ElMjIzODQ4OWQyYy05MzI5LTRhMmQtYWNkMi1hYWI0MTUzM2YzZjYlMjIlN0Q%3D).
- Fixed undefined x-blob-host issue
- Extended missing CORS configuration for future non proxy dependency and handling it in UI for both existing proxy users and forceCORS users. **TLDR; There are [additional headers](https://github.com/microsoft/iot-cardboard-js/pull/902/files#diff-b2f45b00cd67f3258399cb2eec172f3ad2a6cb0552ac98fd38beb92ca91ce79cR538) we should support if we move to forceCORS to include headers which was already used in our proxy middleware. ~~This will let the existing proxy users to see the CORS error again to add the missing headers~~ and forceCORS or private link users to see a new error message with CORS related content with learn more link.**
- Some small bug-fixes

### Testing 🧪
You can test it by setting your blobContainerUrl to empty string in your secret.users file to test the first problem, and then for each scenario explained [here](https://microsoft.sharepoint.com/:fl:/s/7e83e18c-fe3f-4f63-8c39-8b65b78daa06/EYQi0zsmSJRIjyu-uywYvOsBAsbK5twv8CIcGwD_5cr-Sg?e=zwrHvK&nav=cz0lMkZzaXRlcyUyRjdlODNlMThjLWZlM2YtNGY2My04YzM5LThiNjViNzhkYWEwNiZkPWIheHBmYWJCRUVCVWFMcF90MTJRbDV0ck04dzZyZHV4NU5xdi1PNWdJLU1MaEpNUU9qVkItUVE3dFhGcm9KbGlDSCZmPTAxUEFYQ0Q0VUVFTEpUV0pTSVNSRUk2SzU2WE1XQlJQSEwmYz0lMkYmZmx1aWQ9MSZhPUxvb3BBcHAmcD0lNDBmbHVpZHglMkZsb29wLXBhZ2UtY29udGFpbmVyJng9JTdCJTIydyUyMiUzQSUyMlQwUlRVSHh0YVdOeWIzTnZablF1YzJoaGNtVndiMmx1ZEM1amIyMThZaUY0Y0daaFlrSkZSVUpWWVV4d1gzUXhNbEZzTlhSeVRUaDNObkprZFhnMVRuRjJMVTgxWjBrdFRVeG9TazFSVDJwV1FpMVJVVGQwV0VaeWIwcHNhVU5JZkRBeFVFRllRMFEwVWpSTVRVTmFWRVJPU0ZsU1IwdFdOMU5KTnpSS1JGaFpWakklM0QlMjIlMkMlMjJpJTIyJTNBJTIyMzg0ODlkMmMtOTMyOS00YTJkLWFjZDItYWFiNDE1MzNmM2Y2JTIyJTdE), changing your CORS settings of your strorage account in portal and come back to ADT3DScenePage story to see handling solutions for both when your cardboard.feature.forceCORS localstorage is set to either true or false. For private link, you can use Pato's adt instance which uses private link to initialize with in your secret file of change it in Environment Picker.

### Checklist ✔️
- [x] Added relevant labels


iremgokce:
I am leaving it in draft mode until we consolidate it this design is okay for handling adding necessary CORS configuration.

src/Adapters/BlobAdapter.ts
```diff
@@ -438,7 +447,7 @@ export default class BlobAdapter implements IBlobAdapter {
                             .map((mapping) => mapping.allowedHeaders)
                             .reduce(
                                 (acc: boolean, headerGroup) =>
-                                    acc && headerGroup[0].includes(header),
+                                    acc && headerGroup?.includes(header),
```
msnyder-msft, line 450: it's intentional that we now search the entire array instead of just the first entry?

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
```diff
@@ -188,7 +189,7 @@ const EnvironmentPicker = ({
         }
         if (isLocalStorageEnabled) {
             const selectedAdtInstanceInLocalStorage = getSelectedAdtInstanceFromLocalStorage(
-                localStorageKey
+                selectedItemLocalStorageKey
```
msnyder-msft, line 192: why do we need this? And can we remove the localSotrageKey prop if we have this new key instead?

src/Adapters/BlobAdapter.ts
```diff
@@ -67,21 +67,25 @@ export default class BlobAdapter implements IBlobAdapter {
...
         if (this.useBlobProxy) {
             return `${this.blobProxyServerPath}${path}`;
         } else {
-            // Need to have this done with only storage account since path will always include container name if required
-            return `${this.getStorageAccountURL()}${path}`;
+            if (this.getStorageAccountURL()) {
+                // Need to have this done with only storage account since path will always include container name if required
+                return `${this.getStorageAccountURL()}${path}`;
+            } else {
+                return null;
```
pasanchMSFT, line 87: Does the request fail if this happens?

src/Components/EnvironmentPicker/EnvironmentPicker.tsx
iremgokce: These are all needed, but I noticed I was using the wrong key here.

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
```diff
@@ -385,6 +395,26 @@ const ADT3DScenePageBase: React.FC<IADT3DScenePageProps> = ({
                         }
                     }
                 });
+            } else if (
+                state?.errors?.[0]?.type ===
+                    ComponentErrorType.DataFetchFailed &&
+                (state?.errors[0].rawError as any).request.status === 0 &&
```
msnyder-msft, line 401: can we make a function to do this complex check? like `isCallSuccessful` or somehting? i see we do this in a couple spots.

pasanchMSFT (commented):
Quick comment, but generally looks good

msnyder-msft (commented):
Looks ok at a high level

src/Adapters/BlobAdapter.ts
iremgokce: I changed it since I wanted to get this loweCase comparsion.

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
iremgokce: I am interested in **status code 0** specifically which happens when you cannot even make a call which is basically a "Network Error" error which might be due to CORS as well. I use this approach in other places in data history etc. So it is not just successful check but a specific status code to relate to CORS.

src/Adapters/BlobAdapter.ts
iremgokce: yes, this is why added null instead of empty string which it was used to. remember, even it is empty string axios makes the call against the window's location. This way, the axis will fail since the url is null.

src/Pages/ADT3DScenePage/ADT3DScenePage.tsx
msnyder-msft: ok. so then maybe we can have a getStatusCode function? It's just a lot of nested properties and null checks to get to the value we care about.

src/Adapters/BlobAdapter.ts
pasanchMSFT: Alright, but how will the user fix this then? Will they be prompted to select a new container or something?

iremgokce: this was going to be fixed when the deeplink state picks the url from localstorage. but if there is no url in the local storage either, let me check the scenario not to block the UI to pick the environment configurations.

iremgokce: okay, so I am now making getCorsProperties call only if the deeplink context has storage url value (used to still do it and axios was adding window location but throwing 500 error which was out of our interest to surface in the UI - not CORS related error). Not I just changed it so that we dont make get cors properties call unncessarily. In this way, we let the app continue its flow as usual, meaning if there is no storage url or adt url in the deeplink context, we show "Configure environment" error in the UI.

iremgokce: So, yes user is prompted to select new container if it is not present in the deeplink context.

src/Resources/Locales/en.json
```diff
@@ -1341,6 +1341,8 @@
         "unknownErrorTitle": "Unknown error",
         "corsErrorTitle": "CORS error",
         "corsErrorMessage": "Your storage account needs to have certain CORS properties enabled to be able to use 3D assets in your scenes.",
+        "partialCorsErrorTitle": "CORS update required",
+        "partialCorsErrorMessage": "Due to recent feature changes, there are some CORS rules which need to be updated on your storage account to be able to continue using 3D assets in your scenes.",
```
msnyder-msft, line 1345: I would remove the word Feature. "In order to keep using 3D scenes, there are some..."

src/Resources/Locales/en.json
```diff
@@ -1358,7 +1360,10 @@
         "noADTInstanceUrlTitle": "ADT instance not set",
         "noADTInstanceUrlMessage": "Your ADT instance url is not set. Click below to set it.",
         "configureEnvironment": "Configure environment",
+        "networkErrorTitle": "Connection error",
         "networkErrorMessage": "This could be a CORS issue, invalid url or a network error. Please contact the owner of the resource or try again later.",
+        "applyUpdates": "Apply updates",
+        "setCorsPropertiesNotAuthorizedMessage": "You do not have access to the storage account to make this change. Please contact the resource owner to make the required changes on CORS settings of the storage account.",
```
msnyder-msft, line 1366: do we still direct them to the docs somehwere so they know _which_ CORS settings to fix?

iremgokce: So I am going to ask Bailey (our content person) to add these headers to the [prerequisite docs](https://learn.microsoft.com/en-us/azure/digital-twins/how-to-use-3d-scenes-studio#prerequisites) where we talk about CORS headers to add. But should we update these docs now @darsney ?

darsney (approved):

PR #907, opened 2023-03-20
Data pusher fix duplication in URL

pasanchMSFT:
### Summary of changes 🔍 
Fixes duplication created by setting a baseURL in the axios instance used for data pusher requests. URL is now managed by `generateURL` function exclusively.

`this.axiosInstance` can only be found in the following methods of ADT Adapter:
- createModels (only found in `GenerateADTAssets.tsx` a data pusher component)
- createTwins (only found in `GenerateADTAssets.tsx` a data pusher component) 
- createRelationships (only found in `GenerateADTAssets.tsx` a data pusher component)

### Testing 🧪
Try out data pusher stories locally with both CORS on and off.

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/src/Adapters/ADTAdapter.ts
+++ b/src/Adapters/ADTAdapter.ts
@@ -128,11 +128,7 @@ export default class ADTAdapter implements IADTAdapter {
             (useAdtProxy || !validateExplorerOrigin(window.origin)) &&
             !forceCORS;
         this.authService.login();
-        this.axiosInstance = axios.create({
-            baseURL: this.useAdtProxy
-                ? this.adtProxyServerPath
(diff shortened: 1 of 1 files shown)


msnyder-msft:
we have had several point fix PRs. How are we doing validations to ensure we don't need more fixes like this?

pasanchMSFT:
> we have had several point fix PRs. How are we doing validations to ensure we don't need more fixes like this?

Basically running scenarios manually, would be useful to have testing for these adapter methods but currently we don't have that in place.

msnyder-msft (approved):

PR #910, opened 2023-03-22
Adding some tools to quickly build Legion components and Shell component

pasanchMSFT:
### Summary of changes 🔍 
Adding Shell component, removing Storybook non-legion stories and making base components for Legion available for quick component creation in Legion folder.

### Testing 🧪
N/A

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -6,8 +6,11 @@ const AppSourceDir = path.join(__dirname, '..', 'src');
 module.exports = {
     stories:
         process.env.NODE_ENV === 'production'
-            ? ['../src/**/*.stories.tsx']
-            : ['../src/**/*.stories.tsx', '../src/**/*.stories.local.tsx'],
+            ? ['../src/Apps/**/*.stories.tsx']
+            : [
+                  '../src/Apps/**/*.stories.tsx',
+                  '../src/Apps/**/*.stories.local.tsx'
+              ],
     // Add any Storybook addons you want here: https://storybook.js.org/addons/
     addons: [
         '@storybook/addon-essentials',
--- a/src/Apps/Legion/Services/StoryUtilities.tsx
+++ b/src/Apps/Legion/Services/StoryUtilities.tsx
@@ -1,0 +1,156 @@
+/**
+ * @summary This file contains helper functions for composing Story tests
+ */
+
+import React, { CSSProperties } from 'react';
+import { fireEvent, screen } from '@storybook/testing-library';
+import { ReactFramework, StoryContext } from '@storybook/react';
+
+/**
+ * Context passed into the story decorator.
+ * <T> is the type for the props of the component
+ */
+export type IStoryContext<T> = StoryContext<
+    ReactFramework,
+    T & {
+        children?: React.ReactNode;
+    }
+>;
+
+/**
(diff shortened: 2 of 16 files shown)


.storybook/main.js
```diff
@@ -6,8 +6,11 @@ const AppSourceDir = path.join(__dirname, '..', 'src');
 module.exports = {
     stories:
         process.env.NODE_ENV === 'production'
-            ? ['../src/**/*.stories.tsx']
-            : ['../src/**/*.stories.tsx', '../src/**/*.stories.local.tsx'],
+            ? ['../src/Apps/**/*.stories.tsx']
+            : [
```
msnyder-msft, line 10: won't this hide everything in storybook except Legion?

src/Apps/Legion/Services/StoryUtilities.tsx
```diff
@@ -0,0 +1,156 @@
+/**
+ * @summary This file contains helper functions for composing Story tests
+ */
```
msnyder-msft, line 3: do we really need a duplicate for this?

tools/generate-legioncomponent-template.js
```diff
@@ -0,0 +1,58 @@
+require('colors');
```
msnyder-msft, line 1: do we need separate scripts for this? Can't we just pass in a path or something to teh existing script? and/or use a different template but use the same script?

msnyder-msft, line 1: Where are you invoking this script from? IF we are going to have a separate script for legion components, then how do we call it? There should be a package.json change to go with this.

msnyder-msft: I would rather extend the script to take an arg for AppName or something and then it can put the componets in the right folder based on that name.

.storybook/main.js
pasanchMSFT: to reduce build times for this feature branch

tools/generate-legioncomponent-template.js
pasanchMSFT: I like that suggestion, I'll create one that takes AppName.

pasanchMSFT: there is a package.json change that invokes the script

src/Apps/Legion/Services/StoryUtilities.tsx
pasanchMSFT: Removed duplicates, I'll bring them back if our plan is to lift and shift the App folder.

src/Apps/Legion/Services/Utils.ts
```diff
@@ -0,0 +1,29 @@
+import { IConsoleLogFunction } from '../Models/Types';
+
+export function getDebugLogger(
```
iremgokce, line 3: So I asked this to Matt and he said we can use the existing methods or whatnot by referencing, and he said dont worry about dependencies (so no duplicates). Can you confirm @darsney?

iremgokce (approved):
Added some comments about duplicates vs dependencies to confirm with Matt, but approved

src/Apps/Legion/Services/Utils.ts
pasanchMSFT: removed duplicates

PR #922, opened 2023-04-10
Adding store list data to local storage

msnyder-msft:
### Summary of changes 🔍 
Moving state management for Store data persistence to local storage. Creating a store now persists it to local storage so you can build up a list of them and see them on subsequent runs. They are stored on the `Legion.TargetDatabases` key.
image

- Created a `LocalStorageManager` for doing local storage read and write operations
- Created a StoreFormModal for creating new stores. Shows when + button is clicked on the list page
- Add cluster URL to the payload for a store (eventually will be hidden since it's system managed)

image

### Testing 🧪
`path=/story/apps-legion--legion`
- Add new store --> Navigates to next screen & persisted across sessions
- Navigating back after creating/selecting will persist selection in list

### DIFF
--- a/src/Apps/Legion/Components/StoreFormModal/StoreFormModal.stories.tsx
+++ b/src/Apps/Legion/Components/StoreFormModal/StoreFormModal.stories.tsx
@@ -1,0 +1,31 @@
+/* eslint-disable no-console */
+import React from 'react';
+import { ComponentStory } from '@storybook/react';
+import { getDefaultStoryDecorator } from '../../../../Models/Services/StoryUtilities';
+import StoreFormModal from './StoreFormModal';
+import { IStoreFormModalProps } from './StoreFormModal.types';
+
+const wrapperStyle = { width: '100%', height: '600px', padding: 8 };
+
+export default {
+    title: 'Components/StoreFormModal',
+    component: StoreFormModal,
+    decorators: [getDefaultStoryDecorator<IStoreFormModalProps>(wrapperStyle)]
+};
+
+type StoreFormModalStory = ComponentStory<typeof StoreFormModal>;
+
+const Template: StoreFormModalStory = (args) => {
+    return <StoreFormModal {...args} />;
+};
+
+export const Base = Template.bind({}) as StoreFormModalStory;
+Base.args = {
+    isOpen: true,
+    onConfirm: (args) => {
+        console.log('OnConfirm', args);
+    },
+    onDismiss: () => {
+        console.log('OnDismiss');
+    }
+} as IStoreFormModalProps;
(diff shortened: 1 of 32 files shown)


src/Apps/Legion/Components/StoreFormModal/StoreFormModal.stories.tsx
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable no-console */
```
pasanchMSFT, line 1: To avoid this, why not do Alert instead of console.log?

src/Apps/Legion/Components/StoreFormModal/StoreFormModal.tsx
```diff
@@ -0,0 +1,111 @@
...
+    const [
+        databaseConnection,
+        setDatabaseConnection
+    ] = useState<ITargetDatabaseConnection>(EMPTY_DATABASE_CONNECTION);
+
+    // hooks
+    const { t } = useTranslation();
+
+    // data
+    const isFormValid = useMemo(() => {
```
pasanchMSFT, line 44: this isn't particularly expensive, I don't think we need the useMemo

pasanchMSFT (approved):
micro-nits, all good otherwise

src/Apps/Legion/Components/StoreFormModal/StoreFormModal.stories.tsx
msnyder-msft: we could. I figured it was less intrusive, but that would work.

src/Apps/Legion/Components/StoreFormModal/StoreFormModal.tsx
msnyder-msft: that's fair. Will take a look.

montgomp (approved):
👍🏻

src/Resources/Locales/en.json
```diff
@@ -1529,7 +1529,14 @@
         "Common": {},
         "StoreListPage": {
             "title": "Welcome to the future",
-            "createStoreButtonText": "Create store"
+            "createStoreButtonText": "Create store",
+            "urlLabel": "Cluster url",
+            "nameLabel": "Graph name",
+            "Modal": {
+                "title": "Create a new graph",
```
iremgokce, line 1536: Should it be "Create a new store"?

src/Apps/Legion/Components/StoreFormModal/StoreFormModal.tsx
msnyder-msft: Done

PR #927, opened 2023-04-13
Legion: Add hooks & types for accessing new data model

msnyder-msft:
### Summary of changes 🔍 
Adding new types for defining the DB and ViewModels for the app level data for the following items:
- `Entity`
- `Type`
- `Relationship`
- `RelationshipType`
- `Property`

Each type has a base interface (ex: `IEntityBase`) which contains the common attributes of the item. Then there is a DB version with the foreign key references and the View version which has the child items resolved into full objects.

image

We also introduced a new `WizardDataContext` that hosts these entities in their denormalized structure with foreign key references. Then we use hooks to access that context and hydrate each type of entity so that components can use the data directly without needing to link data across the different entities.
- `useEntities`
- `useTypes`
- `useRelationships`
- `useProperties`

Each hook exposes methods for doing updates on that kind of entity and will push it back into the global state.

NOTE: the callback implementations are not deep operations. We will have to decide which callbacks will propogate changes to the child items. Ex: if you update a `property` on a `type` of an `entity` do we reflect that everywhere? What's the right model for doing that and the cross dependency between hooks? For now, this PR will have a naive implementation and will give us the types and functions to call. The actual implementations can get revved in future PRs.

### Testing 🧪
This has not yet been consumed anywhere.


src/Apps/Legion/Models/AppTypes.ts
```diff
@@ -0,0 +1,126 @@
+/**
+ * This file is for containing the top level types for use across the entire app. It should not contain any types specific to particular components or segments of the app.
```
iremgokce, line 2: is there a reason why this is called "AppTypes" but not just "Types"?

src/Apps/Legion/Services/AppTypeUtilities.ts
```diff
@@ -0,0 +1,162 @@
...
+        friendlyName: viewModel.friendlyName,
+        id: viewModel.id,
+        sourceConnectionString: viewModel.sourceConnectionString,
+        sourceEntityId: viewModel.sourceEntityId,
+        values: viewModel.values, // TODO: do we need to serialize each property value into a string?
+        typeId: viewModel.type.id
+    };
+}
+
+export function convertEntityToView(
```
iremgokce, line 26:
```suggestion
export function convertDbEntityToView(
```

src/Apps/Legion/Services/AppTypeUtilities.ts
```diff
@@ -0,0 +1,162 @@
...
+    IDbType,
+    IViewEntity,
+    IViewProperty,
+    IViewRelationship,
+    IViewRelationshipType,
+    IViewType
+} from '../Models';
+
+// #region Entities
+export function convertEntityToDb(viewModel: IViewEntity): IDbEntity {
```
iremgokce, line 15:
```suggestion
export function convertViewEntityToDb(viewModel: IViewEntity): IDbEntity {
```

src/Apps/Legion/Services/AppTypeUtilities.ts
```diff
@@ -0,0 +1,162 @@
...
+export function convertEntityToView(
+    dbModel: IDbEntity,
+    state: {
+        types: IDbType[];
+        properties: IDbProperty[];
+    }
+): IViewEntity {
+    const type = state.types.find((x) => x.id === dbModel.typeId);
+    return {
+        friendlyName: dbModel.friendlyName,
```
iremgokce, line 35: same comment for the following methods to use spread method to copy the same fields to the new object except some field

src/Apps/Legion/Services/AppTypeUtilities.ts
```diff
@@ -0,0 +1,162 @@
...
+    IDbRelationshipType,
+    IDbType,
+    IViewEntity,
+    IViewProperty,
+    IViewRelationship,
+    IViewRelationshipType,
+    IViewType
+} from '../Models';
+
+// #region Entities
```
iremgokce, line 14: Same question here, why not just "TypeUtilities" as the name of the file?

iremgokce:
Ahh it is just got merged, please take a look at @msnyder-msft if any of those make sense moving forward.

src/Apps/Legion/Models/AppTypes.ts
msnyder-msft: Yeah. I wanted a file just for the data types we use for the wizard. Maybe I should have called it Wizard.types.ts.

src/Apps/Legion/Services/AppTypeUtilities.ts
```diff
@@ -0,0 +1,162 @@
...
+export function convertEntityToDb(viewModel: IViewEntity): IDbEntity {
+    return {
+        friendlyName: viewModel.friendlyName,
```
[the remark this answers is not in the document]
msnyder-msft: I thought about it, but decided not to. Spread will also include other data in the type even if it's not on the type definition so I wanted to have a clean conversion that doesn't bring anything extra in.

PR #930, opened 2023-04-17
Legion: Add `isNew` and `isDeleted` fields to wizard type definitions

msnyder-msft:
### Summary of changes 🔍 
- Adding the `isNew` and `isDeleted` flags to all the type definitions in the wizard.
- Updated the context to soft delete by setting the flag instead of removing from the collection

### DIFF
--- a/src/Apps/Legion/Services/WizardTypes.utils.ts
+++ b/src/Apps/Legion/Services/WizardTypes.utils.ts
@@ -72,22 +72,22 @@ export function convertViewTypeToDb(viewModel: IViewType): IDbType {
 }
 
 export function convertDbTypeToView(
-    dbModel: IDbType,
+    dbType: IDbType,
     state: {
         properties: IDbProperty[];
     }
 ): IViewType {
     const properties: IViewProperty[] = [];
-    dbModel.propertyIds.forEach((id) => {
+    dbType.propertyIds.forEach((id) => {
         const property = state.properties.find((x) => x.id === id);
         property && properties.push(convertDbPropertyToView(property));
     });
     return {
-        ...getBase(dbModel),
-        friendlyName: dbModel.friendlyName,
-        color: dbModel.color,
-        icon: dbModel.icon,
-        kind: dbModel.kind,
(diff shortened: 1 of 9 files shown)


src/Apps/Legion/Services/WizardTypes.utils.ts
```diff
@@ -11,19 +11,35 @@ import {
     IViewType
 } from '../Models';
 
+// #region Base
+
+function initializeBase(args: {
```
iremgokce, line 16: in fact this is not initialization I think, just fetching base properties of objects. I mean seeing ...initializeBase(something) makes me think if it is setting initial values.
```suggestion
function getBase(args: {
```

msnyder-msft: Agreed. Done.

iremgokce:
```suggestion
function initializeBase(args: IBase): IBase
```

msnyder-msft: I didn't export the base class since I didn't want anything to take a dependency on that low of an interface.

iremgokce: okay

iremgokce (approved):

src/Apps/Legion/Services/WizardTypes.utils.ts
```diff
@@ -32,30 +48,30 @@ export function convertEntityToView(
...
         color: viewModel.color,
         icon: viewModel.icon,
         kind: viewModel.kind,
         propertyIds: viewModel.properties.map((x) => x.id)
     };
 }
 
-export function convertTypeToView(
+export function convertDbTypeToView(
     dbModel: IDbType,
```
iremgokce, line 75:
```suggestion
    dbType: IDbType,
```

msnyder-msft: Fixed

PR #931, opened 2023-04-17
New non-styled templates

pasanchMSFT:
### Summary of changes 🔍 
Adding some templates for Legion that do not have Fluent's `styled` pattern. Now npm run generate can take a `new` flag to use new templates.

### Testing 🧪
N/A

### Checklist ✔️
- [ ] Added relevant labels

### DIFF
--- a/tools/generate-component-template.js
+++ b/tools/generate-component-template.js
@@ -1,6 +1,7 @@
 require('colors');
 const fs = require('fs');
 const componentTemplates = require('./templates/component');
+const legionTemplates = require('./templates/legion-component');
 const { exec } = require('child_process');
 
 // Grab and validate componentName from script args
@@ -8,6 +9,9 @@ const componentName = process.argv[2];
 
 // Grab and validate appName from script args
 const appName = process.argv[3];
+
+// Check if new templates flag is present
+const newTemplates = process.argv[4] == 'new';
 
 if (!componentName) {
     console.error(
@@ -62,9 +66,18 @@ fs.mkdirSync(componentDirectory);
 
 // Generate sub component /Consume and /Create templates
 
-const generatedTemplates = componentTemplates.map((template) =>
-    template(componentName, componentDirectory)
-);
+let generatedTemplates;
+
+if (newTemplates) {
+    console.log('New templates being used');
+    generatedTemplates = legionTemplates.map((template) =>
+        template(componentName, componentDirectory)
+    );
+} else {
+    generatedTemplates = componentTemplates.map((template) =>
+        template(componentName, componentDirectory)
+    );
+}
 
 generatedTemplates.forEach((template) => {
     fs.writeFileSync(
--- a/tools/templates/legion-component/component.js
+++ b/tools/templates/legion-component/component.js
@@ -1,0 +1,35 @@
+module.exports = (componentName) => ({
+    content: `import React from 'react';
+    import { getDebugLogger } from '../../../../Models/Services/Utils';
+    import {
+        I${componentName}Props
+    } from './${componentName}.types';
+    import { getStyles } from './${componentName}.styles';
+
+    const debugLogging = false;
+    const logDebugConsole = getDebugLogger('${componentName}', debugLogging);
+    
+    const ${componentName}: React.FC<I${componentName}Props> = (_props) => {
+        // contexts
+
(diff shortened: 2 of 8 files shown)


tools/generate-component-template.js
```diff
@@ -9,6 +10,9 @@ const componentName = process.argv[2];
 // Grab and validate appName from script args
 const appName = process.argv[3];
 
+// Check if new templates flag is present
+const newTemplates = process.argv[4] == 'new';
```
msnyder-msft, line 14: this would work, but might be simpler to just have `npm run generate-legion <same args as before>`

tools/templates/legion-component/component.js
```diff
@@ -0,0 +1,35 @@
+module.exports = (componentName) => ({
+    content: `import React from 'react';
+    import { getDebugLogger } from '../../../../Models/Services/Utils';
```
msnyder-msft, line 3: are these paths all correct relative to the root of the Legion Components folder?

tools/templates/legion-component/component.types.js
```diff
@@ -0,0 +1,14 @@
+module.exports = (componentName) => ({
+    content: `import { IStyle } from '@fluentui/react';
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    export interface I${componentName}Props {
+    }
+    
+    export interface I${componentName}Styles {
```
msnyder-msft, line 8: do we gain anything from having the strict type definition on the styles? in some ways it's nice not to have to be explicit and simply be able to add properties in the style file. But if this lets us give it a type besides IStyle (like we do with sub component styles) then I'm for that.

tools/templates/legion-component/component.styles.js
```diff
@@ -0,0 +1,16 @@
+
+module.exports = (componentName) => ({
+    content: `import {
+        I${componentName}Styles
+    } from './${componentName}.types';
+    import { IProcessedStyleSet, mergeStyleSets } from '@fluentui/react';
+
+    export const getStyles = (): IProcessedStyleSet<I${componentName}Styles> => {
+        return mergeStyleSets({
```
msnyder-msft, line 9: how do you style fluent controls in this pattern? Like when you waht to change a combo box or button style, can you still use the nested properties with strong types?

tools/generate-component-template.js
```diff
@@ -62,9 +66,18 @@ fs.mkdirSync(componentDirectory);
...
 // Generate sub component /Consume and /Create templates
 
-const generatedTemplates = componentTemplates.map((template) =>
-    template(componentName, componentDirectory)
-);
+let generatedTemplates;
+
+if (newTemplates) {
+    console.log('New templates being used');
+    generatedTemplates = legionTemplates.map((template) =>
```
msnyder-msft, line 73: can we spit these out in the Legion Components folder?

tools/templates/legion-component/component.js
pasanchMSFT: yeah, I tested it for Legion folders

tools/templates/legion-component/component.styles.js
pasanchMSFT: I haven't been able to replicate this behavior, even with the classnames. I need to dig deeper to find out how, however as a temporary solution we could define the styles prop value for a subcomponent in the styles file.

tools/templates/legion-component/component.types.js
pasanchMSFT: This definitely allows us to add other types (like IButtonStyles) but I haven't figured out how to use those with mergeStyleSets.

msnyder-msft (approved):

tools/templates/legion-component/component.styles.js
```diff
@@ -0,0 +1,16 @@
+
+module.exports = (componentName) => ({
+    content: `import {
+        I${componentName}Styles
+    } from './${componentName}.types';
+    import { IProcessedStyleSet, mergeStyleSets } from '@fluentui/react';
+
+    export const getStyles = (): IProcessedStyleSet<I${componentName}Styles> => {
```
msnyder-msft, line 8: can we add the `memoizeFunction` wrapper here so we don't recreate the styles?

PR #933, opened 2023-04-18
Legion: Create Parent modal for graph

msnyder-msft:
### Summary of changes 🔍 

Create a new entity
image

Use an existing entity to make the parent
image

The parent types populate from existing types but user can create a new one if desired
image

Relationship options populate from existing values, but user can create a new name if desired.
image

### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


.github/workflows/build.yml
```diff
@@ -1,7 +1,7 @@
 name: build
 on:
     push:
-        branches: [main]
+        branches: [main, feature-Legion]
```
pasanchMSFT, line 4: will this publish the package to npm?

package.json
```diff
@@ -28,6 +28,7 @@
         "post-build": "npm run copy-theme-sheet && npm run copy-package && npm run copy-readme && npm run pack",
         "prettier:check": "prettier --check src/**/*.{js,ts,tsx,scss}",
         "prettier:fix": "prettier --write src/**/*.{js,ts,tsx,scss}",
+        "start": "npm run storybook",
```
pasanchMSFT, line 31: nice QOL improvement

src/Apps/Legion/Components/UserDefinedEntityForm/UserDefinedEntityForm.view.tsx
```diff
@@ -0,0 +1,415 @@
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
```
pasanchMSFT, line 1: `.view.tsx`? Why not a new component name?

src/Apps/Legion/Components/UserDefinedEntityForm/UserDefinedEntityForm.view.tsx
```diff
@@ -0,0 +1,415 @@
...
+        relationshipNameOptions,
+        selectedExistingEntity?.key,
+        t
+    ]);
+
+    logDebugConsole('debug', 'Render', parentTypeOptions, selectedParentType);
+
+    return (
+        <Stack className={classNames.root} tokens={{ childrenGap: 8 }}>
+            {formMode === 'New' ? newEntityForm : existingForm}
```
pasanchMSFT, line 406: Would we benefit from making each of these its own component?

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/ModifyStep.tsx
```diff
@@ -24,7 +25,8 @@ const ModifyStep: React.FC<IModifyStepProps> = (props) => {
         {
             disabled: false,
             onClick: () => {
-                console.log('Open modal here');
+                alert('open modal');
+                // console.log('Open modal here');
```
pasanchMSFT, line 29: remove comment

src/Apps/Legion/Components/WizardShell/WizardShellMockData.ts
```diff
@@ -320,35 +320,35 @@ const setDairyData = (data: IWizardDataContextState) => {
     // TYPES
     const typePastTs = getType({
         friendlyName: 'Pasteurizer TS',
-        color: 'C13099',
+        color: '#C13099',
```
pasanchMSFT, line 323: Thank you for adding these, just please look at the `TypeIcon` component to make sure it didn't break, because I added the `#` in its styles file.

pasanchMSFT (commented):
A couple of nits here and there.

.github/workflows/build.yml
msnyder-msft: Nope. It just runs the build. There is a different YML file for release.

package.json
msnyder-msft: yeah. just seems like it aligns with stadards. You still have to run `npm run start` but at least you don't have to know we use storybook.

src/Apps/Legion/Components/UserDefinedEntityForm/UserDefinedEntityForm.view.tsx
msnyder-msft: Was trying something new. This is the renderer and the main component fetches the data.

src/Apps/Legion/Components/WizardShell/WizardShellMockData.ts
msnyder-msft: Good callout. Fixed it.

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/ModifyStep.tsx
msnyder-msft: Done

src/Apps/Legion/Components/UserDefinedEntityForm/UserDefinedEntityForm.view.tsx
msnyder-msft: Was trying something new. This is the renderer and the main component fetches the data.

src/Apps/Legion/Components/UserDefinedEntityForm/UserDefinedEntityForm.view.tsx
msnyder-msft: I considered that. But there's enough overlap that I just thought it made sense to keep them together. You could probably split them and have it be fine but I didn't want to have to keep the props in sync between them and whatnot so this seemed simpler.

msnyder-msft: There's also a very good chance this from will go away entirely so I figured it was better to just get this branch merged and if we still need it for H2 we will revamp it to fit the new models.

src/Apps/Legion/Components/UserDefinedEntityForm/UserDefinedEntityForm.view.tsx
pasanchMSFT: Naming sounds good, but this looks very complicated to just be a renderer. How much data manipulation could be passed down as a prop or be extracted to hooks?

pasanchMSFT (approved):

src/Apps/Legion/Components/UserDefinedEntityForm/UserDefinedEntityForm.view.tsx
msnyder-msft: Well it takes care of knowing which type of form to show. In my opinion that is what the renderer should do. The parent doesn't need to know that, it just wants to know the output of the form and takes care of persisting that to state. Since both versions of the form result in the same data it seemed reasonable to keep that abstracted away from the parent.

msnyder-msft: Is it the lookups to find existing entities that you're talking about? Otherwise, I think most of the logic here is just transforming data from the form back into a standard form to send to the parent. I did go back and forth though about whether to resolve the "existing" items in the renderer vs in the parent. I do think it'd probably have been "ideal" to do that in the parent, but that meant that the child then would have different props back based on the selections and I opted for the cleaner interface. You're right though, it would be nice if the form would just report out on what was selected or entered and not have to care about other data in the system.

PR #940, opened 2023-04-27
Legion: Types tab

pasanchMSFT:
### Summary of changes 🔍 
Added components and styling for Types tab. Types name state persists, however selecting/de-selecting properties is still not enabled.

image

### Testing 🧪
Open Wizard shell stories and select Types tab in Modify step.

### Checklist ✔️
- [ ] Added relevant labels


src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/Internal/TypesCard/TypesCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+                        console.log(checked);
+                    }}
+                    defaultChecked={!item.isDeleted}
+                />
+            );
+        }
+    },
+    {
+        key: 'source-name-column',
+        name: 'Source property name',
```
msnyder-msft, line 44: we'll want to loc these eventaully. But can probably punt for now.

src/Apps/Legion/Hooks/useEntities.tsx
```diff
@@ -97,6 +117,8 @@ export const useEntities = () => {
          * Callback to delete the entity from state.
          * NOTE: this is not a deep update. It will only delete the root level element
          */
-        deleteEntity: deleteEntity
+        deleteEntity: deleteEntity,
+        /** Callback to get entity count per state based on Type */
+        getEntityStateCount: getEntityStateCount
```
msnyder-msft, line 122: does this need to be a function? can't it just be an object that gets returned? The function would have to be recreated on every state update anyways so I don't think you gain much by exposing a function over the actual result itself.
Thoughts?

src/Apps/Legion/Hooks/useTypes.tsx
```diff
@@ -75,23 +75,32 @@ export const useTypes = () => {
...
         /**
          * Callback to delete the type from state.
          * NOTE: this is not a deep update. It will only delete the root level element
          */
-        deleteProperty: deleteType
+        deleteType: deleteType,
+        /**
+         * Callback to get a type from an id
+         */
+        getTypeById: getTypeById
```
msnyder-msft, line 104: this is an interesting idea. 🤔 I guess it makes sense over having some utility. Good add.

src/Apps/Legion/Hooks/useTypes.tsx
```diff
@@ -75,23 +75,32 @@ export const useTypes = () => {
         [wizardDataState]
     );
 
+    // Single item
+    const getTypeById = (typeId: string): IViewType => {
```
msnyder-msft, line 79: let's make this a useCallback so that it doesn't recreate the object down below and bust memoization

src/Apps/Legion/Models/Wizard.types.ts
```diff
@@ -44,6 +44,12 @@ export interface IViewEntity extends IBaseEntity {
     /** the resolved view model of the `Type` the entity has */
     type: IViewType;
 }
+
+export interface IEntityCounters {
+    discovered: number;
```
msnyder-msft, line 49: can we call it "new" intead of "discovered" since that's our semantic in the type?

src/Resources/Locales/en.json
```diff
@@ -1610,7 +1610,12 @@
             "secondaryButton": "Merge entities",
             "idColumnHeader": "Entity ID",
             "typeColumnHeader": "Type",
-            "exampleColumnHeader": "Example values"
+            "exampleColumnHeader": "Example values",
+            "typeFieldLabel": "Type name",
+            "kind": "Kind",
+            "discoveredEntities": "{{entityCount}} discovered entities",
```
msnyder-msft, line 1616: do we want to handle plural vs singular?

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/TypesTab.styles.ts
```diff
@@ -0,0 +1,21 @@
...
+export const getStyles = memoizeFunction(
+    (): IProcessedStyleSet<ITypesTabStyles> => {
+        return mergeStyleSets({
+            root: {
+                maxHeight: 400,
+                overflowY: 'auto',
+                display: 'flex',
+                flexDirection: 'column',
+                rowGap: 8,
+                marginTop: 8
```
msnyder-msft, line 17: is `rowGap` how you get the Stack functionality?

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/TypesTab.styles.ts
```diff
@@ -0,0 +1,21 @@
...
+    IProcessedStyleSet,
+    mergeStyleSets,
+    memoizeFunction
+} from '@fluentui/react';
+
+export const getStyles = memoizeFunction(
+    (): IProcessedStyleSet<ITypesTabStyles> => {
+        return mergeStyleSets({
+            root: {
+                maxHeight: 400,
```
msnyder-msft, line 12: I don't love that teh tab has to know how big the modal is going to be. 🤔 I'm assuming this is what causes scrolling to work right?

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/TypesTab.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+export default {
+    title: 'Apps/Legion/TypesTab',
+    component: TypesTab,
+    decorators: [getDefaultStoryDecorator<ITypesTabProps>(wrapperStyle)]
+};
+
+type TypesTabStory = ComponentStory<typeof TypesTab>;
+
+const Template: TypesTabStory = (args) => {
+    return <TypesTab {...args} />;
```
msnyder-msft, line 18: do you think we should have stories for the tabs? or just have stories for the wizard and interactions that navigate you to the right tab? Then we don't have to set up all the contexts and mock data in both places.

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/Internal/TypesCard/TypesCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+import { useEntities } from '../../../../../../../../Hooks/useEntities';
+import TypeIcon from '../../../../../../../TypeIcon/TypeIcon';
+import { IViewProperty } from '../../../../../../../../Models';
+import { useExtendedTheme } from '../../../../../../../../../../Models/Hooks/useExtendedTheme';
+import { useTranslation } from 'react-i18next';
+
+const debugLogging = false;
+const logDebugConsole = getDebugLogger('TypesCard', debugLogging);
+
+const getColumns = (): IColumn[] => [
```
msnyder-msft, line 22: could use the `memoizeFunction` helper here to avoid rerendering the details list on every render of this component.

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/Internal/TypesCard/TypesCard.styles.ts
```diff
@@ -0,0 +1,42 @@
...
+                margin: 16
+            },
+            rightSide: {
+                flexGrow: 1,
+                margin: 8
+            },
+            leftSideTitle: {
+                marginBottom: 0,
+                marginTop: 4,
+                fontWeight: 600
```
msnyder-msft, line 32: I'd rather use the FontWeight enums if we can. But techincally is the same.

msnyder-msft (approved):
I didn't run the experience but the code looks clean and simple.

src/Apps/Legion/Hooks/useEntities.tsx
pasanchMSFT: I'll change the name to getEntityCounts

src/Apps/Legion/Hooks/useEntities.tsx
```diff
@@ -97,6 +117,8 @@ export const useEntities = () => {
          * Callback to delete the entity from state.
          * NOTE: this is not a deep update. It will only delete the root level element
          */
-        deleteEntity: deleteEntity
+        deleteEntity: deleteEntity,
```
iremgokce, line 120: we can just use the key names since they have same names with values variables.
```suggestion
        deleteEntity,
```

src/Apps/Legion/Hooks/useEntities.tsx
```diff
@@ -80,6 +80,26 @@ export const useEntities = () => {
...
+                    counters.discovered += 1;
+                } else {
+                    counters.existing += 1;
+                }
+            }
+        });
+        return counters;
+    };
+
     return {
```
iremgokce, line 103: lets return memoized object using useMemo as in useTimeSeriesData.ts hook

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/Internal/TypesCard/TypesCard.tsx
```diff
@@ -0,0 +1,131 @@
...
+    },
+    {
+        key: 'friendly-name-column',
+        name: 'Friendly name',
+        minWidth: 200,
+        fieldName: 'friendlyName'
+    }
+];
+
+const TypesCard: React.FC<ITypesCardProps> = (props) => {
```
iremgokce, line 56: I see that TypesCard is an internal element for TypesTab. But it creates two level of /Internal folder structure, can we make single level /Internal meaning moving the TypesCard one level above as a sibling to TypesTab internal component. The reason is that it is only one scingle component under second /Internal which I think does not require deep nesting.

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/TypesTab.stories.tsx
pasanchMSFT: I thought about that, I think I'll remove the tab stories, never really use them.

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/TypesTab.styles.ts
pasanchMSFT: yes, this is equivalent to the Stack token functionality

src/Apps/Legion/Hooks/useTypes.tsx
pasanchMSFT: this was one of the reasons why I wanted to move to hooks, have 'utility' functions in the same place where your objects are defined seemed neat.

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/TypesTab.styles.ts
pasanchMSFT: Will defer this to a later PR for a better solution.

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/TypesTab/Internal/TypesCard/TypesCard.tsx
pasanchMSFT: I think infinite nesting of internal components allows us to keep the same pattern regardless of depth. I like the consistency, but we can discuss if it is something that you would rather not have.

iremgokce (approved):

msnyder-msft (approved):

src/Apps/Legion/Models/Wizard.types.ts
```diff
@@ -44,6 +44,12 @@ export interface IViewEntity extends IBaseEntity {
     /** the resolved view model of the `Type` the entity has */
     type: IViewType;
 }
+
+export interface IEntityCounters {
+    new: number;
```
montgomp, line 49: `new` feels like it's going to give use sadness at some point (like `delete`). Could we use "created" or something similar?

pasanchMSFT: @msnyder-msft left a comment asking for `new` it is consistent with schema stuff. Should I revert to `discovered`?

msnyder-msft: ahhh that could be fair. Fine. Revert it if you think that's a problem. I think acessing liek like `myObject.new` **should** be fine, but whatevs. Don't feel strongly.

pasanchMSFT: I would rather just send it @montgomp, thoughts?

PR #942, opened 2023-05-01
Just start stubbing in things for feedback

montgomp:
### Summary of changes 🔍 
Starting to stub out bits and pieces to have some skeleton to start building on.

### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Apps/Legion/Contexts/AppDataContext/AppDataContext.types.ts
```diff
@@ -1,4 +1,4 @@
-import { IExtensionClient } from '../../EntryPoints/LegionApp/ShellTypes';
```
msnyder-msft, line 1: What would you think about having another "App" called "Core" or "Shared" and putting stuff like this in there?

src/Apps/IndustrialMetaverse/Site/Site.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { ISiteProps } from './Site.types';
+
+const Site: React.FC<ISiteProps> = ({ extensionClient }) => {
+    if (extensionClient) {
+        return <div>Site</div>;
+    } else {
+        return <div>Site without client</div>;
```
msnyder-msft, line 8: great idea.

msnyder-msft (approved):

src/Apps/IndustrialMetaverse/Site/Site.tsx
montgomp: We should be able to leverage react-router-dom hook `useLocation` to get any pathing and query string parameters to pass from here into the various sub-explorers

src/Apps/IndustrialMetaverse/Site/KPIExplorer/index.ts
```diff
@@ -0,0 +1 @@
+//Not really sure what all needs to go here yet
```
msnyder-msft, line 1: can we keep the same folder structure as we have in the LegionApp where it's got the Adapter, COmponent, Contexts folders under the root?

msnyder-msft: image

montgomp: We have Legion and IM apps. IM only has Site (I could move it to an EntryPoints folder). Site has the two main "explorers" which could have Contexts/Adapters/Hooks of their own or have shared Adapters/Contexts/Hooks in Site. I definitely need to move any .tsx into Components folders and .types.ts into Models.

msnyder-msft (approved):

PR #943, opened 2023-05-02
Save step

ambika-dubey:
### Summary of changes 🔍 
- Added/renamed functions as: getThingCounts (for entity, type, relationship)
- Used those callbacks in SaveStep to conditionally render a line based on the number of new items
- Modified wizardShell to accept the button icon for primary button, and override default text with what's passed in
- Renamed IEntityCounters to IGenericCounters so we can use it for all three things (for now)
Screenshot 2023-05-02 at 12 59 54 PM

### Testing 🧪

### Checklist ✔️
- [ ] Added relevant labels


src/Apps/Legion/Hooks/useEntities.tsx
```diff
@@ -124,9 +129,18 @@ export const useEntities = () => {
              */
             deleteEntity: deleteEntity,
             /** Callback to get entity count per state based on Type */
-            getEntityCount
+            getEntityCount: getEntityCount,
```
msnyder-msft, line 132: can we not use the existing count function? I think it returns an object and one of those values in teh "new" items?

src/Apps/Legion/Hooks/useRelationships.tsx
```diff
@@ -121,7 +127,9 @@ export const useRelationships = () => {
          * Callback to delete the relationship from state.
          * NOTE: this is not a deep update. It will only delete the root level element
          */
-        deleteRelationship: deleteRelationship
+        deleteRelationship: deleteRelationship,
```
msnyder-msft, line 130: instead of a function specific to new, can we do the same thing as entities and just return an object with all the different count types?

src/Apps/Legion/Hooks/useEntities.tsx
```diff
@@ -89,7 +89,7 @@ export const useEntities = () => {
                 deleted: 0
             };
             wizardDataState.entities.forEach((x) => {
-                if (x.typeId === typeId) {
+                if (typeId !== undefined && x.typeId === typeId) {
```
msnyder-msft, line 92: it's probably safer to use the `isDefined` helper since it will also check for null. OR just do `!!typeId`

src/Apps/Legion/Hooks/useTypes.tsx
```diff
@@ -83,6 +83,34 @@ export const useTypes = () => {
...
+                    counters.timeSeries += 1;
+                } else {
+                    counters.asset += 1;
+                }
+            }
+        });
+        return counters;
+    }, [wizardDataState.types]);
+
+    const getNewTypeCount = useCallback((): number => {
```
msnyder-msft, line 109: let's use the same callback pattern here for the count of deleted, etc.

src/Apps/Legion/Hooks/useCustomNavigation.tsx
```diff
@@ -27,7 +27,15 @@ export const useCustomNavigation = (
                     : { onClick: onClick, disabled: disabled }
             }
         });
-    }, [disabled, onClick, primaryAction, wizardNavigationContextDispatch]);
+        // Stringify avoids infinite loop
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
msnyder-msft, line 31: why do we have to disable the rule? Aren't we using primaryAction? or does it not see that?

msnyder-msft:
Can you include a screenshot of different relevant variations of the new screen?

msnyder-msft (commented):
The code looks solid. Let's just take a pass at consolidating the callbacks on the hooks and then add some screenshots to the PR. Nice work!

src/Apps/Legion/Hooks/useEntities.tsx
ambika-dubey: Oops I didn't realize it was looking at the same property! Swapping

src/Apps/Legion/Hooks/useEntities.tsx
ambika-dubey: Originally I had modified this function to take an optional typeId so that if the type was undefined we get total counts not scoped to a type. I need to add that logic back since we'll use this function for total new types. It'll be something like this:
if (!isDefined(typeId) || (isDefined(typeId) && x.typeId === typeId))

src/Apps/Legion/Hooks/useCustomNavigation.tsx
ambika-dubey: I think it gets confused because it's stringified. When I remove the two eslint disable lines, I get the yellow squiggle until I add both (or add primaryAction directly, but that was causing an infinite rerender)

pasanchMSFT: I would have expected you to only need one, similar to the function below, but if you need them both leave them there.

src/Apps/Legion/Models/Wizard.types.ts
```diff
@@ -45,7 +45,7 @@ export interface IViewEntity extends IBaseEntity {
     type: IViewType;
 }
 
-export interface IEntityCounters {
+export interface IGenericCounters {
     new: number;
```
pasanchMSFT, line 49: I think @montgomp was right about this, can we change to 'created' or 'discovered'?

pasanchMSFT (approved):
Looks good, the only thing I would ask you to change is the name of GenericCounters, from `new` to something else.

msnyder-msft (approved):

PR #944, opened 2023-05-03
[Legion: DataSourceStep with PID] Revised source selection and removed old data context

iremgokce:
### Summary of changes 🔍 
- Revised picker components to reuse `CookSource` as a wrapper component both in DataPusher and DataSourceStep and added `TablePicker` for it
- Added mock data for PID cooking
- Removed old data management context and revised existing ones to pass adapter and to set cooked assets in state
- Some style changes to show "new" entities just cooked
- Consolidated all translations (no need to distinguish between 'source' and 'target')

image
image
image
image

### Testing 🧪
You can test the newly added TablePicker and CookSource component stories. Also test changes for DataSourceStep both in its own story and WizardShell stories (both mock and ADX)

### Checklist ✔️
- [x] Added relevant labels


src/Apps/Legion/Components/CookSource/CookSource.tsx
```diff
@@ -0,0 +1,284 @@
...
+            );
+            setSourceTableColumnOptions(
+                data.Columns.map((d) => ({
+                    key: d.columnName,
+                    text: d.columnName
+                }))
+            );
+            onGetTableData?.(data);
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
pasanchMSFT, line 191: why do we need to disable deps here?

src/Apps/Legion/Components/CookSource/CookSource.tsx
```diff
@@ -0,0 +1,284 @@
...
+    }, [getTableState?.adapterResult]);
+
+    useEffect(() => {
+        if (!isClusterVisible) {
+            setSelectedSource({
+                ...selectedSource,
+                cluster: adapter.connectionString
+            });
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
pasanchMSFT, line 201: why disable?

src/Apps/Legion/Components/CookSource/CookSource.types.ts
```diff
@@ -0,0 +1,17 @@
...
+import { IADXAdapterTargetContext } from '../../Models/Interfaces';
+import { SourceType } from '../DataPusher/DataPusher.types';
+import { ITable } from '../../Adapters/Standalone/DataManagement/Models/DataManagementAdapter.types';
+import { ICookSource } from '../../Models';
+
+export interface ICookSourceProps {
+    onSourceTypeChange: (sourceType: SourceType) => void;
+    onSourceChange: (source: ICookSource) => void;
+    onGetTableData?: (table: ITable) => void;
+    targetAdapterContext?: React.Context<IADXAdapterTargetContext>;
```
pasanchMSFT, line 11: how does context as a prop work? Why not useContext instead?

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/ModifyStep.stories.tsx
```diff
@@ -23,7 +23,7 @@ type ModifyStepStory = ComponentStory<typeof ModifyStep>;
 const Template: ModifyStepStory = (args) => {
     return (
         <WizardDataContextProvider
-            initialState={GET_DEFAULT_MOCK_WIZARD_DATA_CONTEXT('Dairy')}
+            initialState={GET_DEFAULT_MOCK_WIZARD_DATA_CONTEXT()}
```
pasanchMSFT, line 26: Is Dairy no longer used or why remove it here?

src/Apps/Legion/Contexts/WizardDataContext/WizardDataContext.tsx
```diff
@@ -121,6 +122,21 @@ export const WizardDataContextReducer: (
                 deletePropertyById(propertyId, draft);
                 break;
             }
+            case WizardDataContextActionType.ADD_COOKED_SOURCE_ASSETS: {
+                const { data } = action.payload;
+                draft.properties = draft.properties.concat(data.properties);
+                const types = deepCopy(data.types);
+                draft.types = draft.types.concat(
+                    types.map((t, idx) => {
+                        t.color = getHighChartColorByIdx(
```
pasanchMSFT, line 131: We should make this a utils function for Legion, no longer rely on highchart colors

pasanchMSFT (commented):
Left a couple of comments, overall looks good and thanks for the refactor to clear out old unused stuff :)

src/Apps/Legion/Components/CookSource/CookSource.tsx
```diff
@@ -0,0 +1,284 @@
...
+    // callbacks
+    const handleSourceTypeChange = useCallback(
+        (_event, option: IDropdownOption) => {
+            setSelectedSourceType(option.key as SourceType);
+            onSourceTypeChange?.(option.key as SourceType);
+            const newSource =
+                option.key === SourceType.Timeseries
+                    ? defaultTimeSeriesSource
+                    : defaultPIDSource;
+            setSelectedSource(newSource);
```
msnyder-msft, line 94: should we create a wrapper to do the "SetSource" operation so it will always call both the local state set and the notification of the callback? I see we repeat that same line in several places, is it worth abstracting so we don't ever set the value wihtout notifying?

src/Apps/Legion/Components/CookSource/CookSource.types.ts
```diff
@@ -0,0 +1,17 @@
+import { IStyle } from '@fluentui/react';
+import { IADXAdapterTargetContext } from '../../Models/Interfaces';
+import { SourceType } from '../DataPusher/DataPusher.types';
+import { ITable } from '../../Adapters/Standalone/DataManagement/Models/DataManagementAdapter.types';
+import { ICookSource } from '../../Models';
+
+export interface ICookSourceProps {
+    onSourceTypeChange: (sourceType: SourceType) => void;
+    onSourceChange: (source: ICookSource) => void;
+    onGetTableData?: (table: ITable) => void;
```
msnyder-msft, line 10: why does the parent get notified when it fetches data?

src/Apps/Legion/Components/DataPusher/DataPusher.tsx
```diff
@@ -40,21 +40,30 @@ const DataPusher: React.FC<IDataPusherProps> = (props) => {
...
                 adapter,
+                selectedClusterUrl,
                 classNames
             }}
         >
             <div className={classNames.root}>
                 <h3>{t('legionApp.dataPusher.title')}</h3>
                 <ClusterPicker
-                    onClusterUrlChange={setSelectedClusterUrl}
+                    onClusterUrlChange={(clusterUrl: string) =>
```
msnyder-msft, line 60: why do we need the wrapper function now? Can't we just pass the function directly if the props are the same?

src/Apps/Legion/Components/Pickers/TablePicker/TablePicker.tsx
```diff
@@ -0,0 +1,94 @@
...
+            const data = getTablesState.adapterResult.getData();
+            setSourceTableOptions(data.map((d) => ({ key: d, text: d })));
+        }
+    }, [getTablesState?.adapterResult]);
+    useEffect(() => {
+        if (databaseName) {
+            setSelectedSourceTable(null);
+            getTablesState.callAdapter({ databaseName: databaseName });
+        }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
msnyder-msft, line 67: why are we overriding the exhaustive deps here?

src/Apps/Legion/Components/CookSource/CookSource.tsx
```diff
@@ -0,0 +1,284 @@
...
+        (clusterUrl: string) => {
+            const newSource: IADXConnection = {
+                ...(selectedSource as IADXConnection),
+                cluster: clusterUrl,
+                database: null,
+                table: null,
+                twinIdColumn: null,
+                tableData: null
+            };
+            setSelectedSource(newSource);
```
msnyder-msft, line 109: seeing so many operations on different parts of state makes me think maybe we should be using a reducer here.

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/ModifyStep.stories.tsx
msnyder-msft: it's the default if not provided.

src/Apps/Legion/Components/WizardShell/WizardShellMockData.ts
```diff
@@ -276,35 +133,35 @@ const setDairyData = (data: IWizardDataContextState) => {
     // PROPERTIES
     const propTemp = getProperty({
         friendlyName: 'Temperature',
-        sourcePropId: 'Temp'
+        sourcePropName: 'Temp'
```
msnyder-msft, line 136: how come we renamed this? I think it was id in the class diagram Matt gave us. 🤔

src/Apps/Legion/Contexts/AppDataContext/AppDataContext.types.ts
```diff
@@ -1,15 +1,19 @@
+import { IDataManagementAdapter } from '../../Adapters/Standalone/DataManagement/Models/DataManagementAdapter.types';
+import { IADXAdapterTargetContext } from '../../Models/Interfaces';
+
 export interface ITargetDatabaseConnection {
     clusterUrl: string;
     databaseName: string;
 }
 export interface IAppDataContextProviderProps {
+    adapter?: IDataManagementAdapter;
```
msnyder-msft, line 9: is it really optional? seems like if we need to fetch data we should juts make someone provide it.

src/Apps/Legion/Contexts/WizardDataContext/WizardDataContext.tsx
msnyder-msft: agreed. Especially since it's not for a chart.

src/Apps/Legion/Services/DataPusherUtils.ts
```diff
@@ -4,176 +4,274 @@ import { ITable } from '../Adapters/Standalone/DataManagement/Models/DataManagem
...
+ * (2) if diagram type, it returns the extracted texts from diagram as twin ids along with other properties like x, y and width, height.
+ * @param sourceType the type of the source
+ * @param source the source to be cooked
+ * @returns ICookedSource which includes models, properties and twins objects
+ */
+export const cookSource = (
+    sourceType: SourceType,
+    source: ICookSource
+): ICookedSource => {
+    switch (sourceType) {
```
msnyder-msft, line 55: can we split out the cooking logic for each type to distinct functions? That will make this scale better as we add more types.

src/Apps/Legion/Components/CookSource/CookSource.tsx
iremgokce: I only want to use side effects with only that change in the dependency list

src/Apps/Legion/Components/CookSource/CookSource.tsx
iremgokce: I only want to use side effects with only that change in the dependency list

src/Apps/Legion/Components/CookSource/CookSource.types.ts
iremgokce: So it is a context property that our useADXAdapter hook will try to fetch the adapter from. inside that hook I use useContext. If there is no context provided to that hook via targetAdapterContext, it will return MockAdapter

src/Apps/Legion/Components/CookSource/CookSource.tsx
msnyder-msft: won't that mean that when propB changes, you might have a stale reference to propA that isn't in the list? maybe not since it isn't a callback? 🤔

src/Apps/Legion/Components/CookSource/CookSource.stories.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export default {
+    title: 'Apps/Legion/Components/CookSource',
+    component: CookSource,
+    decorators: [getDefaultStoryDecorator<ICookSourceProps>(wrapperStyle)]
+};
+
+type CookSourceStory = ComponentStory<typeof CookSource>;
+
+const Template: CookSourceStory = (args) => {
```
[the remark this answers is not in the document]
iremgokce: yes it works, it is a standalone component, and since it is a mock story the useADXAdapter hook internally will use MockAdapter to fetch the data since there is no context with adapter is provided.

src/Apps/Legion/Components/CookSource/CookSource.types.ts
iremgokce: I need table information in some cases (e.g. to show table content in data pusher after selecting table)

src/Apps/Legion/Components/WizardShell/WizardShellMockData.ts
iremgokce: It should have been sourcePropName in the first place:
image

src/Apps/Legion/Contexts/AppDataContext/AppDataContext.types.ts
iremgokce: if you dont provide an adapter MockAdapter is going to be used through useADXAdapter hook

src/Apps/Legion/Components/CookSource/CookSource.tsx
iremgokce: added use-effects for it

src/Apps/Legion/Components/CookSource/CookSource.tsx
iremgokce: There is only 3 state variables, will keep it as is now.

src/Apps/Legion/Components/Pickers/TablePicker/TablePicker.tsx
iremgokce: I only need that side effect when a single value changes in the dependency list

src/Apps/Legion/Components/CookSource/CookSource.tsx
iremgokce: okay, it seems like it is fine here, removing the disabling

src/Apps/Legion/Components/CookSource/CookSource.tsx
iremgokce: Okay, seems like I dont need that, removing disabling

src/Apps/Legion/Contexts/WizardDataContext/WizardDataContext.tsx
pasanchMSFT: I'm ok with the palette, it just felt weird using a chart utility function for this.

pasanchMSFT (approved):

src/Apps/Legion/Contexts/AppDataContext/AppDataContext.types.ts
msnyder-msft: right. I just kinda feel like the props should be designed for a real implmentation vs designing for mock data.

PR #946, opened 2023-05-03
Fluent 9 in Cardboard

pasanchMSFT:
### Summary of changes 🔍 
Added Fluent 9 package, babel configuration and linting rules.

Added Fluent theme provider to `BaseComponent`, and also created very custom Fluent 9 themes so they could match our current themes. Current solution is planned to be temporary, and themes will probably move to what Fluent 9 offers.

Created a component with new Fluent 9 patterns and also some templates to start developing new components with Fluent 9.

image

### Testing 🧪
Open any story and take a look at styles, everything should be the same across all themes.
Open `ConsumeShell` story to take a look at a new component built with Fluent 9 styling.

### Checklist ✔️
- [ ] Added relevant labels


package.json
```diff
@@ -21,6 +21,7 @@
         "generate-schema": "node .\\tools\\generate-3d-schema-types.js",
         "generate": "node ./tools/generate-component-template.js",
         "generate-legion": "node ./tools/generate-legion-template.js",
```
msnyder-msft, line 23: What does `generate-next` do?

src/Apps/Legion/Components/ConsumeShell/ConsumeShell.tsx
```diff
@@ -0,0 +1,52 @@
...
+                style={{ [`${gridBorderVar}`]: theme.palette.glassyBorder }}
+            >
+                <GlobeRegular fontSize={'18px'} />
+                <h3 className={classNames.headerText}>
+                    {appDataState?.targetDatabase?.databaseName}
+                </h3>
+            </div>
+            <div
+                className={classNames.leftNav}
+                style={{ [`${gridBorderVar}`]: theme.palette.glassyBorder }}
```
msnyder-msft, line 43: do you need it in the special ticks? or can it just be the variable directly?

src/Apps/Legion/Components/ConsumeShell/ConsumeShell.types.ts
```diff
@@ -0,0 +1,9 @@
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+export interface IConsumeShellProps {}
+
+export type TConsumeShellClassNames =
```
msnyder-msft, line 4: I'm probably just missing it, but where does this get used?

src/Theming/Fluent9DarkTheme.ts
```diff
@@ -0,0 +1,397 @@
+import { Theme } from '@fluentui/react-components';
+
+export const Fluent9DarkTheme: Theme = {
```
msnyder-msft, line 3: so you have to define everything explicitly? it's not just the overrides?

src/Apps/Legion/Components/WizardShell/Internal/SaveStep/SaveStep.styles.ts
```diff
@@ -9,7 +9,12 @@ const classNames = {
 // export const SAVESTEP_CLASS_NAMES = classNames;
 export const getStyles = (_props: ISaveStepStyleProps): ISaveStepStyles => {
     return {
-        root: [classNames.root],
+        root: [
+            classNames.root,
+            {
+                height: '440px'
```
msnyder-msft, line 15: is this the same height assumption as the other screen? Could we move that to a constant so we can see all the references if/when we replace it?

src/Theming/Fluent9DarkTheme.ts
```diff
@@ -0,0 +1,397 @@
...
+    colorPaletteLightGreenBackground1: '#f2fbf2',
+    colorPaletteLightGreenBackground2: '#a7e3a5',
+    colorPaletteLightGreenBackground3: '#13a10e',
+    colorPaletteLightGreenForeground1: '#11910d',
+    colorPaletteLightGreenForeground2: '#0b5a08',
+    colorPaletteLightGreenForeground3: '#13a10e',
+    colorPaletteLightGreenBorderActive: '#13a10e',
+    colorPaletteLightGreenBorder1: '#a7e3a5',
+    colorPaletteLightGreenBorder2: '#13a10e',
+    colorPaletteMarigoldBackground1: '#fefbf4',
```
msnyder-msft, line 285: Maybe we can have defaults in one file and then have our overrides in another variable or file so we know what we have changed from the default styles?

src/Theming/FluentThemes.ts
```diff
@@ -30,6 +37,21 @@ export const getFluentTheme = (theme: Theme): ITheme => {
     }
 };
 
+export const getVNextTheme = (theme: Theme): Fluent9Theme => {
+    switch (theme) {
+        case Theme.Dark:
+            return Fluent9DarkTheme;
+        case Theme.Light:
```
msnyder-msft, line 44: nit: can remove the light theme and just let it fall through.

msnyder-msft (approved):

package.json
pasanchMSFT: generate fluent 9 template components

src/Apps/Legion/Components/ConsumeShell/ConsumeShell.tsx
pasanchMSFT: I tried using the variable directly, but typescript had issues with understanding it was a string. I can try to type it and see if that works.

package.json
```diff
@@ -21,6 +21,7 @@
...
+        "generate-next": "node ./tools/generate-next-template.js",
```
msnyder-msft, line 24: could we replace `npm run generate` to run this command and rename the existing command to `npm run generate-legacy`?

src/Apps/Legion/Components/ConsumeShell/ConsumeShell.types.ts
pasanchMSFT: This is the equivalent of `IComponentNameStyles` it gives the style function the keys to the styles you want (i.e., root). `makeStyles<IComponentClassNames>` makes sure you don't miss any class name.

src/Theming/Fluent9DarkTheme.ts
pasanchMSFT: I'll iterate on this, see if I can use Partial<Theme> or something else to just define overrides. Just wanted to push this as a v0.

src/Theming/Fluent9ExplorerTheme.ts
```diff
@@ -0,0 +1,397 @@
+import { Theme } from '@fluentui/react-components';
+
+export const Fluent9ExplorerTheme: Theme = {
+    borderRadiusNone: '0',
```
iremgokce, line 4: hmm is there a way to extend these from? Seems like we have the same set of variables both in other theme ts files?

iremgokce (approved):
I left one comment for extending common css properties from theme files.

src/Apps/Legion/Components/ConsumeShell/ConsumeShell.types.ts
msnyder-msft: hm ok. I just wasn't seeing where it was used. I must have missed it.

PR #947, opened 2023-05-10
Entities tab with data grid

pasanchMSFT:
### Summary of changes 🔍 
In this PR a new Fluent component replaces an existing list. Data grid is now being used in Entities tab. Some other minor fixes for themes and pivots were also introduced in this PR.

image

### Testing 🧪
Open WizardShell story and navigate to Modify step. Take a look at Entities tab specifically.

### Checklist ✔️
- [ ] Added relevant labels


src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/EntitiesTab/EntitiesTab.styles.ts
```diff
@@ -1,26 +1,29 @@
...
-        },
-        newEntityIcon: {
-            minWidth: 10,
-            margin: '-16px 0 0 2px'
-        }
+export const EntityTabCSSVar = '--legion-entity-root-max-height';
+
+export const useEntitiesTabClassNames = makeStyles<TEntitiesTabStyles>({
+    root: {
+        maxHeight: `var(${EntityTabCSSVar})`,
```
iremgokce, line 8: is there a reason why we used a css var here and defined the value of that var in the component with style? couldn't we keep it as is?

iremgokce (approved):

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/EntitiesTab/EntitiesTab.styles.ts
iremgokce: also build test fails, can you check it out please?

pasanchMSFT: The height is a calculated value, based on an exported variable. I tried keeping it as is, but it seems `makeStyles` does not allow us to have calculated values as part of its CSS rules. I can try to calculate the value outside of the function. I'll let you know the result of that.

On another note: build is failing because of a snapshot, fixing it.

pasanchMSFT: Looks like `makeStyles` is gets its values on build time. Variables and calculated values need to be assigned to CSS vars, otherwise the build will fail.

iremgokce: oh got it!

iremgokce (approved):

PR #948, opened 2023-05-15
[Legion App: Diagram] Added Diagram component and tab to Wizard

iremgokce:
### Summary of changes 🔍 
- Added standalone `Diagram` component and consumed it in `WizardShell`'s "Modify" step if the selected source is PID in "Add Source" step.
- Added [url-loader](https://v4.webpack.js.org/loaders/url-loader/) module loader for png/jpg to the Storybook's webpack config to be able to render images. I had trouble using the file-loader, so instead using url-loader with file size limit which converts the fetched image src file to base64 string if the size is less than that limit, otherwise it uses file-loader as fallback which resolves the file into a url. I found this [post ](https://medium.com/a-beginners-guide-for-webpack-2/handling-images-e1a2a2c28f8d) helpful.
- Moved some types around to common files as they were consumed by multiple components.
- Some styling changes

image
image

### Testing 🧪
 You can check `Diagram` component story and `WizardShell`'s mock story by selecting the PID type source to see the diagram. We limit the PID options to dropdown selection without letting user enter a pib blob url since we already have the extracted text/cooked information from mock data.

### Checklist ✔️
- [ ] Added relevant labels


src/Apps/Legion/Components/Diagram/Diagram.tsx
```diff
@@ -0,0 +1,137 @@
...
+            <Image
+                ref={imageRef}
+                width={imgWidth}
+                height={imgHeight}
+                onLoad={handleOnImgLoad}
+                src={imageUrl}
+            />
+            {annotationPlacements.map((a, idx) => (
+                <div
+                    key={imageUrl + idx}
```
msnyder-msft, line 118: what do we gain by having hte image url in the key? Doesn't every item in the map have the same url? Can we pick something more unique to each item so that the key is more guaranteed to be unique?
Also, we have seen issues in the past with using index as the primary discriminator in keys so let's not rely on that here since it causes issues if the ordering of items changes in the collection. React then thinks nothing changed and won't rerender those items where the key still matches even though the contents of that item should have been updated.

src/Apps/Legion/Components/Diagram/Diagram.types.ts
```diff
@@ -0,0 +1,22 @@
+import { IViewEntity } from '../../Models';
+
+export interface IDiagramProps {
+    parentRef: React.MutableRefObject<any>;
```
msnyder-msft, line 4: what does the ref give us if we don't know the type of hte ref? Can we not have `any` here?

src/Apps/Legion/Components/Diagram/Internal/Annotation/Annotation.styles.ts
```diff
@@ -0,0 +1,20 @@
+import { FontWeights } from '@fluentui/react';
+import { TAnnotationClassNames } from './Annotation.types';
+import { makeStyles } from '@fluentui/react-components';
+import { shorthands } from '@griffel/react';
+
+export const useClassNames = makeStyles<TAnnotationClassNames>({
+    root: {
+        width: 'fit-content',
+        color: 'black',
```
msnyder-msft, line 9: is the hard coding of black intentional here since we are assuming the image is going to have a white background? Could we make the annotations chagne with the theme and still have things show up properly?

src/Apps/Legion/Components/Diagram/Internal/Annotation/Annotation.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+    logDebugConsole('debug', 'Render');
+
+    return (
+        <div className={classNames.root}>
+            <div className={classNames.type}>
+                <TypeIcon icon={icon} color={color} />
+            </div>
+            <span className={classNames.text}>{text}</span>
+            {isNew && <Image className={classNames.newIcon} src={NewIcon} />}
```
msnyder-msft, line 26: should we create a component for the "new" indicator? so we can just use it anywhere?

src/Apps/Legion/Components/Diagram/__mockData__/CoffeeRoastery.png
msnyder-msft, line 1: Can we get a version of this with a transparent background? That way the tool can add the background?

src/Apps/Legion/Contexts/WizardNavigationContext/WizardNavigationContext.types.ts
```diff
@@ -46,6 +54,7 @@ export type WizardNavigationContextAction =
           type: WizardNavigationContextActionType.NAVIGATE_TO;
           payload: {
               stepNumber: number;
+              initialProps?: IAdditionalStepProps;
```
msnyder-msft, line 57: this doesn't seem like the right approach from a types perspective. Now any step you send will accept these props but only a single one will do anything with the values which is confusing to a consumer. Can we find an alternative way to handle this?

src/Apps/Legion/Components/Diagram/Diagram.tsx
iremgokce: the idea was in case of multiple diagrams are shown up in the same UI, the image url could be used as unique key that I do not see any problem with unless used along with an idx. So, I can generate an id with useId hook instead of the image url.

src/Apps/Legion/Components/Diagram/Diagram.tsx
```diff
@@ -0,0 +1,137 @@
...
+        >
+            {!isReady && (
+                <Spinner
+                    className={classNames.spinnerWrapper}
+                    size={SpinnerSize.medium}
+                    label={t('loading')}
+                />
+            )}
+            <Image
+                ref={imageRef}
```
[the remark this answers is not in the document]
iremgokce: ahh it was left from my previous attempt for the logic to change the width and height of the image (scaling) based on its parent container. Just removed

src/Apps/Legion/Components/Diagram/Diagram.types.ts
iremgokce: It is referring to the container parent element which is HTMLDivElement in our use case. I am adding it now.

src/Apps/Legion/Components/Diagram/Internal/Annotation/Annotation.styles.ts
iremgokce: yeah since the images we have always have white background, that is why I kept it there as black. But I changed it to theme.palette.blueDark to change it with theme. It was the closest to black which shows always a dark color no matter what the selected theme is.

src/Apps/Legion/Components/Diagram/__mockData__/CoffeeRoastery.png
iremgokce: even if we have a transparent background, we need to have versions of the png with the right light line colors to show the lines in the image properly along with the selected theme. I do not think we should put effort into this now.

src/Apps/Legion/Components/Diagram/Diagram.tsx
msnyder-msft: Gotcha. Actually, instead of useId I would just use something unique (or a combo) from the object. The idea of key is that it is generated based off of the item being rendered so it knows whether to rerender the child or not. If the key doesn't change, then when the parent renders it will use the existing value in the virtual DOM instead of recreating fresh DOM elements.
So in this case, if you can use icon+text or something that should be enough.

msnyder-msft (commented):
Responded to a couple comments

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/DiagramTab/DiagramTab.styles.ts
```diff
@@ -0,0 +1,14 @@
+import { TDiagramTabClassNames } from './DiagramTab.types';
+import { makeStyles, shorthands } from '@fluentui/react-components';
+
+export const useClassNames = makeStyles<TDiagramTabClassNames>({
+    root: {},
+    diagramSelector: {
+        width: '308px',
```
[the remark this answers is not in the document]
iremgokce: for picker dropdowns, we have been using a fixed width for either itself or for the parent component when consuming it. Because we want to make sure to show the content of the items properly enough and not to take much space. I can set it to a percentage instead here like 40% but it will unnecessarily take up space when the item content is not that long. I can move it to constant, but I think I prefer it to live in the types file of that component because that constant is only for this component.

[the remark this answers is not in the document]
msnyder-msft: yeah. I guess for dropdowns lets use fixed amounts. I had seen widths in a few other spots and just commented here.

src/Apps/Legion/Models/Wizard.types.ts
```diff
@@ -32,7 +32,7 @@ interface IBaseEntity extends IBase {
      * a property bag of the values that correspond to the properties on the `Type`.
      * Note: the names of the properties here should match the names of the properties on the `Type`.
      */
-    values: Record<string, string>;
+    values: Record<string, any>;
```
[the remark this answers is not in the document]
iremgokce: values can be numbers too. I will make it string | number then.

src/Apps/Legion/Contexts/WizardNavigationContext/WizardNavigationContext.types.ts
iremgokce: I think it is flexible enough right now with optional keys. So if a step requires any initial props to be set along with the step number we can extend the IAdditionalStepProps with optional keys. This way, for any step user can specify which additional props they want. Do you want me to create a mapping of additional props to the current step?

pasanchMSFT: I think the mapping makes sense. Something like
`payload: StepWithProps | StepWithoutProps` or `payload: StepPayload | ModifyStepPayload` and have `ModifyStepPayload` have stepNumber predefined to Modify step's number.

src/Apps/Legion/Components/Diagram/Diagram.tsx
```diff
@@ -0,0 +1,142 @@
+import React, { useCallback, useEffect, useRef, useState } from 'react';
+import { getDebugLogger } from '../../../../Models/Services/Utils';
+import { IDiagramProps, TDiagramAnnotationPlacement } from './Diagram.types';
+import { useClassNames } from './Diagram.styles';
+import { Image, Spinner, SpinnerSize } from '@fluentui/react';
```
pasanchMSFT, line 5: Spinner and Image are available on Fluent 9, would be a good idea to see if those work here.

src/Apps/Legion/Components/WizardShell/Internal/ModifyStep/Internal/DiagramTab/DiagramTab.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { useEntities } from '../../../../../../Hooks/useEntities';
+import { Kind } from '../../../../../../Models';
+import Diagram from '../../../../../Diagram/Diagram';
+import { TDiagramAnnotation } from '../../../../../Diagram/Diagram.types';
+import {
+    PIDSourceUrls,
+    PIDSourceUrlsToImgUrlMapping,
+    PID_EXTRACTED_PROPERTIES
+} from '../../../../../../Models/Constants';
+import { Dropdown, IDropdownOption } from '@fluentui/react';
```
pasanchMSFT, line 20: Dropdown is available in Fluent 9, could be interesting to see if it works here.

pasanchMSFT (approved):
Great work! The component looks awesome. If you want to try out some Fluent 9 components feel free, otherwise I'll approve.

src/Apps/Legion/Contexts/WizardNavigationContext/WizardNavigationContext.types.ts
iremgokce: used discriminating unions for the type here for this action

PR #960, opened 2023-07-19
[Data History Widget]: ADX Cluster url validation fix

iremgokce:
S### Summary of changes 🔍 
- Fixed regex for ADX cluster url validation to let dashes in the name. This will fix (1) not populating the connection string in the widget form UI and (2) being able to query against ADX api with those cluster urls.
- Added console error logs to let the customers trace this error for invalid cluster url so that when they see the generic error message in the widget in viewer mode they can see the error.
- Added translation for invalid ADX cluster url to the form (e.g. in case 3d scene config file is edited manually after creating one). Translation content is subject to change.
- Bugfix for not to override the connection information of existing widgets with adt instance's connection information when editing

image
image
image

### Testing 🧪
- You can test it by creating an adx cluster with name including dashes and replace your data history connection for your instance with that one to see it is working. For the invalid scenario, you need to update your 3d scene config file with an invalid connection cluster url and try to edit that data history widget.

### Checklist ✔️
- [x] Added relevant labels


src/Adapters/ADXAdapter.ts
```diff
@@ -46,15 +46,14 @@ export default class ADXAdapter implements IADXAdapter {
                 : this.adxConnectionInformation.kustoClusterUrl;
 
             if (!isValidADXClusterUrl(clusterUrl)) {
-                adapterMethodSandbox.pushError({
-                    type: ComponentErrorType.DataFetchFailed,
-                    isCatastrophic: true
-                });
                 logDebugConsole(
```
iremgokce, line 49: since pushError throws, logging was getting swallowed, changed the order.

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
```diff
@@ -121,7 +122,8 @@ const DataHistoryWidgetBuilder: React.FC<IDataHistoryWidgetBuilderProps> = ({
             const connection = adxConnectionInformation.connection;
             updateWidgetData(
                 produce(formData, (draft) => {
-                    draft.widgetConfiguration.connection = {
+                    draft.widgetConfiguration.connection = formData
+                        .widgetConfiguration.connection ?? {
```
iremgokce, line 126: not to override existing widget connection information with the adt instance's connection information

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
```diff
@@ -132,7 +134,8 @@ const DataHistoryWidgetBuilder: React.FC<IDataHistoryWidgetBuilderProps> = ({
     }, [adxConnectionInformation]);
 
     const connectionString = formData.widgetConfiguration.connection
-        ? generateConnectionString(formData.widgetConfiguration.connection)
+        ? generateConnectionString(formData.widgetConfiguration.connection) ||
+          t(LOC_KEYS.connectionStringInvalid)
```
iremgokce, line 138: if connection string generation fails due to invalid adx url, shows invalid url message in the input field

src/Models/Services/Utils.ts
```diff
@@ -921,7 +921,7 @@ export const getUrlFromString = (urlString: string): URL => {
 /** Checking if a given ADX cluster url is a safe url following a certain regex and hostname */
 export const isValidADXClusterUrl = (clusterUrl: string): boolean => {
     const isValidADXClusterHostUrl = (urlPrefix) =>
-        /^[a-zA-Z0-9]{4,22}.[a-zA-Z0-9]{1,}\b$/.test(urlPrefix);
+        /^[a-z][a-z0-9-]{3,21}.[a-zA-Z0-9]{1,}\b$/.test(urlPrefix);
```
iremgokce, line 924: Following the docs saying "Cluster name must begin with a letter and contain between 4 and 22 lowercase alphanumeric characters." and it includes dashes even though it is not documented.

src/Resources/Locales/en.json
```diff
@@ -673,6 +673,7 @@
                 "connectionStringInformation": "Data history widgets target data in the Azure Data Explorer instance already connected to this Azure Digital Twins instance.",
                 "connectionLoadingText": "Loading connection information...",
                 "noConnectionInformationText": "No connection information found",
+                "connectionStringInvalid": "Invalid connection string!",
```
iremgokce, line 676: Feel free to suggest any other content here @msnyder-msft @pasanchMSFT

msnyder-msft: yeah. let's remove the ! no need to yell at them. :)
Personally, I'd rather do "Connection string is invalid"

src/Models/Services/Utils.ts
msnyder-msft: In other places, we have used this regex, should we converge them?
const isValidAdtHostUrl = urlPrefix => /^[a-zA-z0-9]{1}[a-zA-Z0-9-]{1,60}[a-zA-Z0-9]{1}(\.api)\.[a-zA-Z0-9]{1,}$/.test(urlPrefix);

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
```diff
@@ -132,7 +134,8 @@ const DataHistoryWidgetBuilder: React.FC<IDataHistoryWidgetBuilderProps> = ({
     }, [adxConnectionInformation]);
 
     const connectionString = formData.widgetConfiguration.connection
-        ? generateConnectionString(formData.widgetConfiguration.connection)
+        ? generateConnectionString(formData.widgetConfiguration.connection) ||
```
msnyder-msft, line 137: how does this look? Is this the value in the input box or some description next to it?

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
msnyder-msft: So if you configured the widget before and now come back to the form, we want to prioritize the previously stored data over some other source? What is that other source?

msnyder-msft (approved):

src/Resources/Locales/en.json
pasanchMSFT: I agree, we should be ok without `!` .

pasanchMSFT (approved):

src/Resources/Locales/en.json
```diff
@@ -673,7 +673,7 @@
                 "connectionStringInformation": "Data history widgets target data in the Azure Data Explorer instance already connected to this Azure Digital Twins instance.",
                 "connectionLoadingText": "Loading connection information...",
                 "noConnectionInformationText": "No connection information found",
-                "connectionStringInvalid": "Invalid connection string!",
+                "connectionStringInvalid": "Connection string is invalid",
```
msnyder-msft, line 676: you'll need to bump the key so that all the other languages regen

src/Components/ADT3DSceneBuilder/Internal/Behaviors/Widgets/WidgetBuilders/DataHistoryWidgetBuilder/DataHistoryWidgetBuilder.tsx
iremgokce: yes, the connection information in the 3d scene config should be the one that we see in the form even if it is not the same as the one of the selected adt instance's data history connection information. It was a bug on our side that I just noticed. The other source is the adt instance's data history connection.

It is uncommon that users would use different 3d scene config files which includes connection information for their widgets which are not the same as their selected adt instance's adx connection information. But since I used a different one when I was testing the customer incident, I noticed that bug. We were overriding the connection string whenever we update a data history widget with the selected adt instance's connection information without relying on existing connection information in the config file.

PR #974, opened 2024-01-11
feat: Build with node 16 + get off of node-sass

msnyder-msft:
### Summary of changes 🔍 
Node-sass [is deprecated](https://sass-lang.com/blog/libsass-is-deprecated/), and is what was pinning us to Node 14 (which is also deprecated) so we will get off of node-sass using Dart-Sass which will free us up to get to node 16 which is currently in LTS.  Dart-sass exposes the same JS API so it's a drop-in replacement.

Migrated our node-sass command line to dart-sass command line - [docs](https://sass-lang.com/documentation/cli/dart-sass/)

Note: NPM package install started failing with a conflicting peer dependency if we go to node 18, so postponing that migration for another time.

### Testing 🧪
- I ran the build locally and in the pipelines to make sure those worked.
- I ran storybook and it looks like all the styling stuff is still looking good.
image


msnyder-msft:
When we try to get to node 18 we get stuck with this failure and it's been a rabbit hole.
image

package.json
```diff
@@ -46,7 +49,7 @@
         "@monaco-editor/react": "^4.3.1",
         "@types/react-transition-group": "^4.4.2",
         "ajv": "^8.10.0",
-        "axios": "^0.21.1",
+        "axios": "^1.6.8",
```
msnyder-msft, line 52: please revert this to 0.21.1

package.json
```diff
@@ -37,6 +37,9 @@
         "test:watch": "jest --watch",
         "test": "jest"
     },
+    "resolutions": {
```
msnyder-msft, line 40: I think we need to remove this as well.

rapotl: good catch! i just removed the resolutions part

package.json
rapotl: updated the axios back to ^0.21.1 and also seeing tarball getting generated

rapotl:
ADT Explorer - TwinGraph Evidence of Test:
twingraph-AdtExp-eot

.github/workflows/build.yml
```diff
@@ -11,12 +11,12 @@ jobs:
...
             - uses: actions/setup-node@v2
               with:
-                  node-version: '14'
+                  node-version: '16.10.0'
                   cache: 'npm'
                   registry-url: 'https://registry.npmjs.org'
 
             - name: Install dependencies
-              run: npm install
+              run: npm install --legacy-peer-deps
```
msnyder-msft, line 19: did you hvae to run with the --legacy-peer-deps locally? I added this early on, but I wonder if it's actually needed now? It'd be ideal if we can remove it.

rapotl: I did use the run: npm install --legacy-peer-deps to install the npm packages, we might need that as executing just "npm i" doesn't really install successfully.

that's what i see when ran "npm install"
image

pasanchMSFT (approved):

pasanchMSFT (approved):

PR #983, opened 2024-05-02
Revert axios

pasanchMSFT:
### Summary of changes 🔍 
Revert dependabot update to axios due to runtime issues with TextEncoder.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,9 +37,6 @@
         "test:watch": "jest --watch",
         "test": "jest"
     },
-    "resolutions": {
-        "axios": "^1.6.0"
-    },
     "dependencies": {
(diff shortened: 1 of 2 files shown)


msnyder-msft (approved):

montgomp (approved):

package.json
```diff
@@ -37,9 +37,6 @@
         "test:watch": "jest --watch",
         "test": "jest"
     },
-    "resolutions": {
```
msnyder-msft, line 40: if I"m reading things right here. https://github.com/axios/axios/pull/5530 we could also jump to version 1.2.0 and should be fine.

PR #999, opened 2025-06-24
Fixed fonts

ElliotWood:
Summary of changes 🔍
Fixed font rendering issues by updating FluentUI React dependency.

Testing 🧪

- Verify fonts are loading correctly across all components
- Test in multiple browsers to ensure consistent font rendering
- Check that no visual regressions occurred from the dependency update 
_(Partial: everything looks centered but doesnt appear to be due to this change?)_

Technical Details:

- Updated @fluentui/react to version ^8.123.0 which resolves the underlying issue where older font references were pointing to an expired CDN endpoint.
- Root Cause:
- Previous FluentUI version was referencing fonts from a CDN that had expired, causing font loading failures.
- Solution:
- The newer FluentUI version includes updated font references that point to active CDN endpoints.

image


ElliotWood:
@msnyder-msft Hey mate if you get a sec to review that'd be awesome

